PR #96, opened 2015-06-04
sample chart demo with package added

frenchbread:


.meteor/packages
```diff
@@ -56,3 +56,4 @@ yogiben:admin
 timmyg:wow
 tsega:skrollr
 accounts-github
+mrt:chartjs
```
brylie: Is there another Chart.js package? I think the `mrt` packages may be deprecated.

frenchbread:
@brylie Yes, there is - [gerardnantel:chartjs](https://atmospherejs.com/gerardnantel/chartjs). I've already updated current branch with new package for chartjs. Tested it and it works. But I faced a small bug - sometimes you would need to update page couple of times until the chart itself loads up.

PR #132, opened 2015-06-14
Feature/material bootstrap

elnzv:


elnzv:
Hey @apinf/developers review this pull request.

.gitignore
```diff
@@ -1,6 +1,6 @@
 .meteor/local
 .meteor/meteorite
 .meteor/.id
+/config.json
```
brylie: We have not decided on 'config.json' for team configuration files. I am still using 'settings.json' in my local development process. We need to talk about what configuration file name to use going forward.

.gitignore
```diff
@@ -1,6 +1,6 @@
...
 settings.json
 .idea/
-settings.json
```
brylie: Removing this line without notice may allow team members to inadvertently commit their private settings.

client/style/js/ripples.js
```diff
@@ -0,0 +1,324 @@
+/* Copyright 2014+, Federico Zivolo, LICENSE at https://github.com/FezVrasta/bootstrap-material-design/blob/master/LICENSE.md */
```
brylie: It is becoming common to put 3rd party libraries, such as this, in *client/compatibility*

http://joshowens.me/how-to-organize-your-meteor-js-app/

client/style/js/material.js
```diff
@@ -0,0 +1,226 @@
+/* globals jQuery */
```
brylie: Consider putting this in *client/compatibility*

.gitignore
frenchbread: Agreed with Brylie at this point, but as additional solution settings.json file can be stored somewhere else, for example on desktop. And is this case launching of the dev server would be:
```
meteor --settings /Users/user/Desktop/settings.json
```
This is just an example. I don't mind keeping settings.json file in .gitignore

elnzv: Yep, that was a duplication.

PR #112, opened 2015-06-12
Feature/form configure api

elnzv:

PR #143, opened 2015-06-16
pull request with fixes to #112

elnzv:
pull request with fixes to #112


collections/backend.js
```diff
@@ -201,7 +204,7 @@ ApiBackendsSchema = new SimpleSchema({
   },
   "required_roles.$": {
```
brylie: Is the '.$' required here? It may work fine to remove them. I think the dollar sign is used for Object and Array types.

Issue #121, opened 2015-06-12
Create basic API grid for API user dashboard


bajiat:
The API user dashboard needs a grid layout to display widgets. Check options available for Bootstrap dashboard grid layouts and implement one of them.

Definition of done
---
* [x] Document the available options or examples of Bootstrap dashboard grid.
* [x] Choose one and implement in API user dashboard.
* [ ] Remove unneeded API user dashboard elements (Posts).

bajiat:
Related to #120 and #119.

JuusoV:
Created basic grid. Looking to hear some feedback on it before moving further.

Im not seeing the documenting part on this task is neccessary as long as the implementation is self-explanatory in the code.

PR #154, opened 2015-06-23
Created grid, removed posts.

JuusoV:
Created grid, removed posts.
Task #121

### DIFF
--- a/client/views/dashboard/dashboard.html
+++ b/client/views/dashboard/dashboard.html
@@ -8,12 +8,12 @@
         </div>
       </div>
       <div class="row">
-        <div class="col-lg-6 col-md-6 col-sm-6">
+        <div class="col-lg-6 col-md-6 col-sm-12">
           <p>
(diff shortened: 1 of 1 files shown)


client/views/dashboard/dashboard.html
```diff
@@ -3,95 +3,56 @@
     <div class="container">
       <div class="row" id="champion-messages">
         <div class="col-md-3">
-          {{> addPost}}
           {{> favoritesSidebar}}
         </div>
-        <div class="col-md-9">
-          {{> posts}}
+
+<div class="container-fluid">
```
brylie: Only one container is necessary (see line 3). Remove extra container div. Also, clean up the indentation, so that it is consistent with the rest of the file.

brylie: There is likely a syntactic error in the markup here. Make sure to check the nesting of your divs, including the closing tags. Specifically, the hierarchy should look like:

* container
  * row
    * col
    * col
  * row
    * col
    * col

PR #155, opened 2015-06-23
Feature/api umbrella user relationships

brylie:
Connecting ApiUmbrellaUsers collection and Meteor.users via autoform and collection helper.


elnzv:
@brylie looks good, merging.

PR #156, opened 2015-06-24
Feature/api backends catalog

JuusoV:
Created /catalog for consumer.
Also changed indentation of adminconfig which caused an error.

### DIFF
--- a/client/views/api_backends/api_backends.html
+++ b/client/views/api_backends/api_backends.html
@@ -1,4 +1,4 @@
-<template name="x">
+<template name="apiBackends">
   <div class="row">
     <div class="col-lg-6 col-lg-offset-3">
       {{#autoForm collection="ApiBackends" doc=this id="apiBackends"}}
--- a/client/views/api_backends/api_backends.html
+++ b/client/views/api_backends/api_backends.html
@@ -1,4 +1,4 @@
-<template name="apiBackends">
+<template name="x">
   <div class="row">
     <div class="col-lg-6 col-lg-offset-3">
       {{#autoForm collection="ApiBackends" doc=this id="apiBackends"}}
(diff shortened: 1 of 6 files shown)


widgets/catalogue.html
```diff
@@ -0,0 +1,12 @@
+<template name="catalog">
+  <div class="col-md-12">
+    <h3>This is a table showing available APIs</h3>
+    {{> table }}
+  </div>
+
+
+</template>
+
+<template name="table">
```
brylie: This template inclusion is redundant, and can simply be placed in the catalog template ({{> tabular ... }}).

client/views/api_backends/api_backends.html
```diff
@@ -1,4 +1,4 @@
-<template name="apiBackends">
+<template name="x">
```
brylie: Remove unused templates and files.

widgets/catalogue.html
```diff
@@ -0,0 +1,12 @@
+<template name="catalog">
```
brylie: This template name conflicts with another template.

client/views/catalogue/catalogue.html
```diff
@@ -0,0 +1,4 @@
+<template name="catalogue">
```
brylie: This template name uses British English version of 'catalogue', and may need to be renamed for consistency.  However, renaming the template to 'catalog' will cause a conflict. Choose one language convention (British or American English) and rename the templates so they are both distinct.

lib/router/main.coffee
```diff
@@ -7,6 +7,10 @@ Router.map ->
     path: "/apibackends"
     layoutTemplate: "homeLayout"
 
+  @route "catalog",
```
brylie: The route name 'catalog' is inconsistent with the template name 'catalogue'. Choose one language convention (British or American English) and rename routes and templates so they are consistent (and distinct where necessary).

PR #161, opened 2015-06-25
Feature/adding chart widget to dashboard

frenchbread:


frenchbread:
Hey @apinf/developers , please review this pull request. Basically the main changes here is couple of packages added for elasctic search and etc. Chart itself is moved from /chart to /dashboard. It provides real data fetched from apiUmbrella ElasticSearch analytics (for now, data is provided for just one month - there are also some suggestions to make this chart dynamic, so that user will be able to set time range and review data). 

Next step is to combine sidebar on which @elnzv is working on, with the chart.

PR #198, opened 2015-07-02
Feature/chart code refactoring

frenchbread:
Here is settings.json file example (elasticsearch.host parameter is required for app to fetch analytics data):

```
{
  "api_umbrella": {
    "api_key": "...",
    "auth_token": "...",
    "base_url": "https://.../api-umbrella/"
  },
  "elasticsearch": {
    "host": "http://...:14002"
  }
}

```


server/elasticSearch.js
```diff
@@ -0,0 +1,70 @@
+es = function () {
```
brylie: `es` might be improved by adding more semantic details. E.g. ApiUmbrellaElastic.

server/elasticSearch.js
```diff
@@ -0,0 +1,70 @@
...
+
+  ElasticSearch = Meteor.npmRequire('elasticsearch');
+
+  EsClientSource = new ElasticSearch.Client({
+    host: Meteor.settings.elasticsearch.host
+  });
+
+  EsClient = Async.wrap(EsClientSource, ['index', 'search']);
```
brylie: Add comments to describe this line.

server/elasticSearch.js
```diff
@@ -0,0 +1,70 @@
...
+
+  // index: index provided within the query
```
brylie: Use JsDoc style [@params](http://usejsdoc.org/tags-param.html) here, as well as the general [JsDoc comment style](http://usejsdoc.org/about-getting-started.html).

server/elasticSearch.js
```diff
@@ -0,0 +1,70 @@
...
+  // type : type of records ro be returned
+  // count: limit of records to be returned
+
+  this.doSearch = function (index, type, limit) {
+    var searchData = EsClient.search({
+      index: index,
+      type: type,
+      body: {
+        query: {
```
brylie: How can we allow users to specify the query here?

frenchbread: This is in progress, as well as filtering, and will we added soon.

server/elasticSearch.js
```diff
@@ -0,0 +1,70 @@
...
+    var items = data.hits.hits;
+
+    if(items){
+
+      var datesArray = [];
+      var monthData = {};
+      var chartDataArr = [];
+      var monthFrames = {
+        monthStart: 1,
+        monthEnd  : 31
```
brylie: Month length should not be hard-coded. Here are a couple of possible methods for getting the number of days in any given month:
* http://stackoverflow.com/a/1810990/1191545
* [moment(year-month-day).endOf('month');](https://github.com/moment/moment/issues/242#issuecomment-38886056)

server/elasticSearch.js
```diff
@@ -0,0 +1,70 @@
...
+      };
+
+      items.forEach(function (e) {
+        var stamp = new Date(e._source.request_at);
+        var date = stamp.getDate();
+        datesArray.push(date);
+      });
+
+      datesArray.forEach(function (j) {
```
brylie: What do `j` and `k` represent? What would be more meaningful as variable names? We will end up reading this code more often than it is written.

server/elasticSearch.js
```diff
@@ -0,0 +1,70 @@
...
+
+      var datesArray = [];
+      var monthData = {};
+      var chartDataArr = [];
+      var monthFrames = {
+        monthStart: 1,
+        monthEnd  : 31
+      };
+
+      items.forEach(function (e) {
```
brylie: Rename `e` to `item`, or some more meaningful name.

server/elasticSearch.js
```diff
@@ -0,0 +1,70 @@
...
+        var stamp = new Date(e._source.request_at);
+        var date = stamp.getDate();
+        datesArray.push(date);
+      });
+
+      datesArray.forEach(function (j) {
+        for (var k = monthFrames.monthStart; k <= monthFrames.monthEnd; k++) {
```
brylie: To improve readability, consider rewriting this by using JS [`while`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/while) or underscore [`times()`](http://underscorejs.org/#times) function.

server/elasticSearch.js
```diff
@@ -0,0 +1,70 @@
...
+          if (k == j) {
+            if(monthData[j]) monthData[j]++;
+            else monthData[j] = 1;
+          }
+        }
+      });
+
```
brylie: Add comment briefly describing this loop in plain words.

server/elasticSearch.js
brylie: Add comment briefly describing this loop in plain words.

server/elasticSearch.js
brylie: Add comment briefly describing this loop in plain words.

server/methods/elasticSearch.js
```diff
@@ -1,65 +1,13 @@
 Meteor.methods({
   "getChartData": function () {
-    var chartData = ReactiveVar();
-    chartData.set("");
-    // loading the npm module
-    ElasticSearch = Meteor.npmRequire('elasticsearch');
+    var newSearch = new es();
```
brylie: Remember to update this `es()` call with your new Class name.

server/methods/elasticSearch.js
```diff
@@ -1,65 +1,13 @@
 Meteor.methods({
   "getChartData": function () {
```
brylie: This method should probably accept the same parameters that any contained functions expect. That way, the parameters are not hard-wired, and the method can be used in other contexts.

server/methods/elasticSearch.js
```diff
@@ -1,65 +1,13 @@
...
-    // loading the npm module
-    ElasticSearch = Meteor.npmRequire('elasticsearch');
+    var newSearch = new es();
 
-    // create the client
-    EsClientSource = new ElasticSearch.Client({
-      host: 'http://46.101.247.242:14002'
-    });
+    // TODO: dynamic input data
+    var searchIndex = 'api-umbrella-logs-v1-2015-07';
```
brylie: Add these variables as parameters to the method.

server/methods/elasticSearch.js
```diff
@@ -1,65 +1,13 @@
...
-
-      for (var l = monthFrames.monthStart; l <= monthFrames.monthEnd; l++){
-        chartDataArr.push(dates[l]);
-      }
-      chartData.set(chartDataArr);
-    }else {
-      chartData.set("Not Found");
-    }
-    return chartData.get();
+    var data = newSearch.doSearch(searchIndex, searchType, returnItemsLimit);
```
brylie: Consider how to make this `data` variable an property of the `es` class. I.e. when constructing an instance of `es()`, you can set the data attribute just as you do from this code:

```
var data = this.doSearch(searchIndex, searchType, returnItemsLimit);
```

server/methods/elasticSearch.js
```diff
@@ -1,65 +1,13 @@
...
+    return newSearch.getMonthAnalytics(data);
```
brylie: Likewise, you can create a property of the `es()` class that is defined when instantiated, such as:

```
var monthlyAnalytics = this.getMonthlyAnalytics(this.searchData);

Comment (brylie) [inline server/methods/elasticSearch.js]: Finally, you would just `return newSearch.monthlyAnalytics`.

File: server/methods/elasticSearch.js [hunk 15]
```diff
@@ -1,65 +1,13 @@
...
   }
```
brylie: One other thing is that we might want to show weekly or other time duration analytics, so it might be advisable to refactor this to take a start and end date.

E.g. 

```
return newSearch.getAnalytics(start, end);
```

Issue #196, opened 2015-07-02
Collection hooks package conflicts.


brylie:
I am getting the following error when starting Meteor:

```
Error: Tabular.Table options must specify collection
at new Tabular.Table (packages/aldeed:tabular/common.js:20:1)
     at packages/yogiben:admin/lib/both/startup.coffee:100:27
     at Function._.each._.forEach (packages/underscore/underscore.js:113:1)
     at adminCreateTables (packages/yogiben:admin/lib/both/startup.coffee:77:4)
     at __coffeescriptShare (packages/yogiben:admin/lib/both/startup.coffee:178:2)
```

What are we doing to resolve this error?

brylie:
I found out that the source of this error is probably related to the matb33:collection-hooks package. When removing the collection-hooks package, the error is gone.

I have opened support requests with three upstream packages:

* [aldeed:tabular #177](https://github.com/aldeed/meteor-tabular/issues/170)
* [matb33:collection-hooks #133](https://github.com/matb33/meteor-collection-hooks/issues/133)
* [yogiben:meteor-admin #126](https://github.com/yogiben/meteor-admin/issues/226)

This blocks progress in task #1

brylie:
I have posted an issue describing the [load order of the 'comments' and 'collection-hooks' packages in yogiben:meteor-starter](https://github.com/yogiben/meteor-starter/issues/84) that may be the underlying cause of this issue.

PR #205, opened 2015-07-06
Added collection-hooks (above comments).

brylie:
Related issue #196


brylie:

PR #223, opened 2015-07-08
elasticrest package added & code cleanup

frenchbread:
Hey @apinf/developers please review this pull request. I have pretty much cleaned up everything as well as added `elasticrest` package. Since `ApiUmbrellaElastic` class with its method has been moved to separate package, just `getMonthAnalytics` function is left in that file.


server/elasticSearch.js
```diff
@@ -1,93 +1,57 @@
...
-      var values = [];     // exact values that are used in chart
-      var counts = {};     // object that handles dates and amount of requests this day
 
-      // time filter for fetching data
-      var timeFrames = {
-        start: moment().subtract(2, "weeks"),
-        end  : moment()
-      };
+    // counting total number of requests per day
+    datesArray.forEach(function(x) { counts[x] = (counts[x] || 0)+1; });
```
brylie: Why do you add one here?

server/elasticSearch.js
```diff
@@ -1,93 +1,57 @@
...
 
-      // looping items, getting the date value from an item and pushing it to datesArray
-      items.forEach(function (item) {
-        var stamp = new Date(item._source.request_at);
-        var date = stamp.getDate();
-        datesArray.push(date);
-      });
+    // looping through all the dates that are in needed range
+    var loopThroughDates = moment().range(timeFrames.start, timeFrames.end).by('days', function(moment){
```
brylie: Check indentation.

server/elasticSearch.js
```diff
@@ -1,93 +1,57 @@
...
 
+      var date = moment.date(); // variable handling current date
```
brylie: How are months displayed? E.g. if the date range is from June 28th - July Nth?

server/elasticSearch.js
frenchbread: @brylie  It is counting function, so it increments a counter value when it finds the same value in an datesArray.

server/methods/elasticSearch.js
```diff
@@ -1,16 +1,17 @@
 Meteor.methods({
   "getChartData": function () {
```
brylie: Parametrize this function to accept index, type, and limit inputs.

server/methods/elasticSearch.js
```diff
@@ -1,16 +1,17 @@
...
-    var newSearch = new ApiUmbrellaElastic();
+
 
     // TODO: dynamic input data
-    var searchIndex = 'api-umbrella-logs-v1-2015-07';
-    var searchType = 'log';
-    var returnItemsLimit = 1000;
+    var index = 'api-umbrella-logs-v1-2015-07';
```
brylie: Get these values from the Method call parameters.

server/elasticSearch.js
brylie: OK, cool. I thought it was a data value displayed in the chart. All good.

brylie:
Thanks for your work here and for clarifying the code with comments.

PR #289, opened 2015-07-28
Feature/elasticsearch fields argument

frenchbread:


frenchbread:
Hey @apinf/developers please review this PR and merge. This is basically some fixes to dc charts. I've updated [meteor-elastic-rest](https://github.com/frenchbread/meteor-elastic-rest) package, so now in accepts one more argument - `fields` to be returned. Time for rendering chart now decreased from couple of seconds to less then one second.
Summing up:
1. Improved charts loading time.
2. Resolved security issues. (since you need to provide fields to be returned)

PR #300, opened 2015-08-03
Moved collection files.

brylie:
All collection files now in `both/collections`.


brylie:

PR #301, opened 2015-08-03
added overview chart initialisation

frenchbread:
@apinf/developers Here is the hot fix regarding overview chart. Chart initialisation function was somehow removed from develop branch after merging. Here is the fix, tested it and generally it works but area selection does not force chart refresh now, still zooming works.


frenchbread:
Tested this one again. And it turns out that the code refactoring that we did (dividing code for rendering charts to separate functions like `drawChart`, `parseData`, `renderCharts`) somehow does not let the lineChart to refresh when some filtering or selection has been made on a `overview` chart.

@brylie Any thoughts at his point? I think we can leave it like this or return to one-function view to have this functionality present.

frenchbread:
@apinf/developers This PR is ready to be merged.

client/views/dashboard/charts/charts.js
```diff
@@ -136,19 +136,36 @@ Template.chartsLayout.created = function () {
 
     var chart = dc.lineChart("#line-chart");
     var countryChart = dc.barChart("#bar-chart");
+    var overview = dc.barChart("#overview-chart");
 
     chart
       .width(1140)
```
brylie: What happens if you omit the chart width? Does the chart become responsive?

client/views/dashboard/charts/charts.js
```diff
@@ -136,19 +136,36 @@ Template.chartsLayout.created = function () {
...
+      .mouseZoomable(true)
+      .rangeChart(overview)
       .renderArea(true)
       .dotRadius(3)
+      .brushOn(false)
       .renderHorizontalGridLines(true)
       .renderVerticalGridLines(true);
 
+    overview
+      .width(1140)
```
brylie: Again, try removing the chart width, to see if it improves the flexibility of the chart width.

frenchbread:
@brylie When I remove `width`, this happens:
chart
so it does not respond to parent container width.

client/views/dashboard/charts/charts.js
frenchbread: @brylie Fixed this one. Chart's width is dynamic and fits the parent container on load. Still not responsive though.

client/views/dashboard/charts/charts.js
frenchbread: Done

Issue #181, opened 2015-06-26
Filter the chart to show a given user's API usage


bajiat:
Create a method to filter API usage by API key.

Definition of done
---
* [x] Create the method for querying usage logs by API key.
* [x] Create chart showing individual API usage.
* [x] Populate existing chart with a given user's API usage data.
* [x] Link user privilege to chart data.

PR #313, opened 2015-08-05
Feature/filtering data by api

frenchbread:
@apinf/developers Please review this PR. #181 - related issue. All visible analytics data is filtered by current user's api key.


server/methods/elasticSearch.js
```diff
@@ -1,11 +1,49 @@
...
+    // get user role & check user role
+    if (Roles.userIsInRole(loggedInUser, ['admin'])) {
+
+      // if admin - match_all
+      // construct query
+      query = {
+        match_all: {}
+      }
+
+    } else if (Roles.userIsInRole(loggedInUser, ['owner'])) {
```
brylie: This is unnecessary, as we may not define the user role 'owner'. The current plan is to add an 'owners' field to each API backend, containing an array of owner IDs or emails.

brylie:
This looks good, but has unnecessary logic for the API Owner role. Please remove the else if statement, and we can add it later based on the [owners field](https://apinf.hackpad.com/Permissions-o6FqN3TtxIz#:h=Proposed-schema-change) (not yet implemented).

Issue #231, opened 2015-07-10
Re-design API configuration form


bajiat:
Define how to show the fields  that are advanced and/or optional in the API backend configuration form. Possible options: collapsible sections or buttons that display popup forms. Related to #216 and #246. You can use API Umbrella API configuration form as reference.

Definition of done
---
* [x] Create wireframe demonstrating desired layout.
* [x] List available Bootstrap UI elements relevant to desired layout.
* [x] Indicate on wireframe the HTML elements and CSS classes.
* [x] Create corresponding mockup with placeholder form elements.
* [x] Add autoform elements or markup to existing mockup.

elnzv:
Added a collapses to hide & show optional sections:
https://github.com/apinf/api-umbrella-dashboard/commit/0cfbd08a5131b1efaa14cd790dede7fff136f966, https://github.com/apinf/api-umbrella-dashboard/commit/c73961e6677469c9bb88ad114c05f61bb3afbb84, https://github.com/apinf/api-umbrella-dashboard/commit/564e57fd060ef4f7ac6add26b5e86256060c2529, https://github.com/apinf/api-umbrella-dashboard/commit/8286d056237095b50aa12cc86ae0731b5906e0ef.

elnzv:
New fields:
https://github.com/apinf/api-umbrella-dashboard/commit/81406bdf2644f820ca1c76c820f7df5c0f0feda3, https://github.com/apinf/api-umbrella-dashboard/commit/c80e6810c0632a4e52fbebcb6ab59ba4365020e0, https://github.com/apinf/api-umbrella-dashboard/commit/0970dc50daa6d041a7308788102a01f2fe76ff3a, https://github.com/apinf/api-umbrella-dashboard/commit/38cca8eac0d2d7a028ba8b6654370d4a2b1cddf0, https://github.com/apinf/api-umbrella-dashboard/commit/9cf5e6f62b12c3615d785945b095fa43c5be2359

PR #315, opened 2015-08-06
Feature/redesign api backend

elnzv:
Closing #231


JuusoV:
When selecting Rate limit mode:  'Custom Rate Limits'  under Sub-URL Request Settings the conditional doesnt work and open up the rate limit settings as it should. 
See screenshot for the field Im refering to: 
screenshot-localhost 3000 2015-08-06 12-04-48

Issue #322, opened 2015-08-07
Refresh heatmap when data is returned from ElasticSearch


bajiat:
Currently the heatmap is only updated when the user interacts with the data, but it should be automatically updated when the data is returned from the query. 

Definition of done
---
* [ ] Determine the point when the map can be refreshed.
* [ ] Implement a hook to refresh the map when the data is available.
* [ ] Ensure that the map is refreshed every time data is filtered is interacting with charts.

frenchbread:
"Refresh" function can be actually placed in filter.js file but the problem is that once leaflet map has been initialized it is not possible to reinitialize one. There is built in "update" function (within leaflet) but even if calling it, data/heat won't update. The solution I came up with that seemed to work is onFiler event, removing DOM element from the page and reinitialize leaflet map with fresh data. Not the best approach but at least this could work

PR #338, opened 2015-08-07
Feature/map for merge

JuusoV:
There is still issue with heat not being rendered to the map when the map is being initialized. Related #322


client/views/dashboard/charts/charts.html
```diff
@@ -22,4 +22,9 @@
       {{> dataTable}}
     </div>
   </div>
+  <div class="row">
+    <div class="col-lg-12 col-md-12 col-sm-12">
```
brylie: Please add AdminLTE markup around the map element.

client/views/map/map.js
```diff
@@ -0,0 +1,45 @@
...
+    // Defines the density for the heatmap
+    var density = 1000;
+    var blur = 5;
+
+    // Gets data from the ElasticSearch
+    Meteor.call("getChartData", input, function(err, data) {
+      var items = data.hits.hits;
+      //loops throught the array of objects
+      items.forEach(function(item) {
+        addressPoints.push([item._source.request_ip_location.lat, item._source.request_ip_location.lon, density, blur])
```
brylie: 'density' and 'blur are not necessary in the address point. The only necessary data are 'lat' and 'lon' coordinates.

JuusoV:
@brylie I made the changes you suggested. Could you check again and merge so we finally get this closed?

PR #356, opened 2015-08-12
Swagger editor

JuusoV:
Integrated Swagger editor with iframe


client/views/swagger/swagger-editor/swagger-editor.html
```diff
@@ -0,0 +1,3 @@
+<template name="editor">
+  <iframe src="http://editor.apinf.io:49160" width="100%" id="swagger-editor"></iframe>
```
brylie: The iframe src is currently hard-coded with our domain name and port. Move these into the config.json file for now, so that the swagger editor can run on any server deployment.

brylie:
This pull request contains only one commit, but two files were edited. In general, strive to separate commits describing each significant file change.

JuusoV:
Made changes as per @brylie suggested.

Issue #347, opened 2015-08-12
Create route and view for editing an existing API backend


bajiat:
Create a view for  editing API backend. Related to #324.

Definition of done
---
* [x] Duplicate Add API form.
* [x] Change the title to "Edit API"
* [x] Change the form method to "Update".
* [x] Create a new route: /api/{id}/edit

frenchbread:
PR - #359

PR #359, opened 2015-08-13
Feature/api backend edit

frenchbread:
@apinf/developers please review this PR. Task is done regarding definition of done listed in #347. But submit button still does not update apiBackend in mongo colletion.

### DIFF
--- a/lib/router/main.js
+++ b/lib/router/main.js
@@ -43,7 +43,7 @@ Router.map(function() {
   this.route("editApiBackend", {
     path: "/api/:_id/edit",
     waitOn: function() {
-      return this.subscribe('singleApiBackend', this.params._id);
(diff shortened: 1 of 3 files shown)


frenchbread:
#347 - related issue

lib/router/main.js
```diff
@@ -39,4 +39,18 @@ Router.map(function() {
...
     render: "importApiConfiguration"
   });
+
+  this.route("editApiBackend", {
+    path: "/api/:_id/edit",
+    waitOn: function() {
+      return this.subscribe('singleApiBackend', this.params._id);
+    },
+    data: function () {
+      var templateDate = ApiBackends.findOne({_id: this.params._id});
```
brylie: Perhaps rename this variable to apiBackend, so it is consistent with its purpose.

PR #365, opened 2015-08-14
Feature/api backend managers

brylie:


brylie:
@apinf/developers please review ASAP.

both/collections/backend.js
```diff
@@ -614,5 +631,32 @@ ApiBackends.attachSchema(ApiBackendsSchema);
...
+    var managerIds = backend.managerIds;
+
+    // Make sure current user is a backend manager
+    if (_.contains(managerIds, userId)) {
+      // User is allowed to perform action
+      return true;
+    } else {
+      // User is not allowded to perform action
+      return false;
+    }
```
frenchbread: Found some code duplicating under `update` and `remove`. As there gonna be some actions performing inside conditional statements, what do you think if using `isBackendManager` method attached to user schema. So the conditional check would be ?
```
if (user.isBackendManager(backend_id)) {
    // ...
}
```

brylie:
@frenchbread good idea.

PR #383, opened 2015-08-18
Feature/first user admin

brylie:


brylie:
@apinf/developers please review ASAP, so we can include in the beta release.

PR #448, opened 2015-09-09
Hotfix/links to social services

frenchbread:
Hey @apinf/developers, this is a hotfix with updated links to social services. I've also added `target="_blank"` attribute to `<a>` tag, so that in will open social service in new tab. Please review.


elnzv:
I think it should first go into develop, not master.

brylie:
Right. Good catch @elnzv. By the way, which branch are we deploying at apinf.io?

frenchbread:
@elnzv @bajiat Thanks for noticing! Closing this PR, new one is #449

PR #516, opened 2015-10-07
Move global schema definitions to attributes of the Schemas object

### LINKED ISSUE #497: Move all schema definitions under the global Schemas object
We are currently defining our schemas as global variables. This causes unnecessary clutter of our global namespace. Move all schema definitions to attributes of the global Schemas object.

Task
---
* [x] Make list of all project schema definitions
* [x] Move each schema definition to an attribute of the Schemas object
* [x] Update the collection-schema attachment lines to use the global Schemas object/attribute

mauriciovieira:
closes #497


brylie:
This looks good. Are all schemas updated now?

We can ignore the linthub errors, because linthub seems to not detect our `.jshintrc` file.

PR #527, opened 2015-10-13
Add error treatment to apiUmbrellaWeb[...].createApiBackend - closes #357

### LINKED ISSUE #357: Inform user of API Umbrella errors when submitting API Backends form
When a user submits the API Backend form, API Umbrella sometimes returns errors to the server. When the server receives an error, the user needs to be informed, as their API Backend may not be properly submitted to API Umbrella. One specific example is that API Umbrella will not accept a backend submission when filling in the `backend_host` field with a domain that does not exist.

# Task

* [x] Use the sAlert package or similar to display errors to the end user. 
* [x] Ensure that the API Backend is not inserted into the apiBackends collection, if there are API Umbrella errors in the response object.

mauriciovieira:
@elnzv or @frenchbread please :mag:

### DIFF
--- a/client/views/api_backends/autoform.js
+++ b/client/views/api_backends/autoform.js
@@ -1,5 +1,13 @@
 AutoForm.hooks({
   apiBackends: {
+    beginSubmit: function () {
+      // Disable form elements while submitting form
+      $('[data-schema-key], button').attr("disabled", "disabled");
+    },
+    endSubmit: function () {
+      // Enable form elements after form submission
+      $('[data-schema-key], button').removeAttr("disabled");
+    },
     onSuccess: function (formType, apiBackendId) {
       // Send the API Backend to API Umbrella
       response = Meteor.call('createApiBackendOnApiUmbrella', apiBackendId, function(error, apiUmbrellaWebResponse) {
@@ -33,9 +41,6 @@ AutoForm.hooks({
             sAlert.error(error, {timeout: 'none'});
             // Figure out a way to send the errors back to the autoform fields, as if it were client validation.
           });
-
-          //Remove apiBackendId from local mongoDB
-          ApiBackends.remove({_id: apiBackendId});
        }
       });
     }
--- a/client/views/api_backends/autoform.js
+++ b/client/views/api_backends/autoform.js
@@ -1,10 +1,43 @@
 AutoForm.hooks({
   apiBackends: {
-    onSuccess: function (formType, backendId) {
+    onSuccess: function (formType, apiBackendId) {
       // Send the API Backend to API Umbrella
-      Meteor.call('createApiBackendOnApiUmbrella', backendId);
-      // Redirect to the just created API Backend page
-      Router.go('viewApiBackend', {_id: backendId});
+      response = Meteor.call('createApiBackendOnApiUmbrella', apiBackendId, function(error, apiUmbrellaWebResponse) {
+
+        //apiUmbrellaWebResponse contents
+        // apiUmbrellaWebResponse = {
+        //   result: {},
+        //   http_status: 200,
+        //   errors: {}
+        // };
+
+        if (apiUmbrellaWebResponse.http_status === 200) {
+          //Redirect to the just created API Backend page
+          Router.go('viewApiBackend', {_id: apiBackendId});
+        } else {
+          // ex 1:
+          // {"default":'{"backend_protocol":["is not included in the list"]}}'
+          // ex 2:
+          // {"errors":{"frontend_host":["must be in the format of \"example.com\""],
+          //            "backend_host":["must be in the format of \"example.com\""],
+          //            "base":["must have at least one url_matches"],
+          //            "servers[0].host":["must be in the format of \"example.com\"","Could not resolve host: no address for http://api.example.com"],
+          //            "servers[0].port":["can't be blank","is not included in the list"]}
+          // }
+
+          var errors = _.values(apiUmbrellaWebResponse.errors);
+          errors = _.flatten(errors);
+          _.each(errors, function(error) {
+            console.log(error);
+            //Display error to the user
+            sAlert.error(error);
+            // Figure out a way to send the errors back to the autoform fields, as if it were client validation.
+          });
+
+          //Remove apiBackendId from local mongoDB
+          ApiBackends.remove({_id: apiBackendId});
+       }
+      });
     }
   }
 });
--- a/client/views/api_backends/autoform.js
+++ b/client/views/api_backends/autoform.js
@@ -11,6 +11,9 @@ AutoForm.hooks({
     before: {
       // Replace `formType` with the form `type` attribute to which this hook applies
       insert: function (apiBackendForm) {
+        // Keep the context to use inside the callback function
+        context = this;
+
         // Send the API Backend to API Umbrella
         response = Meteor.call('createApiBackendOnApiUmbrella', apiBackendForm, function(error, apiUmbrellaWebResponse) {
 
@@ -23,10 +26,7 @@ AutoForm.hooks({
 
           if (apiUmbrellaWebResponse.http_status === 200) {
             //Return asynchronously
-            this.result(apiBackendForm);
-
-            //Redirect to the just created API Backend page
-            Router.go('viewApiBackend', {_id: apiBackendId});
+            context.result(apiBackendForm);
           } else {
             // ex 1:
             // {"default":'{"backend_protocol":["is not included in the list"]}}'
@@ -48,10 +48,14 @@ AutoForm.hooks({
             });
 
             //Return error asynchronously
-            this.result(false);
+            context.result(false);
          }
         });
       }
     },
+    onSuccess: function (formType, apiBackendId) {
+      //Redirect to the just created API Backend page
+      Router.go('viewApiBackend', {_id: apiBackendId});
+    }
   }
 });
--- a/client/views/api_backends/autoform.js
+++ b/client/views/api_backends/autoform.js
@@ -8,41 +8,50 @@ AutoForm.hooks({
       // Enable form elements after form submission
       $('[data-schema-key], button').removeAttr("disabled");
     },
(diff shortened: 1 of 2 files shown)


elnzv:
1. Error message disappears too fast, extend time of showing or leave message until user closes it.
2. It is not clear that there is a check after submission - form is cleaned and only after couple seconds error message is showing. I think we need to put a loading icon or some indication that the form is submitting.
3. If there is an error - form is cleaned, I think we need to keep it and just indicate the field which causing the error.

If I get an error, server console says this:

screenshot from 2015-10-16 12 37 15

What is this [object Object]?

mauriciovieira:
1. Ok, I'll do this.
2. This happens because first autoform inserts the new API to meteorjs mongo db and then we try to insert it in the API Umbrella through _brylie:api-umbrella_ rest-client package. I agree with the loading icon and will do this ASAP.
3. Yeah, I didn't not research how to send the errors back to the client, _manually_, without the autoform magic. I hope it is not impossible, or we would have to abdicate the use of autoform in this complex screen.
4. I removed that unuseful _Return response ..._ log.

elnzv:
@mauriciovieira, let me know when PR is ready to be reviewed by removing WIP label.

mauriciovieira:
@bajiat I am trying to fix this as @elnzv commented.

client/views/api_backends/autoform.js
```diff
@@ -1,10 +1,43 @@
 AutoForm.hooks({
   apiBackends: {
-    onSuccess: function (formType, backendId) {
+    onSuccess: function (formType, apiBackendId) {
```
brylie: We might need to change this to use `before` or `onSubmit` [AutoForm hooks](https://github.com/aldeed/meteor-autoform#callbackshooks), so that we can validate the API Umbrella response.

client/views/api_backends/autoform.js
```diff
@@ -1,10 +1,43 @@
...
+          //            "servers[0].host":["must be in the format of \"example.com\"","Could not resolve host: no address for http://api.example.com"],
+          //            "servers[0].port":["can't be blank","is not included in the list"]}
+          // }
+
+          var errors = _.values(apiUmbrellaWebResponse.errors);
+          errors = _.flatten(errors);
+          _.each(errors, function(error) {
+            console.log(error);
+            //Display error to the user
+            sAlert.error(error);
```
brylie: You can keep the sAlert box visible by passing in a configuration object with `timeout` set to `none`:

```
sAlert.error(error, {timeout: 'none'});
```

client/views/api_backends/autoform.js
```diff
@@ -1,10 +1,43 @@
...
+            // Figure out a way to send the errors back to the autoform fields, as if it were client validation.
+          });
+
+          //Remove apiBackendId from local mongoDB
+          ApiBackends.remove({_id: apiBackendId});
```
brylie: Rather than removing the backend, lets keep the user at the Create API Backend form, so that they can fix errors in the form and resubmit.

elnzv:
@mauriciovieira here is one approach to show that form is submitting:

```js

AutoForm.hooks({
  apiBackends: {
    beginSubmit: function () {
      // Disable form elements while submitting form
      $('[data-schema-key],button').attr("disabled", "disabled");
    },
    endSubmit: function () {
      // Enable form elements after form submission
      $('[data-schema-key],button').removeAttr("disabled");
    }
  }
});

```
It simply disables all field during the submission.

mauriciovieira:
@elnzv @brylie thank you guys for the tips. I already added some fixes to the code.

client/views/api_backends/autoform.js
mauriciovieira: The problem here is that I am afraid we might not have the apiBackend saved in mongoDB prior sending it to api umbrella. I will try this

client/views/api_backends/autoform.js
mauriciovieira: Since what we have at this point is a invalid apiBackend in the ApiBackends collection, what I thought was to remove it. But the best approach is in fact not saving it and then only save if api-umbrella answered with no errors.

brylie:
I note that there are two objects in you examples:

```
{
  default: {} 
}
```

```
{
  errors: {} 
}
```

What is the difference that causes the object to sometimes be named `default` and other times be named `errors`?

mauriciovieira:
@brylie, it is not:
```
{
  default: {} //object
}
```

but

```
{
  default: 'string with errors' 
}
```

That means, all we get from ```apiUmbrellaWeb.adminApi.v1.apiBackends.createApiBackend(constructedBackend);``` is a string, the error.message. As it is not treating the error and returning a json object, but only a string while [brylie:api-umbrella issue#1](https://github.com/brylie/meteor-api-umbrella/issues/1) is not resolved. After it is resolved, we might get a json object 

```
{
  errors: {} 
}
```

with the errors for each.

So, the format 

```
{
  errors: {} 
}
```

is never returned untill we fix the meteor-package error handling.

brylie:
@elnzv please review again. @mauriciovieira and I did pair programming over the past two days to get things working.

client/views/api_backends/autoform.js
```diff
@@ -1,10 +1,61 @@
...
+    beginSubmit: function () {
+      // Disable form elements while submitting form
+      $('[data-schema-key], button').attr("disabled", "disabled");
+    },
+    endSubmit: function () {
+      // Enable form elements after form submission
+      $('[data-schema-key], button').removeAttr("disabled");
+    },
+    before: {
+      // Replace `formType` with the form `type` attribute to which this hook applies
```
brylie: Cleanup comment

client/views/api_backends/autoform.js
```diff
@@ -1,10 +1,61 @@
...
+      insert: function (apiBackendForm) {
+        // Keep the context to use inside the callback function
+        context = this;
```
brylie: Add `var` to the beginning of this declaration.

client/views/api_backends/autoform.js
```diff
@@ -1,10 +1,61 @@
...
+
+          //apiUmbrellaWebResponse contents
+          // apiUmbrellaWebResponse = {
+          //   result: {},
+          //   http_status: 200,
+          //   errors: {}
+          // };
+
+          if (apiUmbrellaWebResponse.http_status === 200) {
+            //Return asynchronously
```
brylie: Give a more meaningful comment here. E.g. 'submit the form on success'

client/views/api_backends/autoform.js
```diff
@@ -1,10 +1,61 @@
...
+            // {"default":'{"backend_protocol":["is not included in the list"]}}'
+            // ex 2:
+            // {"errors":{"frontend_host":["must be in the format of \"example.com\""],
+            //            "backend_host":["must be in the format of \"example.com\""],
+            //            "base":["must have at least one url_matches"],
+            //            "servers[0].host":["must be in the format of \"example.com\"","Could not resolve host: no address for http://api.example.com"],
+            //            "servers[0].port":["can't be blank","is not included in the list"]}
+            // }
+
+            var errors = _.values(apiUmbrellaWebResponse.errors);
```
brylie: Add a descriptive comment above this line. What is `errors`? What action does the code perform in plain language?

client/views/api_backends/autoform.js
```diff
@@ -1,10 +1,61 @@
...
+            errors = _.flatten(errors);
+            _.each(errors, function(error) {
+              console.log(error);
```
brylie: Clean up console log, unless needed.

client/views/api_backends/autoform.js
```diff
@@ -1,10 +1,61 @@
...
+              //Display error to the user, keep the sAlert box visible.
+              sAlert.error(error, {timeout: 'none'});
+              // Figure out a way to send the errors back to the autoform fields, as if it were client validation.
```
brylie: Add 'TODO:' to beginning of comment line

client/views/api_backends/autoform.js
```diff
@@ -1,10 +1,61 @@
...
+            });
+
+            //Return error asynchronously
```
brylie: Give a more meaningful comment here. E.g. 'Cancel form submission on error, so user can edit fields'

server/methods/apiBackends.js
```diff
@@ -17,19 +17,45 @@ Meteor.methods({
       });
     };
   },
-  'createApiBackendOnApiUmbrella': function (apiBackendId) {
-    console.log('Submitting Backend to API Umbrella.')
-    // Get the API Backend object
-    var apiBackend = ApiBackends.findOne(apiBackendId);
+  createApiBackendOnApiUmbrella: function (apiBackendForm) {
+    console.log('Submitting Backend to API Umbrella.');
```
brylie: Perhaps we don't need this console.log any more. Cleanup, if unneeded.

server/methods/apiBackends.js
```diff
@@ -17,19 +17,45 @@ Meteor.methods({
...
 
     // Construct an API Backend object for API Umbrella with one 'api' key
     var constructedBackend = {
-      "api": apiBackend
+      "api": apiBackendForm
+    };
+
+    var apiUmbrellaWebResponse = {
```
brylie: Add comment describing the purpose and structure of this object.

server/methods/apiBackends.js
```diff
@@ -17,19 +17,45 @@ Meteor.methods({
...
+      result: {},
+      http_status: 200,
+      errors: {}
     };
 
     // Send the API Backend to API Umbrella
-    var response = apiUmbrellaWeb.adminApi.v1.apiBackends.createApiBackend(constructedBackend);
+    try {
+      apiUmbrellaWebResponse.result = apiUmbrellaWeb.adminApi.v1.apiBackends.createApiBackend(constructedBackend);
```
brylie: Add short, descriptive comment here. This line is long and hard to read, so a comment will help quickly clarify.

Issue #517, opened 2015-10-08
Use API Umbrella for REST calls made through Swagger UI


bajiat:
When making API calls in documentation browser, i.e. Swagger UI, it is desirable to  use API Umbrella proxy, e.g. umbrella.apinf.io, for requests.

The API Umbrella proxy URL can be found in config.json (as api_umbrella.base_url)

One approach to this task would be to rewrite the .swagger file, changing the `host` property to match the `api_umbrella.base_url`.

# Task
* [x] Determine how to change .swagger file to use api_umbrella.base_url for the API Base URL
  * [x] Specifically, the `host` in the .swagger file should be set to the same value as  `api_umbrella.base_url`
* [x] Test that API Calls through Swagger UI are being sent to API Umbrella

# Reference
Check the [Swagger specification](http://swagger.io/specification/) for information about the `host` property.

# Related
Related to #76.

Co-pilot: @elnzv

bajiat:
@elnzv How about this one - would you be interested? This is a continuation of task #76, which is being implemented by @frenchbread. Please estimate the size if you can take it.

elnzv:
Will estimate the size after discussion with @frenchbread, since he was working on the related issue.

brylie:
@frenchbread status report?

frenchbread:
@brylie We kind of stuck with this one. 

With @elnzv we found out that the only way to change the proxy to apiUmbrella is to download swagger.json file itself, change the host and base path values to be that we want and then use the same file for future requests in Swagger UI base url field.

brylie:
Alright, that makes sense. Should we get together to plan this feature a bit? We can all three have a planning meeting later this sprint. @elnzv @frenchbread

bajiat:
@elnzv @frenchbread @brylie I talked about this issue with @kyyberi as
well. If I understood correctly, he confirmed the approach @elnzv and
@frenchbread had figured out, i.e. changing the host in swagger.json.

2015-10-26 9:53 GMT+02:00 Brylie Christopher Oxley <notifications@github.com
>:

> Alright, that makes sense. Should we get together to plan this feature a
> bit? We can all three have a planning meeting later this sprint. @elnzv
> <https://github.com/elnzv> @frenchbread <https://github.com/frenchbread>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/apinf/api-umbrella-dashboard/issues/517#issuecomment-151052156>
> .
>



-- 
---------
[...]

PR #534, opened 2015-10-19
Feature/pass api key to swagger - closes #76

### LINKED ISSUE #76: Create API testing functionality in documentation browser
We have integrated the Swagger UI, which includes sandbox functionality. The functionality should be improved in order to use the actual API key and the API Umbrella proxy.

The Swagger UI is embedded in the `/documentation` template and has two fields at the top:
swaggerui-urlwithapi_key

In the Swagger UI, we want to populate the `api_key` field with the current user's API key.

# User Story
```
 As Minh (API consumer), I want to test API calls in a sandbox.
```

# Definition of done

* [x] Determine how to modify the [Swagger UI](http://swagger.io/swagger-ui/) API testing functionality.
  * [x] Modifications to Swagger UI may be unnecessary, if we can pass arguments in to the iframe
* [x] Pass the APInf user API key to Swagger UI sandbox. (e.g. via an iframe argument or some other method)
* [ ] If possible, ensure the Swagger UI uses API Umbrella proxy, e.g. umbrella.apinf.io, for requests.
  * The API Umbrella proxy URL can be found in [config.json](https://github.com/apinf/api-umbrella-dashboard#settingsjson) (as `api_umbrella.base_url`)

frenchbread:


frenchbread:
#76 related issue. @apinf/developers Please review

client/views/documentation/view/documentation.js
```diff
@@ -29,6 +29,17 @@ Template.documentation.events({
 
     // Set Swagger document URL to selected value
     instance.swaggerDocumentUrl.set(event.target.value);
+  },
+  'load #documentation': function () {
```
mauriciovieira: We should try to adhere to some conventions, like use ' or " to name the events.
As the other events and helpers are using ", I suggest replacing ' for ".

client/views/documentation/view/documentation.js
```diff
@@ -29,6 +29,17 @@ Template.documentation.events({
...
+
+    // gets current users object
+    var user = Meteor.user();
+
+    // gets current user's apiKey
+    var apiKey = user.profile.apiKey;
+
+    // passes api key to iFrame once it is loaded
+    $('#documentation').contents().find('#input_apiKey').val(apiKey);
```
mauriciovieira: We should try to adhere to some conventions, like use ' or " to findthe selectors.
As the other selectors are using ", I suggest replacing ' for ".

 Compare with https://github.com/apinf/api-umbrella-dashboard/blob/develop/client/views/documentation/view/documentation.js#L18

mauriciovieira:
After these tiny changes, :+1:, feel free to merge.

frenchbread:
@mauriciovieira Thanks!

brylie:
@mauriciovieira is there any way I can help here?

mauriciovieira:
@brylie I think it is done

frenchbread:
@brylie Actually I came up with better solution for this issue after stated working on #517. 

In `public/swagger/index.html` there are commented out lines:
```
// if you have an apiKey you would like to pre-populate on the page for demonstration purposes...
 /*
    var apiKey = "myApiKeyXXXX123456789";
    $('#input_apiKey').val(apiKey);
*/
```
This would be the best place to preload user's api key. The thing is that Swagger UI is currently loaded using iframe in `client/views/documentation/view/documentation.html`
```
<iframe src="/swagger/index.html?url={{ swaggerDocumentUrl }}" width="100%" id="documentation"></iframe>
```
Meaning that file `index.html` which I mentioned above is not used at all, so approach that I am suggesting is not suitable.

In addition, it would be also possible to change `path` (e.g. `url`) to swagger.json file (line 30 in `public/swagger/index.html`)

PR #546, opened 2015-10-27
Feature/466 copy to clipboard btn - closes #466

### LINKED ISSUE #466: Add a "Copy" button for API key
Figure out how make copying the API key to clipboard easier than having to highlight the key and then press Ctrl+C or use other methods for copying (context menu etc.) One option would be adding a "Copy" button next to API key on the Profile page.

frenchbread:


frenchbread:
@apinf/developers Please review. #466 is related.

client/views/profile/profile.js
```diff
@@ -8,3 +8,31 @@ AutoForm.hooks({
     }
   }
 });
+
+Template.profile.rendered = function () {
+
+  // initializes button
+  var copyButton = $("<a class=\"btn btn-default btn-xs\" id=\"copyApi\"> Copy API to clipboard</a>");
```
brylie: Why are we creating/appending the copy button in the JS? Is it conditionally loaded?

client/views/profile/profile.less
```diff
@@ -0,0 +1,3 @@
+#copyApi{
+  margin-top:10px;
```
brylie: Please use relative values; em or rem.

frenchbread:
@elnzv Thanks for fix.

PR #552, opened 2015-10-28
Add waffle analytics; development status section

brylie:


brylie:
@apinf/developers please review before the 0.1 release.

PR #562, opened 2015-10-30
fixed block position

elnzv:
simple fix to profile block position


jykae:
@elnzv Was this small hotfix for layout because upcoming cool settings page? 
Works for me. Merging and closing.

PR #590, opened 2015-11-04
Feature/517 API calls thorough umbrella proxy - closes #517

### LINKED ISSUE #517: Use API Umbrella for REST calls made through Swagger UI

frenchbread:
@apinf/developers Please review. #517 related.

#### The approach is
1. Read the text file when selected.
2. Convert text data to JSON object to be able to refer to its `values` by `key`.
3. Modify values of `host` and `basePath` to what we need.
4. Convert object back to JSON string.
5. Construct `new File( )` using updated file data. (This is implemented in two steps: First constructing `new Blob( )`. Second: `new File( )` that takes `Blob` as parameter)
6. Saving file in Mongodb collection

#### Issues
* Currently custom values of `basePath` and `host` are hardcoded. Since all the above process is happening on a client-side, there is no way to access Meteor.settings data.

#### Testing 
1. Create new `apiBackend`. While this process, upload new documentation file e.g swagger.json. (For example default one: http://petstore.swagger.io/v2/swagger.json)
2. Go to documentation viewer page.
3. In drop down manu select uploaded file.
4. iFrame will give an error since the remote swagger UI was not able to access local file. 
5. The path to documentation file is provided in there. 
6. Copy this path to a new tab to see updated `host` and `basePath` values.


bajiat:
@frenchbread Is this still work in progress or should the PR be assigned?

frenchbread:
@bajiat Yes, this it is still in progress.

frenchbread:
Currently getting an error when running meteor:
```
Error: Invalid characters. Only 0-9 characters are allowed.
```

@apinf/developers any ideas?

brylie:
This comes from the [`syncApiBackends` server method](https://github.com/apinf/api-umbrella-dashboard/blob/develop/server/methods/apiBackends.js#L2). 

Basically, an API Backend was entered directly in API Umbrella with zero for the port number. When our system tries to import that backend, port zero doesn't pass [our validation (which looks for a port with two to five numeric characters)](https://github.com/apinf/api-umbrella-dashboard/blob/develop/both/collections/backend.js#L4). We (@elnzv) are going to remove the API Backend from API Umbrella and loosen up our RegEx validation. 

You can take any temporary measures to continue working on this task, including commenting out the [line with the db insert](https://github.com/apinf/api-umbrella-dashboard/blob/develop/server/methods/apiBackends.js#L15).

client/views/api_backends/import_documentation/import.js
```diff
@@ -1,9 +1,46 @@
 Template.importApiDocumentation.events({
   'change #apiDocumentationFile': function (event/*, template*/) {
-    // Get the submitted file
+
+    var customProxy = "umbrella.apinf.io";
```
brylie: This shouldn't be hard-coded, as other organizations may deploy Apinf. This `customProxy` variable needs to come from `Meteor.settings.apiUmbrella.baseUrl`.

client/views/api_backends/import_documentation/import.js
```diff
@@ -1,9 +1,46 @@
...
+      var documentationFile = JSON.stringify(dataObj);
+
+      // Creates blob object with content type of JSON
+      var blobData = new Blob([documentationFile], {type: "application/json"});
+
+      // Creates file object from Blob object
+      var updatedFile = new File([blobData], file.name, {type: "application/json"});
+
+      // Insert into filesystem collection
+      ApiDocumentation.insert(updatedFile);
```
brylie: We need to keep the original file in-tact. Can we override the `host` and `basePath` on the client before rendering in Swagger UI? This would be better, and would allow the variables to change over time, e.g. if the deployment environment changes.

client/views/api_backends/import_documentation/import.js
frenchbread: @brylie `Meteor.settings...` is not available on a client side. Any suggestions to that?

client/views/api_backends/import_documentation/import.js
frenchbread: Working on it.

frenchbread:
@brylie Having trouble to fetch a file on route callback:

```javascript
JsonRoutes.add("get", "cfs/files/apiDocumentation/:id/:file", function (req, res, next) {

  var id = req.params.id;

  var b = [];

  var doc = ApiDocumentation.findOne({_id: id});
  
  var readStream = doc.createReadStream();
  
  var buffer = new Buffer(0);
  
  readStream.on('readable', function() {
    buffer = Buffer.concat([buffer, readStream.read()]);
  });
  
  readStream.on('end', function() {
    console.log(buffer.toString('base64'));
    b.push(buffer.toString('base64'));
  });

  JsonRoutes.sendResult(res, 200, b );
});

```

Found a solution to read binary data of file stored in db with collectionFS [here](https://github.com/CollectionFS/Meteor-CollectionFS/issues/457). Though this solution did not work for me. Any suggestions how to read file from CollectionFS on a server-side?

frenchbread:
After hours of debugging and reading tones of [docs](https://github.com/CollectionFS/Meteor-CollectionFS/blob/b46cb20d2d86a806e19e9ffbd72955c2a4ae37e7/packages/file/api.md) turned out that for some reason `doc.createReadStream()` come as undefined..

frenchbread:
After updating cfs:* packages `.createReadStream()` is available.

Still not able to read the file.

brylie:
Make sure your JsonRoute code is running on the server. Also, console log doc to make sure it is a FS object.

frenchbread:
It is an FS object. Double checked that. And also JsonRoute on the server side. I am able to log anything to server console but file. It comes as an empty string.

brylie:
Damir, one more idea. Store the API doc in a regular mongo collection, like you were thinking previously. I.e. parse the Swagger into an object and insert it into the collection (no schema necessary). 

Then, in your JsonRoute you just return a db cursor (`Collection.find(id)`) or JS object (`fetch ()`)-- exactly like the example code.

brylie:
It is not essential to keep the Swahher file. We just need to preserve the original Swagger object structure.

We can then override the `baseUrl` on the fly like we are trying.

frenchbread:
So basically we potentially can get rid of FS and store the actual Swagger object in the mongo collection.

Then, in JsonRoute, fetch the file, modify it & return. Correct?

brylie:
Jep. Make sure to create a 'foreign key' field, such as `apiBackendId`, when storing the swagger document in the `ApiDocs` collection (or some similar name).

frenchbread:
Ok. Thanks!

frenchbread:
Huh. 
Changes are pushed. Seems like everything works as expected now with clear code implementation and logic. Bonus check for protocol in `base_url` added. :smile: 

ToDo:
1. Store foreign key for ApiDoc in ApiBackend document. (Issue: Since templates are different for documentation file upload & apiBackend submit, probably have to wait until related issue about redesigning `addApiBackend page` is implemented)
2. Provide ApiDoc url to Select box at `viewApiBackend` page.

@brylie @elnzv @bajiat

brylie:
Make sure to use camel case now:`baseUrl` instead of `base_url`. This has changed recently.

client/views/api_backends/import_documentation/import.js
```diff
@@ -1,9 +1,91 @@
...
+              ApiDocs.insert(doc);
+
+            } else {
+
+              // Notifies user if not able to parse the file either as JSON or YAML objects.
+              FlashMessages.sendError("Error when reading the file. Please check file contents for errors.");
+
+            }
+          }
+        }else{
```
brylie: Use consistent whitespace. Should be `} else {` instead of `}else{`.

client/views/api_backends/import_documentation/import.js
```diff
@@ -1,9 +1,91 @@
...
+            if (JSON.parse(importedFile) || jsyaml.safeLoad(importedFile)) {
+
+              var doc = {};
+
+              // Checks file's extension for its secure conversion to JSON object
+              if (instance.endsWith(fileName, ['json'])){
+
+                // Converts JSON string to JSON object
+                doc = JSON.parse(importedFile);
+              } else if (instance.endsWith(fileName, ['yaml', 'yml'])){
```
brylie: Use consistent whitespace. Should be a space between the `if ( ... )` and `{`.

client/views/api_backends/import_documentation/import.js
```diff
@@ -1,9 +1,91 @@
...
+          // Notifies user if file extension provided is not supported
+          FlashMessages.sendError("Only .json and .yaml(.yml) files are supported.");
+        }
+      }
+    });
   }
 });
+
+Template.importApiDocumentation.created = function () {
+
```
brylie: Add line comment. 'Get reference to template instance'.

client/views/api_backends/import_documentation/import.js
```diff
@@ -1,9 +1,91 @@
...
+  var instance = this;
+
+  // function attached to template instance checks file extension
+  instance.endsWith = function (filename, suffixList) {
```
brylie: Rename this to be more semantic: e.g. `fileEndsWith`

lib/router/jsonRoutes.js
```diff
@@ -0,0 +1,31 @@
+JsonRoutes.add("get", "files/apiDoc/:id", function (req, res, next) {
```
brylie: Consider the route `api/:_id/swagger.json` to be consistent with our other [API route(s)](https://github.com/apinf/api-umbrella-dashboard/blob/develop/lib/router/main.js#L60).

lib/router/jsonRoutes.js
```diff
@@ -0,0 +1,31 @@
...
+
+  var baseURL = Meteor.settings.api_umbrella.base_url;
```
brylie: This should use camel case now:

`Meteor.settings.apiUmbrella.baseUrl;`

lib/router/jsonRoutes.js
```diff
@@ -0,0 +1,31 @@
...
+
+  var clearedString = "";
+
+  // Checks what protocol is in the string and removes one from a string
+  if(baseURL.indexOf('https://') >= 0){
```
brylie: Consider using a [Regular Expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions?redirectlocale=en-US&redirectslug=JavaScript%2FGuide%2FRegular_Expressions) (Regex) here to [select the latter part of the URL](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions?redirectlocale=en-US&redirectslug=JavaScript%2FGuide%2FRegular_Expressions#Using_parenthesized_substring_matches). E.g. the Regex would look for a pattern matching `http://` or `https://` and grab the URL string after that pattern.

brylie: Try the [URL.js package](https://atmospherejs.com/olragon/uri-js) to simplify this code.

lib/router/jsonRoutes.js
```diff
@@ -0,0 +1,31 @@
...
+  } else if (baseURL.indexOf('http://') >= 0){
+
+    clearedString = baseURL.replace('http://','');
+  }
+  
+  var splittedValues = clearedString.split('/');
+  var customHost = splittedValues[0];
+  var customBasePath = '/' + splittedValues[1];
+
+  var id = req.params.id;
```
brylie: Spell out the word `request`.

lib/router/jsonRoutes.js
brylie: Go ahead and use the full words `request` and `response`.

lib/router/jsonRoutes.js
```diff
@@ -0,0 +1,31 @@
...
+
+  var ApiDoc = ApiDocs.findOne(id);
+
+  // Updates values to custom ones
+  ApiDoc.host = customHost;
+  ApiDoc.basePath = customBasePath;
+
+  // Sends result back to client
+  JsonRoutes.sendResult(res, 200, ApiDoc);
```
brylie: Use the full word `response` here.

client/views/api_backends/import_documentation/import.js
```diff
@@ -1,9 +1,91 @@
...
+
+    // variable that keeps state of is this filename contains provided extensions - false by default
+    var state = false;
+
+    // iterating through extensions passed into suffixList array
+    for (var i=0; i <suffixList.length; i++){
```
brylie: You can use the URI library [`suffix()`](http://medialize.github.io/URI.js/docs.html#accessors-suffix) method along with underscore.js [`contains()`](http://underscorejs.org/#contains) method:

```JavaScript
// List of allowed extensions
var allowedExtensions = ['yaml', 'yml', 'json'];

// Parse the file string to URI object
var file = new URI('file.json');

// Get the file extension
var fileExtension = url.suffix(); // returns 'json'

// Check if the file suffix is in the allowed extensions list
var extensionAllowed = _.contains(allowedExtensions, fileExtension);

// Do something if the extension is allowed
if (extensionAllowed) {
  // Do something
};
```

client/views/api_backends/import_documentation/import.js
```diff
@@ -1,9 +1,91 @@
...
+
+Template.importApiDocumentation.created = function () {
+
+  var instance = this;
+
+  // function attached to template instance checks file extension
+  instance.endsWith = function (filename, suffixList) {
+
+    // variable that keeps state of is this filename contains provided extensions - false by default
+    var state = false;
```
brylie: You can actually remove this 'state' variable, if you rewrite the function based on the example below.

client/views/api_backends/import_documentation/import.js
```diff
@@ -1,9 +1,91 @@
...
+
+      if (file) {
+
+        // Get file's name
+        var fileName = file.name;
+
+        // Allowed file extensions for API documentation file
+        var acceptedExtensions = ["yaml", "yml", "json"];
+
+        if (instance.endsWith(fileName, acceptedExtensions)) {
```
brylie: You can use the URI library [`suffix()`](http://medialize.github.io/URI.js/docs.html#accessors-suffix) method along with underscore.js [`contains()`](http://underscorejs.org/#contains) method:

```JavaScript
// List of allowed extensions
var allowedExtensions = ['yaml', 'yml', 'json'];

// Parse the file string to URI object
var file = new URI('file.json'); // Example file name

// Get the file extension
var fileExtension = file.suffix(); // returns 'json'

// Check if the file suffix is in the allowed extensions list
var extensionAllowed = _.contains(allowedExtensions, fileExtension);

// Do something if the extension is allowed
if (extensionAllowed) {
  // Do something
};
```

frenchbread:
Please review.

Good suggestion about  the `URI.js` library!

frenchbread:
Please review.

client/views/api_backends/import_documentation/import.js
```diff
@@ -1,9 +1,72 @@
...
+            // Gets file contents
+            var importedFile = event.target.result;
+
+            // Checks for correct JSON or YAML syntax in file contents
+            if (JSON.parse(importedFile) || jsyaml.safeLoad(importedFile)) {
+
+              var doc = {};
+
+              // Checks file's extension for its conversion to JSON object
+              if ((fileExtension == 'json')) {
```
brylie: Why do you use double parenthesis here? Also, can you use triple equals ('===') instead of double equals ('==')?

lib/router/jsonRoutes.js
```diff
@@ -0,0 +1,24 @@
+JsonRoutes.add("get", "api/:id/swagger.json", function (request, response, next) {
+
+  // Get basePath for apiUmbrella from Meteor settings file
+  var apiUmbrellaBaseUrl = Meteor.settings.apiUmbrella.baseUrl;
+
+  // Remove trailing slash
+  apiUmbrellaBaseUrl = apiUmbrellaBaseUrl.replace(/\/$/, '');
```
brylie: Would it work to use the `new URI(apiUmbrellaBaseUrl);` here, instead of stripping the trailing slash? E.g. is the trailing slash not allowing the URI object to be created?

frenchbread:
@brylie About trailing slash: 

If leave trailing slash, it will stack with remaining path that comes from swagger UI so it will be like: `http://host.com/basePath//some_queries..`.

If I'll remove trailing slash from setting.json, `meteor-api-umbrella` package will argue.

frenchbread:
@brylie if doing `new URI(...)`, it will return same url with trailing slash present..

lib/router/jsonRoutes.js
brylie: E.g., the following works:

```JavaScript
var testing = new URI('https://example.com');
testing.domain() //outputs 'example.com'  without trailing slash
```

frenchbread: @brylie This, does not:
```javascript
var testing = new URI('https://example.com/basepath/');
testing.path(); //  returns '/basepath/'
```

brylie: Ah, right. What fails if the basepath has leading and trailing slashes? E.g. `/api-umbrella/`

frenchbread: Only trailing. Leading slash is fine as long as `url.hostname()` returns pure hostname with no slashes.

frenchbread: I mean leading slash for `basepath` is present in `url` anyways

brylie: Is SwaggerUI adding a trailing slash to the `basePath`? The [specification docs](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md) only mention a leading slash:

> The base path on which the API is served, which is relative to the host. If it is not included, the API is served directly under the host. The value MUST start with a leading slash (/). The basePath does not support path templating.

brylie:
I made a small change to the file, as discussed earlier. 

```
// Get basePath for apiUmbrella from Meteor settings file
  var baseUrl = new URI(Meteor.settings.apiUmbrella.baseUrl);

// Fetch the ApiDoc ...

// Updates values to custom ones
  ApiDoc.host = baseUrl.hostname();
  ApiDoc.basePath = baseUrl.path();
```

When generating a `swagger.json` based on the `Meteor.settings.apiUmbrella.baseURl`, the output from the JsonRoute has a structure like this:

```JavaScript
...
  "host": "umbrella.apinf.io",
  "basePath": "/api-umbrella/",
...
```

Note the trailing slash in basePath. If this proves to be problematic, we can open a bug report.

brylie:
I will refactor the API Umbrella integration module so that it doesn't use trailing slash.

brylie:
I just reviewed the [URI.js docs](http://medialize.github.io/URI.js/docs.html) and found out there is a [`directory()`](http://medialize.github.io/URI.js/docs.html#accessors-directory) method that works here. I have pushed a quick change. Please test @frenchbread 

If this works, we can go ahead and merge the changes, otherwise we will revert to your working version and merge :-)

frenchbread:
@brylie Tested and it works, but now, if trailing slash will not be provided, then `basePath` is equal to `/` : 
```javascript
var str = new URI('https://sample.com/pathname/'); // with trailing slash
str.directory(); // prints '/pathname'

var str = new URI('https://sample.com/pathname/'); // without trailing slash
str.directory(); // prints '/'
```

brylie:
Right. This is tricky. OK, I'll revert my commits and merge.

Issue #585, opened 2015-11-03
Sign out from Accounts page shows error 'Data not found'


saralavanip:
# Steps to reproduce

1. Visit https://apinf.io
1. Login
1. Go to Account section
1. Sign out 

## Result
- Successful sign out with error: 'Data not found'  

## Screenshot

![signin_goto_profile_signout_data_not_found](https://cloud.githubusercontent.com/assets/13309358/10915847/40af4c44-8261-11e5-80a2-3a20c52177f3.png)

# Server console
Additionally, there is a traceback in the server console:

```
Exception while invoking method 'getChartData' TypeError: Cannot read property 'profile' of null
at [object Object].Meteor.methods.getChartData (server/methods/elasticSearch.js:26:1)
...
```

This suggests the user is being redirected to the dashboard page, so check the Router.onBeforeAction declarations. Also, the `getChartData` function needs to be refactored to handle exceptions (try ... catch).

# Related
There is a [related issue](https://github.com/apinf/api-umbrella-dashboard/issues/617) regarding exception handling.

bajiat:
Related to  #583.

bajiat:
Co-pilot: @elnzv

frenchbread:
PR - #608

brylie:
@frenchbread I have updated this issue description with console messages. Also note, the methods may need to handle exceptions.

saralavanip:
Hi,

Could you please let me know if the latest changes been deployed.


Regards
Sarala

brylie:
We will deploy changes next Monday or Tuesday, after we make an alpha release.

@bajiat please confirm.

Issue #583, opened 2015-11-03
Sign out from profile page shows error 'Data not found'


saralavanip:
Steps:   

1.Visit https://apinf.io
2.Sign in
3.Go to Profile section 
4.Sign out

- Successful sign out with error: Data not found

Screenshot:  

![signin_goto_profile_signout_data_not_found](https://cloud.githubusercontent.com/assets/13309358/10915655/5a2b449e-8260-11e5-8ede-ab9b1b643f01.png)

bajiat:
Related to  #585

bajiat:
Co-pilot: @elnzv

frenchbread:
PR - #608

Issue #582, opened 2015-11-03
'Data not found' error after sign out


saralavanip:
Steps:

1.Visit https://apinf.io
2.Go to sign in page
3.Enter email address
4.Enter  the password
5.Click on 'sign in ' button
6. Go to Menu and click on  'Catalogue'
7.Sign out

Expected: Successful sign out
Actual: Successful sign out with error message 'data not found'  

Screenshot:  
![signin_open_menu_page_signout_data_not_found_error](https://cloud.githubusercontent.com/assets/13309358/10915564/0c112152-8260-11e5-9e6a-a67a1887ac07.png)

bajiat:
Co-pilot: @elnzv

frenchbread:
PR - #608

PR #608, opened 2015-11-11
Hotfix for a bug 'Data not fould' closes - #582, #583, #585

frenchbread:
@apinf/developers  please review.

Removed nested alert() in chart dashboard. Now, if data was not found user is still notified but data returned from previous query is present.

screen shot 2015-11-11 at 14 32 03

### DIFF
--- a/client/views/dashboard/charts/charts.js
+++ b/client/views/dashboard/charts/charts.js
@@ -81,18 +81,21 @@ Template.chartsLayout.created = function () {
   instance.getDashboardData = function (input) {
 
     // calling method that returns data from elastic search
-    Meteor.call("getChartData", input, function (err, data) {
+    Meteor.call("getChartData", input, function (err, response) {
 
       // error checking
-      if (err) {
+      if (err || !response.isOk) {
+
+        // Gets custom error message from i18n
+        var errorMessage = TAPi18n.__("filterData-Title");
 
         // removing loading state once loaded
-        $('#loadingState').html("Data is not found! Please check your filtering options.");
+        $('#loadingState').html(errorMessage);
 
       } else {
 
         // gets to level in object with needed data
-        var dashboardData = data.hits.hits;
+        var dashboardData = response.searchResults.hits.hits;
 
         // parse the returned data for DC
         var parsedChartData = instance.parseChartData(dashboardData);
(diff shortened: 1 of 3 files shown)


client/views/dashboard/charts/charts.js
```diff
@@ -87,8 +87,7 @@ Template.chartsLayout.created = function () {
       if (err) {
 
         // removing loading state once loaded
-        $('#loadingState').html("Loaded");
-        alert("Data is not found!");
+        $('#loadingState').html("Data is not found! Please check your filtering options.");
```
brylie: We need an i18n string here, as this text is shown to the end-user.

frenchbread: @brylie How to access `en.i18n.json` file data from a file that you highlighted?

frenchbread: @brylie And as far as I understand there is either search results or an error returned from ES. So we don't have much context at this point.

client/views/dashboard/charts/charts.js
```diff
@@ -81,19 +81,27 @@ Template.chartsLayout.created = function () {
...
+
+      // response object:
+      // {
+      //    isOk: Boolean,
+      //    searchResults: Object
+      // }
 
       // error checking
-      if (err) {
+      if (err || !response.isOk) {
```
brylie: In your `getChartData` method, you can [raise an error](http://stackoverflow.com/a/20868067/1191545) rather than adding the `isOk` property. That way, you only have to check for the error here.

i18n/en.i18n.json
```diff
@@ -152,6 +152,8 @@
   "feedback-feedbackForm-description": "We welcome problem reports, feature ideas and general comments.",
 
   /*Charts*/
+  /*Error messages*/
+  "esData-notFound": "You have no API usage statistics. Please make some calls with your API Key.",
```
brylie: This appears to be a message for API Consumers, who have not used their API Key yet. What message would API Managers see when users have not called their API? How can we differentiate between these cases?

server/methods/elasticSearch.js
```diff
@@ -42,6 +44,23 @@ Meteor.methods({
...
+
+      searchResults = newSearch.doSearch();
+
+      // Changes "isOk" state to true if "try{}" didn't break before
+      isOk = true;
+
+    }catch(err){
+
+      // Providing empty object to be returned within "isOk:false" for consistency
+      searchResults = {};
```
brylie: Here is where you can [throw an error](http://stackoverflow.com/a/20868067/1191545), which would be caught in the client side by the callback function.

PR #660, opened 2015-11-27
Feature/unify api forms

### LINKED ISSUE #368: Unify the Edit API and Add API forms
There should be no need to maintain two separate instances of the API configuration form. Additionally, this can cause misalignment, if one of the forms gets updated. The forms should be unified with minimal conditional logic to change the form according to current form method (adding or editing).

Definition of done
---
* [x] List the differences between the two forms.
* [x] Choose one for the forms and rename it to API configuration form.
* [x] Add conditional logic to update form depending on context: adding or editing API.
  * [ ] UI text, e.g. title, button.
  * [x] Form action or method, e.g. insert or update.

elnzv:
Closes #368


elnzv:
@mauriciovieira and/or @frenchbread, could you please review this PR, since we were working on it together with Brylie?

jykae:
Will this fix issues https://github.com/apinf/api-umbrella-dashboard/issues/602 and https://github.com/apinf/api-umbrella-dashboard/issues/633 ?

frenchbread:
Found an issue: 
If current user does not have a permission to edit an apiBackend -> page counties to load and no error (e.g. 404) is given. 

If a backend, to which user is trying to navigate does not exist, error is also not given. Page infinitely continues to load.

@elnzv Could this one be fixed?

frenchbread:
Code looks good though!

elnzv:
@frenchbread, same issue in develop branch. 
I think we need to create a separate issue for this.

frenchbread:
Ok then. Merging this.

PR #678, opened 2015-12-02
Feature/elastic rest package update - closes #665

### LINKED ISSUE #665: Refactor elastic-rest package for meteor
## Research

- [x] Go through ES & read the docs.

## Refactor

- [x] Allow providing configuration object as an additional parameter for `ElasticRest()` class e.g.:
- [x] Add error handling.

```javascript
var config = {
    "host": "es_host"
};

var options = {
    "index": "index_value",
    "type": "type_value",
    "limit": 100,
    "query": "query_object",
    "fields": "fields_array"
};

new ElasticRest(config, options);
```

## Apinf

- [ ] Refactor `getChartsData` method to support updated API in `elastic-rest` package.

frenchbread:

#665 related issue.

Also have a look at  the [updated meteor package](https://github.com/frenchbread/meteor-elastic-rest).

#### Testing:
When testing, make sure that the package is updated & then open dashboard page. Analytics data should be there.

### DIFF
--- a/server/methods/elasticSearch.js
+++ b/server/methods/elasticSearch.js
@@ -1,50 +1,53 @@
 Meteor.methods({
   "getChartData": function (data) {
 
-    // initialise variables
+    // Initialise variables
     var loggedInUser, apiKey, query, searchResults;
 
-    // get user object
+    // Get user object
     loggedInUser = Meteor.user();
 
-    // get user role & check user role
+    // Get user role & check user role
     if (Roles.userIsInRole(loggedInUser, ['admin'])) {
 
-      // if admin - match_all
-      // construct query
+      // If current user is Admin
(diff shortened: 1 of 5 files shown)


server/methods/elasticSearch.js
```diff
@@ -1,50 +1,53 @@
...
 
-    // New instance of ElasticRest class with query as constructor
-    var newSearch = new ElasticRest(
-      data.index,
-      data.type,
-      data.limit,
-      query,
-      data.fields
-    );
+    var config = {
```
elnzv: Is it possible to define it in *server/startup.js* file?

frenchbread: @elnzv Why would you need that?

If you want you can change the host value to be fetched directly from settings collection rather than from settings.json file.

PR #682, opened 2015-12-04
Feature/admin settings page

### LINKED ISSUE #327: Create a settings page for Apinf configuration options
For organizations to host Apinf as a service, they need an easy way to configure the settings like mailgun. Create a settings page to hold current and future configuration options as well as a collection to store the configs.

# Research
* [x] Look for existing settings/ configuration pages in open-source Meteor projects
* [x] Look for configuration page examples in other web applications
* [x] Create a design document and list the solutions you found

# Design
* [x] Add a description of the configuration page to the design document
* [x] Create one or more wireframes to demonstrate the intended design (*do not skip this step*)

# Implement
* [x] Create a collection to store configuration settings
* [x] Create one or more templates to hold configuration forms
* [x] Create a schema to define configuration settings
* [x] Use AutoForm to build one or more configuration forms on the settings page(s)

Co-pilot: @brylie

elnzv:
Closes #327


elnzv:
PR is ready to be reviewed.

elnzv:
@jykae, please, review this PR ASAP.
Other PR's can be merged and create conflicts.

brylie:
We need to add checks for Github configuration. The server crashes on startup without a Github clientId:

```js
TypeError: Cannot read property 'clientId' of undefined
     at server/githubConfiguration.js:24:1
     at server/githubConfiguration.js:43:1
```

jykae:
Getting type errors to console are confusing when starting the application.
Also there are TODO's remaining in the code.

As for the Settings page, imho, and admin user experience, if it's possible to get current URL where application is running("Host") and for Umbrella settings, I would like enter only the Umbrella URL and API key/token information.

Possibly redirect to Settings page if we are getting undefined settings values in the startup.
However the requirements for this task are met. Could have been split to redesigning settings & implementing Settings page.

I will open up suggestion for redesigning settings.

jykae:
Cannot enter localhost with port.

nayttokuva 2015-12-17 kello 15 12 09

jykae:
Cannot merge this yet.

brylie:
Ah, interesting. We need to figure out if port is a requirement of this task.

brylie:
In fairness here, [we are using](https://github.com/apinf/api-umbrella-dashboard/pull/682/files#diff-1cc5fd9ba21916162a33ce0cd7bd13abR23) the [SimpleSchema URL RegEx](https://github.com/aldeed/meteor-simple-schema/blob/f9f2d6adb25d312494441666468cbf2cfb89b1f9/simple-schema.js#L547). URL regular expressions are typically complex, and might add undue complexity to this task.

Here are a couple of other possibilities we can pursue:
* File an upstream feature request that URL RegEx support port number
* Refactor the API Umbrella settings object to break the URL into components
  * base url
  * port
  * path

bajiat:
@jykae has promised to make the changes

brylie:
Cool. For a simple solution, we can just remove the URL Regex. However, then we are not guaranteed to have a valid URL, since there will be no validation.

jykae:
Looks like Illya already removed the validation regex after my prev. testing, so this seems to work now, as the validation is removed from URL. Though I had chat today with @brylie in Skype and we decided to fix it rather than remove the check.

jykae:
Minor enhancement issue opened: https://github.com/apinf/api-umbrella-dashboard/issues/734

brylie:
@jykae good catch. I didn't realize 'localhost' was causing the regex to fail.

PR #689, opened 2015-12-06
Fix for Issue/654

kyyberi:
Added "in ms" to UI where is now says "Response time"


bajiat:
@kyyberi Can you create a merge request into develop instead of master? (We can also do it, but it would great to have a contribution from you :))

kyyberi:
Ok, in the future I will. I was wondering which to use.

PR #705, opened 2015-12-09
Feature/701 no anonymous ratings. Closes #701

### LINKED ISSUE #701: Disallow anonymous API Backend ratings
As part of making our API Catalogue visible to the public, we have discovered that Anonymous users can vote for API Backends. Disable anonymous votes by adding an allow check for Meteor.userId(), and possibly adding a schema to validate API Backend votes.

brylie:


elnzv:
When anonymous user votes starts are still changing.
Even if results are not saved, visually it looks so.



I think visual part of voting should be disabled for anonymous user too.

brylie:
OK, I updated the code to be read only if user is not logged in. Good catch.

Issue #685, opened 2015-12-04
Implement keyword based search for frontpage


bajiat:
Related to #629 and #684.

frenchbread:
Interested.

frenchbread:
Definition of done as well as implementation updates are provided in issue #684.

[`working-branch`](https://github.com/apinf/api-umbrella-dashboard/compare/feature/search-apis)

bajiat:
To be closed. As @frenchbread commented, all progress can be seen from #684.

PR #719, opened 2015-12-16
Feature/search apis - closes #684

### LINKED ISSUE #684: Design keyword based search for frontpage
Related to issue #629.

# Definition of done

## Frontpage
* [x] There is a search field prominently visible on frontpage.
* [x] Search field contains meaningful helptext.
* [x] Search fields queries API backend collection.
  * [x] Query uses API backend `title` field.
  * [x] Query uses API backend `backend_host` field. - ?
* [x] All search query fields are indexed.
* [x] Search query is case insensitive.
* [x] Search query accepts arbitrary length string, which can contain spaces.

## Results page
* [x] Search results are displayed on a new page
* [x] Page has its own route.
  * [x] Page route contains the search keyword(s)
* [x] Page displays search query texts
* [x] Page displayes number of results
* [x] Page display search field (open items: does it contain the current search query)

## Search results
* [x] Individual results are listed separately.
* [x] Result contains title text.
  * [x] Title text displays prominently.
  * [x] Title text links to View API backend.

frenchbread:
Closes #684 & #685.


client/views/search/search.js
```diff
@@ -0,0 +1,120 @@
...
+
+  instance.autorun(function () {
+
+    // Check if search parameter is set
+    if (Router.current().params.query.q) {
+
+      var searchValue = Router.current().params.query.q;
+
+      // Update search field with current search value
+      $('#search-text').val(searchValue);
```
brylie: Interacting with DOM elements should happen after the DOM has rendered. I.e. in `Template.search.rendered`. You can define this `instance.autorun()` function inside the rendered callback.

brylie:
@frenchbread please review this PR. I have made several changes.

frenchbread:
@brylie Not able to do search now. Documents are not displayed. 
Also when submitting the search form on a search page the query value drops to empty (`/search?`). - For that reason I used two event handlers for processing a search: `keyup` and `submit`.

frenchbread:
@brylie  In addition, If you tested search functionality that was implemented previously, it was done in a way that URL's query value changed only when `submit` event occurs. Otherwise I did not see a reason for changing URL's query value on `keyup` event. We can keep this "reactive-query" feature which is awesomely done by `universe:reactive-queries` package if it won't affect the search core. For now it seems like it does.

At the same time, when the search was submitted in the header search form (being located on a `/search` page), the search text has been "mirrored" to the main search form. Now this functionality is lost.

brylie:
Are you logged in when searching? There is some strange behavior with the API Backends subscriptions. Anonymous users always have all API Backends availble on the search page.

brylie:
Part of the confusion here is due to having two search boxes:
* main layout top menu
* search page

The main lyout top menu necessarily changes the URL query string when the form is submitted (via Router.go). The search page search form does not need a submit event as there is no button to trigger the event ant pressing enter triggers a keyup event.

brylie:
We need to figure out what is happening with the API Backend subscriptions, so the search has data to query. We have to also use the query filter on the client side, since other sub-templates have additional active subscriptions (specifically the sidebar template).

brylie:
A simple fix here would be to make sure the client has subscribed to *all api backends with only required fields* when the search template renders. E.g.

```js
Meteor.publish("allApiBackendsForSearchPage", function () {
  // Get all API Backends, returning only fields required for search
  var apiBackends = ApiBackends.find({}, {fields: {name: 1, backend_host: 1 });

  return apiBackends;
});
```

Then, in the search page created callback, subscribe to the `apiBackendsForSearchPage` publication.

brylie:
I will try it with your previously created `searchApiBackends` publication.

brylie:
OK, I re-added the search subscription and things work for me locally. Please test again @frenchbread

brylie:
I also hid the navbar search form on the search page, since it was causing complexity with the URL parameters, etc. It may also have been confusing to people to have two search forms visible at once.

frenchbread:
Great job @brylie! Thanks for additions. Works as expected.

One thing that could be added to search feature later is that when the search query is empty, NOT to return all list of API Backends. Because if there would be lots of documents,  it is a mess to show them all on a one page. But this can be done in another sprint.

Merging.

brylie:
Good suggestion for the empty search query. Will you add it to our backlog as an enhancement issue @frenchbread?

jykae:
@bajiat Ok, looks like this got merged already.

frenchbread:
@brylie Done. Created issue #739.

PR #724, opened 2015-12-21
Feature/628 add backend wizard. Closes #628

### LINKED ISSUE #628: Re-design adding new API UI
Implement adding new API backend based on the workflow and design described in the wireframes below.

>I was totally lost with the UI when I was adding new API to service. I'm no UI designer nor graphics designer, but I decided to give it a shot and sketch new UI. The idea is to force API owner give information is smaller patches. This is highly simplified model and does not contain all the flavours which should be added if this path is selected. 

Definition of done
---
* [x] Create wireframes for the user flow.
* [x] Create flow diagram combining the wireframes.
* [x] Choose how to implement tabbed flow UI.
* [x] Determine how to track the state (current tab/step).
* [x] Pre-populate fields based on user feedback from user input.
* [x] Allow users to go back to previous steps by choosing the step.
* [x] When a user completes a step, toggle active state to subsequent step.
* [x] Validate each field.
* [x] Ensure that the submission to API Umbrella is successful.

# Stages
## Add Base information 
When API owner clicks add new API backend, the following will be shown. 
In the view API owner can see sections, which are expected to be filled. In the below phase "Base" is active, rest are links too but in purpose a little in the background. Owner gives 
* API name, long format
* API slug, short one word. It will be used as "prefix" in umbrella to differentiate it from rest of the APIS
* Description of the API, around 500 chars, not the whole documentation. 

Perhaps there should be a button "next" or something below the last field and from there API owner is taken to next phase. The owner can always switch the phase from top bar. 

apinf-add-api-suggetion-1

## Add servers (host)
In the second phase, owner adds host and servers. The text "Servers" in top bar might be better as "hosts" or something. In that part, owner adds:
* Host name
* Host port
* selects protocol
apinf-add-api-suggetion-2

## Add documentation
In the third phase, owner defines documentation. It can be either link to external site or Swagger file. Here should be also information how and where the documentation will be shown eg what is means if external is selected or if swagger is used. 
apinf-add-api-suggetion-3

## Method mapping
In this phase API owner defines the mapping between OWN API methods and APIKA methods. The owner does not define the umbrella.xxx.xx url, it is as text/given. Note that the slug given in phase 1 is used in umbrella url. Also the OWN API hostname is taken from previously given field (phase 2), in here it is apisuomi.fi.  

What the owner gives is the url path to given method in OWN API and then adds from what url it should be responding from umbrella. Owner can add multiple mappings from below "+" icon. 
apinf-add-api-suggetion-4

## Ready to proceed
The last view which is now lacking should be something that verifies that all needed is now given and we are ready to save and publish API. In other words the last view should include "Save" and "Save and Publish" buttons. 

### Additional notes
This could be also used in editing existing APIs. I have no idea how well this behaves and does it make sense in all cases, but to me this kind of approach would be make it easier to add new APIs.

# Resources
The following Meteor package will enable a the described workflow:
* [Meteor Wizard](https://github.com/forwarder/meteor-wizard/)

brylie:


brylie:

brylie:
@apinf/design adding a couple more changes.

jykae:
@brylie This is excellent, and very easy-to-use. Works, approved.

bajiat:
@brylie Thanks for finalising the wizard!

frenchbread:
Looks great!

Issue #595, opened 2015-11-06
Ensure user name exists for all users


bajiat:
We need enhancements to user profile in connection to API feedback / communication channel, etc. Improve the user profile to collect more detail, such as user name.

# Definition of done

* [ ]  Username  should be obligatory. This is related to for instance showing names in communication channel, selecting API managers.
* [ ] Sign up form should contain mandatory field for username.
* [ ] If username is not given in third-party sign up (currently only Github), re-direct users to profile page.
  * [ ] Modify the [existing router hook to ensure that users have usernames](https://github.com/apinf/api-umbrella-dashboard/blob/develop/lib/_config/router.js#L60-L62).
* [ ] Also, remove the Config.username check, as we do not use the global Config object.
* [ ] Allow only unique usernames. If the username already exists, prompt for another username.

sbe:
"Username (either nick or a real user's name) should be obligatory."
Is this about the field "name" in the Schema UserProfile and the "username" field in the User Collection? 

I would like to use the package: https://github.com/meteor-utilities/avatar/ to implement the avatar features, there is an inbuild featue for default profile pictures. the image path could then be added as a field in the UserProfile Schema.

brylie:
@sbe, may we assign you to this task? I am available for pair programming.

At the very least we need to collect a username, but it is up to you whether to make the `name` field mandatory.

sbe:
I am happy to take this task. When are you available for pair programming?

2015-11-27 12:13 GMT+01:00 Brylie Christopher Oxley <
notifications@github.com>:

brylie:
Also, the Avatar package looks great! I am available for pair programming today, and can be found in IRC.

@baijat, may we add @sbe to the Apinf organization (development team), so that he can be assigned this task?

bajiat:
@sbe invited to Apinf organisation and developers team.

bajiat:
@sbe Hi, we are checking sprint 17 task statuses in order to start sprint 18 planning. Can you tell me the status of this task?

sbe:
in progress

bajiat:
@sbe Let us know what the branch is so that we can check progress. Thanks!

sbe:
This is the branch: https://github.com/apinf/api-umbrella-dashboard/tree/feature/595-userprofile-enhancement
I will continue development on wednesday

brylie:
Sandstorm.io has a simple user profile, which is shown to new users automatically when signing up:

[...]

PR #751, opened 2016-01-05
Feature/595 userprofile enhancement

sebbel:
This PR is still WIP.

There are a few minor bugs as stated in #595.

- If chosing an already existant username, the usernameExists function is overridden and a "Mongo db duplicate key error" is shown to the user.
- Removed the rules in Meteor.user.allow in order to make it work for now. Will re-implement the rules when I know what's going on.

### DIFF
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -87,3 +87,4 @@ zenorocha:clipboard
 harrison:papa-parse
 simple:json-routes
 olragon:uri-js
+utilities:avatar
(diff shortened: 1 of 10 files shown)


sebbel:
I would like to remove this Pullrequest in favor of #752

brylie:
OK.

brylie:
@sebbel delete the feature branch, if it is not relevant to your other pull request.

sebbel:
Okay, will delete it now

PR #770, opened 2016-01-15
Ensure branding and color_theme exist before using

brylie:


brylie:
@apinf/developers please review ASAP. This fixes a regression in the `develop` branch.

PR #794, opened 2016-01-25
Adding apiDocumentationEditor to collection and settings page.

sebbel:


sebbel:
Please review @apinf/developers

client/layouts/master_layout/sidebar/sidebar.js
```diff
@@ -6,10 +6,15 @@ Template.sidebar.helpers({
     var backendsCount  = ApiBackends.find({managerIds: currentUserId}).count();
     // return true if user has backends
     return backendsCount > 0;
+  },
+  apiDoc_notEmpty : function() {
```
brylie: While this works, it can be slightly more semantic. I.e. it can be more descriptive:

*"If api documentation editor is configured"*

```
apiDocumentationEditorIsEnabled: function () {}
```

And then in the template:

```html
{{# if apiDocumentationEditorIsEnabled }}
  ...
{{/ if }}
```

client/layouts/master_layout/sidebar/sidebar.js
```diff
@@ -6,10 +6,15 @@ Template.sidebar.helpers({
...
+    //Return true if apiDocumentation was set and false if else.
+    return Settings.findOne({"apiDocumentationEditor.host": {$exists: true}})
```
brylie: This is a logical error, since the `Settings.findOne` actually returns an object or `undefined`. While [JavaScript objects are truthy](http://james.padolsey.com/javascript/truthy-falsey/), they may not technically `true` or `false`. The code works, but may introduce bugs because we are not explicitly passing `true` or `false`, while it appears the intention of the code is to return a boolean.

client/layouts/master_layout/sidebar/sidebar.js
sebbel: Isn't this about whether the apidochost setting being set or not?

Am Dienstag, 26. Januar 2016 schrieb Brylie Christopher Oxley :

client/layouts/master_layout/sidebar/sidebar.js
```diff
@@ -6,10 +6,15 @@ Template.sidebar.helpers({
...
   }
 });
 
 Template.sidebar.created = function () {
   var instance = this;
   instance.subscribe('myManagedApis');
+  instance.subscribe('settings');
```
brylie: Why are we subscribing to settings here?

client/layouts/master_layout/sidebar/sidebar.js
brylie: Yes, and the ApiDocHost setting happens to enable the documentation editor, hence the verbosity.

brylie: In effect, the `not_empty` part, while accurate, does not describe the purpose of the check. Simply renaming the function can make the purpose more clear.

brylie:
After a bit of thought, I have a suggestion to make our settings check more 'high-level' semantic and accurate to the underlying JavaScript. Add a sub-field to the `apiDocumentationEditor` field, called `enabled` with a type of `Boolean`. When the `enabled` field is `true` the `host` field is marked as required. Then, in your template helper, simply return the value of the `enabled` field.

client/layouts/master_layout/sidebar/sidebar.js
sebbel: 'Twas a debug measure.

sebbel:
This would also need a hook which would turn enabled on and off depending on the presence of `apiDocumentationEditor.host`

brylie:
@sebbel, good point. Lets leave the hook for later, if necessary, and simply document the `enabled` field. That way, the host can be passed in and stored, even if the feature is not enabled (e.g. if the editor service is temporarily offline, but the admin wants to keep the host setting preserved.)

sebbel:
Having Problems to grab the value of apiDocumentationEditor.enabled

sebbel:
I have changed it according to your suggestions @brylie, but I am not really pleased with the outcome.
- The checkbox is not reactive, which means that the Documentation Editor Link only appears after the User has saved the settings.
- The Link appears even when there is no host set, I don't know if thats desired.

brylie:
You can tell autoform to auto-save when the form is changed. However, the save step is a good check, to make sure the settings are correct before writing them to the collection.

brylie:
Also, the `host` setting should be conditionally required. In effect, you can tell AutoForm that "when the enabled field is true, the host field is required." Here is an [custom validation](https://github.com/aldeed/meteor-simple-schema#custom-validation) function you can add to the `apiDocumentationEditor` field schema:

```js
"apiDocumentationEditor": {
  type: Object,
  optional: true
},
"apiDocumentationEditor.enabled": {
  type: Boolean,
  optional: true
}
"apiDocumentationEditor.host": {
  type: String,
  regEx: SimpleSchema.RegEx.Url,
  optional: true, // Optional must be true for custom validation
  custom: function () {
      // Custom validator logic, 
      // check the value of enabled field
      // if enabled is true, host field is required
     if (this.field("apiDocumentationEditor.enabled").value === true) {
      return true; // host field is required
    }
  }
}
```

client/layouts/master_layout/sidebar/sidebar.js
```diff
@@ -6,6 +6,13 @@ Template.sidebar.helpers({
     var backendsCount  = ApiBackends.find({managerIds: currentUserId}).count();
     // return true if user has backends
     return backendsCount > 0;
+  },
+  apiDocumentationEditorIsEnabled : function() {
+    // Saving the fields apiDocumentationEditor.enabled into the variable
+    // settings.
+    var settings = Settings.findOne({},{"apiDocumentationEditor.enabled": 1, _id:0})
```
brylie: Where is the subscription that provides data for this query? We may need a new publication that returns only a single field from the Settings document, just as you are doing here. E.g.

```js
Meteor.publish("singleSetting", function (settingKey) {
  // Return a cursor containing only the requested setting from the Settings document
  let settingCursor = Settings.find({}, {settingKey: 1, limit: 1});

  return settingCursor;
});
```

both/collections/settings.js
```diff
@@ -5,6 +5,21 @@ Schemas.SettingsSchema = new SimpleSchema({
...
+  apiDocumentationEditor: {
+    type: Object,
+    optional: true
+  },
+  "apiDocumentationEditor.enabled": {
+    type: Boolean,
+    label: "Enable apiDocumentation Editor",
+    optional:true
+  },
+  "apiDocumentationEditor.host": {
```
brylie: Here is a [custom validation](https://github.com/aldeed/meteor-simple-schema#custom-validation) function you can add to the `apiDocumentationEditor` field schema:

```js
"apiDocumentationEditor": {
  type: Object,
  optional: true
},
"apiDocumentationEditor.enabled": {
  type: Boolean,
  optional: true
}
"apiDocumentationEditor.host": {
  type: String,
  regEx: SimpleSchema.RegEx.Url,
  optional: true, // Optional must be true for custom validation
  custom: function () { // Custom validator logic
    // get the value of apiDocumentationEditor.enabled field
    let enabledFieldValue = this.field("apiDocumentationEditor.enabled").value;

    // if enabled is true, host field is required
    if (enabledFieldValue === true) {
      return true; // host field is required
    }
  }
}
```

sebbel:
Thanks for your feedback @brylie, Really like the generic publish method!

both/collections/settings.js
```diff
@@ -5,6 +5,29 @@ Schemas.SettingsSchema = new SimpleSchema({
...
+    type: String,
+    regEx: SimpleSchema.RegEx.Url,
+    label: "Host",
+    optional: true, // Optional must be true for custom validation
+    custom: function () { // Custom validator logic
+    // get the value of apiDocumentationEditor.enabled field
```
brylie: Check indentation here. Nested blocks should be one level (two spaces) indented.

brylie:
This PR is nearly ready. There is just a minor issue with indentation (two spaces per indent level). Please be consistent with whitespace.

sebbel:
Hope my edits helped maintain a clean whitespace.

brylie:
Looks great! Thanks Sebastian :-)

PR #804, opened 2016-01-28
Fix latest apis row issue on frontpage.

jykae:
I just could not look the overflow of the boxes..

### DIFF
--- a/client/views/latest_api/latestApi.js
+++ b/client/views/latest_api/latestApi.js
@@ -23,7 +23,7 @@ Template.latestApiBackends.created = function () {
 };
 
 Template.latestApiBackends.helpers({
-  'latestBackends': function (limit) {
+  'latestBackends': function (rowSize) {
 
     // Reference to Template instance
     var instance = Template.instance();
@@ -38,6 +38,15 @@ Template.latestApiBackends.helpers({
       apiBackend.relative_created_at = moment(apiBackend.created_at).fromNow();
     });
 
-    return latestApiBackendsList;
+    // Chunk latestApiBackendsList per rowSize
+    chunks = [];
+    rowSize = 4;
+    while (latestApiBackendsList.length > rowSize) {
+        chunks.push({ row: latestApiBackendsList.slice(0, rowSize)});
+        latestApiBackendsList = latestApiBackendsList.slice(rowSize);
+    }
+    chunks.push({row: latestApiBackendsList});
+
+    return chunks;
   }
 });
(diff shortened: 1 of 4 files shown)


brylie:
@elnzv or @frenchbread will you please review?

elnzv:
Still have this issue (Chrome & Firefox)

screencapture-localhost-3000-1454062490709

brylie:
Try fixing this strictly with CSS, rather than writing JavaScript. What is the underlying issue? Overflow can be fixed with CSS - and might simply be fixed by allowing the text to wrap.

client/views/latest_api/latestApi.js
```diff
@@ -38,6 +38,15 @@ Template.latestApiBackends.helpers({
       apiBackend.relative_created_at = moment(apiBackend.created_at).fromNow();
     });
 
-    return latestApiBackendsList;
+    // Chunk latestApiBackendsList per rowSize
```
brylie: How can we rely on Bootstrap to wrap the rows? This seems to be a default pattern of CSS/Bootstrap - wrap divs when they exceed the parent container width.

client/views/latest_api/latestApi.js
```diff
@@ -38,6 +38,15 @@ Template.latestApiBackends.helpers({
...
+    chunks = [];
+    rowSize = 4;
```
brylie: This line hard-codes the row size to 4, which is something we have neither designed nor discussed. The template code renders each API Backend panel width at a variable number of rows, depending on screen resolution:

``html
<div class="col-md-3 col-sm-6 hero-feature">
``

Hard-coding this row size is unnecessary, and possibly interferes with our CSS framework.

bajiat:
@jykae Could you also open an issue for this to describe the observed problem you want to solve with this PR?

jykae:
This adds dynamically new row <div class="row">
I noticed behaviour in APIKA when removing and adding API backends. All elements are put in the same row, and the browser don't know how to render when there's no space. Behaviour is unexpected, could go well or be a mess like it was in APIKA frontpage.
Might be related to APIKA styles only because this was not happening in apinf.io.
This fix is in APIKA already, because I wanted it look good in Helsinki.

brylie:
@jykae please open an issue with the above description and screenshot.

brylie:
I think it may be possible to fix this issue by using CSS/HTML. E.g. http://www.bootply.com/jWoHaY7efR

I am not able to find many examples of other people experiencing this sort of layout issue with Bootstrap panels, which leads me to believe there may be some conflict we have introduced in our own code. We should double check whether any of our custom styles are overriding Bootstrap default behavior.

jykae:
Traced the issue with @brylie 
Added issue https://github.com/apinf/api-umbrella-dashboard/issues/824

Trying to find better fix later on with CSS.

PR #828, opened 2016-02-01
Ensure objects exist before using. Closes #826

### LINKED ISSUE #826: Exception in 'apiDocumentationEditorIsEnabled' in sidebar.js
In the master layout, sidebar.js contains a logical error that causes an exception when settings are not present. Specifically, the funciton tries to return `settings.apiDocumentationEditor.enabled` when `settings` may not be available. Change the function to resemble the following:

```js
apiDocumentationEditorIsEnabled : function() {
    // Saving the fields apiDocumentationEditor.enabled into the variable
    // settings.
    var settings = Settings.findOne({},{"apiDocumentationEditor.enabled": 1, _id:0})
    // Accessing the value of enabled in the apiDocumentationEditor Object
    if(settings && settings.apiDocumentationEditor) {
      return settings.apiDocumentationEditor.enabled
    }
  }
```

brylie:


jykae:
Verified, fixes the issue.

PR #832, opened 2016-02-01
Hotfix/usage instructions - closes #815

### LINKED ISSUE #815: Enhance API usage instructions
Improve the API usage instructions on View API backend page.
* There should be two alternatives for instructions:
  * User is logged in: current instructions, but remove the last sentence "Note: you may need to use your API Key, found on your profile page."
  * User is not logged in: Have a call example without API key. Instruct the users to log in.
* Add error checking for user not logged in to show the latter version of instructions.
* Also, add i18n tags for both versions.

frenchbread:


client/views/api_backends/view_api_backend/usageInstructions/usageInstructions.html
```diff
@@ -1,11 +1,18 @@
...
+        <h4>{{_ "apiBackend_Usage_Instructions_Headline"}}</h4>
+        <em>{{ apiUmbrellaBaseUrl }}</em>
+      </p>
+      {{#unless currentUser}}
+        <p>
+          <strong>
+              {{_ "apiBackend_Usage_Instructions_Note"}}
+          </strong>
+            {{_ "apiBackend_Usage_Instructions_Note_Text"}}
+            <a href="/profile">{{_ "apiBackend_Usage_Instructions_Profile_Page_Link"}}</a>.
```
brylie: This profile link is within an unless block that checks for current user. It reads like "unless there is a current user, show the profile link."  If I am interpreting the code correctly, this means that anonymous users will be shown the profile link.

lib/i18n/en.i18n.json
```diff
@@ -98,6 +98,10 @@
   "view_ApiBackend_ExportYaml": "Export .yaml",
   "view_ApiBackend_Metadata_Title": "Metadata",
   "view_ApiBackend_Metadata_NotFound": "No metadata found.",
+  "apiBackend_Usage_Instructions_Headline": "Make API calls to:",
+  "apiBackend_Usage_Instructions_Note": "Note:",
+  "apiBackend_Usage_Instructions_Note_Text": "You can find your API Key on your",
```
brylie: We want to be careful about partial translation strings, as they are difficult to translate accurately. I.e. different languages have different word order, and sometimes use one word in place of multiple English language words.

PR #837, opened 2016-02-02
Feature/autoform i18n. Closes #764

### LINKED ISSUE #764: Enable translation of AutoForm/SimpleSchema labels
There is a package that extends tap:i18n to allow SimpleSchema/AutoForm labels to be internationalized. Locate this package, and determine whether we should install it. If installed, add i18n strings to project schemas.

Definition of done
---
* [ ] Package enabling SimpleSchema/AutoForm labels is included in project.
* [ ] SimpleSchema/AutoForm label strings have been added to the schema (to be clarified).

# Resources
* [AutoForm-i18n](https://github.com/gwendall/meteor-autoform-i18n) package

brylie:
Add AutoForm i18n package. Add i18n capabilities to Add API Wizard as example that can be followed in subsequent tasks.


brylie:

lib/i18n/en.i18n.json
```diff
@@ -210,5 +210,38 @@
   "editApi_hints_require_https": "Choose 'required' if HTTPS is mandatory for accessing your apis.",
   "editApi_hints_api_key_verification_level": "Choose 'required' if API key is required to access your API",
   "editApi_hints_api_balance_algorithm": "In case you have provided multiple Front-end : Back-end host pairs, choose the balancing algorithm using which umbrella should forward the requests to your api backend.",
-  "editApi_hints_settings": "Define header values that will be set in the response regardless of whether the header is already set in the response.<br><br> For example, to force CORS headers on all responses:<br><br> 'Access-Control-Allow-Origin: *'"
+  "editApi_hints_settings": "Define header values that will be set in the response regardless of whether the header is already set in the response.<br><br> For example, to force CORS headers on all responses:<br><br> 'Access-Control-Allow-Origin: *'",
+  "schemas": {
```
elnzv: Maybe, we need to add those new strings to Finnish json file as well.
That's gonna be easier for interpreter to work.

client/views/api_backends/add-backend-workflow/wizard.js
```diff
@@ -43,19 +41,20 @@ Template.addApiBackendWizard.helpers({
         optional: true
       },
       "url_matches.frontend_prefix": {
-        label: 'Frontend prefix',
         optional: true,
         type: String,
         regEx: SimpleSchema.RegEx.Prefix
       },
       "url_matches.backend_prefix": {
-        label: 'Backend prefix',
```
elnzv: What about using TAPi18n function here?
This way: 
```
label: TAPi18n.__('wizard_url_matches_frontend_prefix')
```

This way we will keep our languages' files consistent.

lib/i18n/en.i18n.json
elnzv: If we use TAPi18n function here, all those strings will be defined the same way as others.

client/views/api_backends/add-backend-workflow/wizard.js
brylie: I removed that line, as labels are handled  by the i18n package.

brylie:
It turns out that the TAPi18n.__ function only retrieves the English translation for schemas that are attached to collections. This means that any AutoForm instance that generates fields based on a collection will not get the correct language strings when the user language is something other than English. I have a local demonstration available, but you can verify this by changing one of the schemas in `/both/collections` to use the `TAPi18n.__()` function for `label`, `placeholder`, and/or `options`, such as the following:

```js
// in apiBacklog.js
ApiBacklogItems.attachSchema(new SimpleSchema({
  title: {
    type: String,
    label: TAPi18n.__('Dashboard'),
    max: 100,
    autoform: {
      placeholder: TAPi18n.__('Dashboard')
    }
  },
  details: {
    type: String,
    label: TAPi18n.__('Dashboard'),
    max: 1000,
    autoform: {
      rows: 5,
      placeholder: TAPi18n.__('Dashboard')
    }
  },
...
```

elnzv:
OK, good to know.
Merging.

PR #844, opened 2016-02-04
Remove defaultLanguage configuration. Closes #814

### LINKED ISSUE #814: Allow site 'default language' to be configured from Settings page, or remove from config
Currently, the site default language is stored in `/lib/_config/_config.js`. Enable the default language to be configured from the Settings page.

* [ ] Move this setting to the [Settings collection](https://github.com/apinf/api-umbrella-dashboard/blob/develop/both/collections/settings.js#L3). 
* [ ] Ensure the [default language is set from the Settings collection](https://github.com/apinf/api-umbrella-dashboard/blob/develop/lib/_config/i18n.js#L3-L7). 
* [ ] Remove the [`defaultLanguage` key from `/lib/_config/_config.js`](https://github.com/apinf/api-umbrella-dashboard/blob/develop/lib/_config/_config.js#L22)

brylie:
Remove the default language setting from Config and startup. TAPi18n auto-detects user locale and sets language, which user can then manually change. I am not sure whether we need/or want a default language override.

### DIFF
--- a/lib/_config/_config.js
+++ b/lib/_config/_config.js
@@ -4,7 +4,6 @@ Config = {
   subtitle: function() {
     return TAPi18n.__('configSubtitle');
   },
-  defaultLanguage: 'en',
   dateFormat: 'D/M/YYYY',
   legal: {
     address: 'Jessnerstrasse 18, 12047 Berlin',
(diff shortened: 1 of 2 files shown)


jykae:
The default language fallbacks in English if it is not defined. Anyway, good cleanup and let's merge this.

In future as an enhancement we could bind the language setting to user profile. I'll create enhancement issue about that.

PR #858, opened 2016-02-08
Feature/737 i18n contains concatenated strings

### LINKED ISSUE #737: i18n file contains concatenated strings
The i18n file contains strings that are parts of sentence. 

The first rule of software internationalisation is never to use concatenated strings (Ui texts that are built from several parts). They can create localisation problems, because either the person translating does not know, which strings go together or translator cannot use natural word order for the entire sentence in target language, because a sentence has been split. 

One example from the Apinf strings: 
* "Meanwhile, you may" (key: Not Found.notFound_Message_Part2)
* "return to a previous page" (key: Not Found.notFound_Message_Part3)

All concatenated strings should be combined into entire meaningful text strings. Use parameters for replaceable items, if necessary.

Definition of done
---
* [x] Find all cases of concatenated strings.
* [x] Modify corresponding template to use only one i18n string in place of multiple parts of string.
* [x] Modify the i18n file to merge multiple parts into one string corresponding template.
* [x] If the strings have been concatenated because of HTML tags, make sure they get included and are not displayed on UI.

elnzv:
Closes #737 

@apinf/developers, review. 
@bajiat, also review if there are any concatenated strings left.


brylie:
@elnzv this looks good now. Are you still wanting to add commas via CSS, or is this ready to merge?

PR #861, opened 2016-02-08
Feature/849 manager role check sidebar, closes #849

### LINKED ISSUE #849: Use 'manager' role to toggle visibility of 'My Managed APIs' link in sidebar; cleanup subscription
Currently, we are using a [`userHasApis`](https://github.com/apinf/api-umbrella-dashboard/blob/develop/client/layouts/master_layout/sidebar/sidebar.js#L2) template helper to toggle visibility of the 'My Managed APIs' link in the master layout sidebar. The helper [performs a database query](https://github.com/apinf/api-umbrella-dashboard/blob/develop/client/layouts/master_layout/sidebar/sidebar.js#L6) and [returns true if there are any query results](https://github.com/apinf/api-umbrella-dashboard/blob/develop/client/layouts/master_layout/sidebar/sidebar.js#L8). This has required a [subscription](https://github.com/apinf/api-umbrella-dashboard/blob/develop/client/layouts/master_layout/sidebar/sidebar.js#L24), which is 'polluting' other features and leading to some false assumptions about data availability.

# Task
* Use the `{{#if isInRole 'manager'}}` helper to toggle visibility of `My Managed APIs` navigation menu link
* Remove the [`userHasApis`](https://github.com/apinf/api-umbrella-dashboard/blob/develop/client/layouts/master_layout/sidebar/sidebar.js#L2) helper
* Remove the [`myManagedApis` subscription](https://github.com/apinf/api-umbrella-dashboard/blob/develop/client/layouts/master_layout/sidebar/sidebar.js#L24)

jykae:
WIP, onSuccess hook not yet working.


jykae:
Did not get onSuccess hook working with wizard, so added it after successful publish.

server/migrations.js
```diff
@@ -0,0 +1,16 @@
+Migrations.add({
+  version: 1,
+  name: 'Add manager role to all users that have APIs.',
+  up: function() {
+    let allManagerIds = [];
```
brylie: Use `const`here.

server/migrations.js
```diff
@@ -0,0 +1,16 @@
...
+    let allApis = ApiBackends.find().fetch();
```
brylie: Use `const`here, since this reference will not be re-assigned.

server/migrations.js
```diff
@@ -0,0 +1,16 @@
...
+    // Iterate through allApis and push managerIds to allManagerIds array
+    for(api in allApis) {
+      allManagerIds.push.apply(allManagerIds, allApis[api].managerIds);
```
brylie: There may be a logical error here, since the `managerIds` field is an array. In effect, this code is creating an array of arrays, which might not behave as expected in the next `for` loop.

brylie: Also, consider re-writing this code using `Array.prototype.map` followed by underscore's `flatten` method, as such:

```js
// Create a nested array, using map
let managerIds = allApis.map(function (api) {
  return api.managerIds;
});

// Flatten the nested array
managerIds = _.flatten(managerIds);
```

server/migrations.js
```diff
@@ -0,0 +1,16 @@
...
+    }
+    for(manager in allManagerIds) {
```
brylie: Consider re-writing this using `Array.prototype.forEach`, for clarity:

```js
managerIds.forEach(function (managerId) {
  // Add user to manager Role
  Roles.addUsersToRoles(managerId, ['manager']);
});
```

brylie: Also note, the role can be a string, rather than an array, for clarity.

brylie:
Another thing I notice is the following error in server console:

'''
Error: Missing 'users' param
 (STDERR)     at Object._.extend._updateUserRoles (packages/alanning_roles/roles_common.js:608:1)
     at Object._.extend.addUsersToRoles (packages/alanning_roles/roles_common.js:145:1)
     at Object.Migrations.add.up (server/migrations.js:13:13)
     at migrate (packages/percolate_migrations/migrations_server.js:209:1)
     at Object.Migrations._migrateTo (packages/percolate_migrations/migrations_server.js:229:1)
     at Object.Migrations.migrateTo (packages/percolate_migrations/migrations_server.js:143:1)
     at server/startup.js:3:14
     at /home/brylie/Code/Apinf/api-umbrella-dashboard/.meteor/local/build/programs/server/boot.js:249:5
'''

This *may* be related to data inconsistency. I.e. some of the API Backends are retrieved directly from API Umbrella and have no `managerIds` field attached, since they were not created through the *current* Apinf instance.

server/migrations.js
jykae: @brylie We can't use const if we are mutating the assigned object/array. This gives error when flattening.

jykae:
@brylie I am now skipping those API Backends that don't have managerIds field assigned, as there's another issue for that.

brylie:
OK, cool. Please resolve conflicts.

brylie:
@jykae this looks good. What is the preventing this from being ready for merging?

jykae:
Probably just conficts, looks like this works. Merge if it also works for you.

PR #872, opened 2016-02-12
Feature/816 date format - closes #816

### LINKED ISSUE #816: Allow 'date format' to be defined from user locale
Currently, the site default date format is stored in `/lib/_config/_config.js`. Enable the date format to be configured from the Settings page.

* [x] Ensure the [date format is set from the user locale within `/lib/utils.js`](https://github.com/apinf/api-umbrella-dashboard/blob/develop/lib/utils.js#L5). 
* [x] Remove the [`defaultLanguage` key from `/lib/_config/_config.js`](https://github.com/apinf/api-umbrella-dashboard/blob/develop/lib/_config/_config.js#L23)
* [x] If user locale is none of the locales defined in project translations, English (en) is used as default locale.

# Reference
* Moment.js [i18n documentation](http://momentjs.com/docs/#/i18n/)

frenchbread:

### DIFF
--- a/both/collections/settings.js
+++ b/both/collections/settings.js
@@ -155,6 +155,14 @@ Schemas.SettingsSchema = new SimpleSchema({
     autoform: {
       placeholder: 'D/M/YYYY'
     }
+  },
+  "userLocale": {
+    type: String,
+    label: "User Locale",
+    optional: true,
+    autoform: {
+      placeholder: 'en'
+    }
   }
 });
 
--- a/both/collections/settings.js
+++ b/both/collections/settings.js
@@ -147,6 +147,14 @@ Schemas.SettingsSchema = new SimpleSchema({
     autoform: {
       placeholder: 'xxx'
     }
+  },
+  "dateFormat": {
+    type: String,
+    label: "Date Format",
+    optional: true,
+    autoform: {
+      placeholder: 'D/M/YYYY'
+    }
   }
 });
 
--- a/client/views/settings/settings.html
+++ b/client/views/settings/settings.html
@@ -69,6 +69,14 @@
             {{> afQuickField name='contactForm.toEmail'}}
           </div>
         </div>
+        <div class="panel panel-default">
+            <div class="panel-heading">
+                <h3 class="panel-title">Other</h3>
+            </div>
+            <div class="panel-body">
+                {{> afQuickField name='dateFormat'}}
+            </div>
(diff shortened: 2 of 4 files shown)


jykae:
I'll review this one

both/collections/settings.js
```diff
@@ -147,6 +147,22 @@ Schemas.SettingsSchema = new SimpleSchema({
     autoform: {
       placeholder: 'xxx'
     }
+  },
+  "dateFormat": {
```
jykae: This could possibly have options of common date format types. And use options list in the Settings page, or validate input with regex.

both/collections/settings.js
```diff
@@ -147,6 +147,22 @@ Schemas.SettingsSchema = new SimpleSchema({
...
+    type: String,
+    label: "Date Format",
+    optional: true,
+    autoform: {
+      placeholder: 'D/M/YYYY'
+    }
+  },
+  "userLocale": {
```
jykae: User locale could use a list of options as well, I suppose.

jykae:
General notice: We are not using prettyDate() function anywhere, so it's hard to verify whether it works or not, when I set date format and user locale.

frenchbread:
@jykae Good suggestion. Wondering if it's okay to "hardcore" user locales (and date formats) because it seems to be a huge list or preload them from somewhere?

`prettyDate()` function is not used anywhere, for now, I guess. The way to test is to fire up `$ meteor shell` and call `Config.dateFormat()` or `Config.userLocale()` since they are global objects.

brylie:
We are deprecating the `Config` object. User locale and date format can be retrieved dynamically. We just need to determine how we want to determine user locale, e.g. via moment.js, tap i18n, etc.

jykae:
@brylie @frenchbread If we are deprecating still the Config object, I would say that let's close this PR or continue on this and mark WIP. Moment usage I think is ok, we used it also in DIpor. 
User locale preference is usually saved in user profile.
Are you continuing this @frenchbread or should I take this? How important this would be for next sprint @brylie ?

bajiat:
@frenchbread @jykae What is the status? Did the config file get deprecated?

frenchbread:
@bajiat I am working on it in [this](https://github.com/apinf/api-umbrella-dashboard/tree/feature/cleanup-config-file) branch. Removed config files.

PR #900, opened 2016-03-01
Add basic docker image generator

Sturgelose:
It creates a basic image with base on meteor-tupperware  to start a MeterorJS container app fast. It still has some problems as it seems it gets stuck when starting the app. It might need some debugging or some ENV variables. So this pull request needs to be improved.

Use `docker build . -t <name_of_the_image>` in order to build the image and then `docker run <name_of_the_image>` to have it running.

You will see in the logs that it fails on start because of some variables missing or not sure...


brylie:
@Sturgelose thank you for working on this. We will try to carry your work forward and resolve the errors. Cheers!

PR #907, opened 2016-03-04
Fixes latest apis rows' view

### LINKED ISSUE #824: Latest APIs row renders wrong with title that has more than one line
Find out way to fix this with CSS if possible.

latestapis

elnzv:
Closes #824

### DIFF
--- a/client/views/latest_api/latestApi.less
+++ b/client/views/latest_api/latestApi.less
@@ -15,6 +15,19 @@
 
   .api-box {
     padding-left: 1em;
+    h3 {
+      overflow: hidden;
+      text-overflow: ellipsis;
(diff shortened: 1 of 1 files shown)


client/views/latest_api/latestApi.less
```diff
@@ -15,6 +15,19 @@
 
   .api-box {
     padding-left: 1em;
+    h3 {
+      overflow: hidden;
+      text-overflow: ellipsis;
```
jykae: This is cool :)

jykae:
Works great,
frontlong

Something could be done also on API view page header, but it was not in definition for this task
longname

PR #925, opened 2016-04-27
Feature/setup wizard

### LINKED ISSUE #922: Add initial setup wizard for new deployments
When setting up a new Apinf instance, present the first administrator with a set-up wizard. This wizard should help the Admin complete any required, or optional, settings. It can optionally include a branding step.

# Suggested solution
One [suggested solution](https://forums.meteor.com/t/first-time-setup-wizard-recipe/22057/2?u=brylie) would be the following:
* [x] Create a *[server method](https://github.com/meteor/guide/blob/master/content/methods.md)* to check for required settings. 
  - E.g. `checkRequiredSettings`
* [x] Add a [global router hook](http://iron-meteor.github.io/iron-router/#hooks), via `Router.onBeforeAction()` or `Router.onRun()`, that calls the *server* method on every route
  - E.g. `Router.onBeforeAction(function () {  Meteor.call('checkRequiredSettings', ... ) });`
* [x] If the server method returns `false` the router should re-direct the user to the configuration screen
* [ ] Project starts successfully by calling only `meteor`. I.e. the `--settings` flag is no longer required.

  - Note: the configuration wizard should make sure an administrator is logged in, to prevent abuse

**Warning**: Avoid publishing the settings to the client, as not all settings should be publicly known. E.g. we have private keys stored in our settings, so the `checkRequiredSettings` function should only run on the server.

# Resources
- See the [discussion thread in Meteor forums](https://forums.meteor.com/t/first-time-startup-wizard-recipe/22057) related to this task.
- See the Meteor Guide [Methods](https://github.com/meteor/guide/blob/master/content/methods.md) article for details about Meteor Methods
- See the Iron Router Guide [Hooks](http://iron-meteor.github.io/iron-router/#hooks) article for details about router hooks

Alapan:
Closes #922 

Important changes:
 
- Created a server method to check the value of a flag called `initialSetupComplete`. If this returns `false`, an initial setup wizard is to be shown.
- The wizard allows an admin to configure settings for the instance.


lib/_config/router.js
```diff
@@ -68,6 +68,13 @@ onAfterAction = function() {
   }, 300);
 };
 
+Router.onBeforeAction(function() {
+  var configRequired = Meteor.call('checkRequiredSettings');
```
brylie: Since this code is relying on a server method, it is asynchronous. So, we need to handle the return value from `checkRequiredSettings` in a callback:

```js
// Instead of
// var configRequired = Meteor.call("checkRequiredSettings");
// It should read

// Get a named reference to current router
// for use inside the server method callback
const currentRouter = this;

Meteor.call("checkRequiredSettings", function (error, result) {
  // if error....

  // do something with the result:
  if (result === true) {
    Router.go("settingsWizard);
  } else {
    // call this.next() using the currentRouter variable from above
    currentRouter.next()
  }
});
```

client/layouts/master_layout/master_layout.js
```diff
@@ -17,5 +17,9 @@ Template.masterLayout.helpers({
     if (lastUploadedLogo) {
       return lastUploadedLogo
     }
+  },
+  initialSetupComplete: function() {
+    var initialSetupComplete = Meteor.call('additionalSetupRequired');
```
brylie: This probably will not work, since method calls are asynchronous. Also, consider a different approach here, rather than modifying the master layout.

In effect, redirecting the user to a setup wizard supersedes this alert, and the alert can be displayed in the setup wizard template instead of the master layout.

client/views/settings_wizard/settingsAlert.html
```diff
@@ -0,0 +1,9 @@
+<template name="settingsAlert">
+  <div class="alert alert-warning">
+    <a href="#" class="close" data-dismiss="alert" aria-label="Close">&times;</a>
+    <p>
+      <strong>Additional configuration required!</strong><br>
```
brylie: Avoid shouting at the user.

client/views/settings_wizard/settingsAlert.html
```diff
@@ -0,0 +1,9 @@
...
+        Hey there, this instance of <strong>APINF</strong> needs additional configuration. Click <a href="{{pathFor route='settingsWizard'}}">here</a> to continue. You must be an admin to do this!!
```
brylie: Again, avoid shouting at the user.

client/views/settings_wizard/wizard/settingsWizard.js
```diff
@@ -0,0 +1,106 @@
...
+    $('#settingsCarousel').carousel('next');
+  },
+  'click #prev-second-slide': function() {
+    // clicking Previous of second slide moves to previous slide
+    $('#settingsCarousel').carousel('prev');
+  },
+  'click #save-settings': function() {
+    // when configuration is done, call server method to set initialSetupComplete to true, so that the settings alert is no longer shown
+    Meteor.call("initialSetupCompleteTrue");
+    Router.go("settingsComplete");
```
brylie: Call this from within the `initialSetupCompleteTrue` callback, so the method has time to run first.

lib/_config/router.js
```diff
@@ -38,6 +38,16 @@ var requireAdminRole = function () {
   }
 };
 
+var additionalSetupRequired = function () {
```
brylie: Use `const` here.

brylie:
Good work so far. Just a couple of minor details to catch at this point.

client/views/settings_wizard/wizard/settingsWizard.js
```diff
@@ -0,0 +1,106 @@
...
+      $('[data-schema-key],button').attr("disabled", "disabled");
+    },
+    endSubmit: function () {
+      // Enable form elements after form submission
+      $('[data-schema-key],button').removeAttr("disabled");
+    }
+  }
+});
+
+AutoForm.addHooks(['settings'], {
```
brylie: This is your AutoForm callback context. You should call the server method from within the `onSuccess` callback.

client/views/settings_wizard/wizard/settingsWizard.js
```diff
@@ -0,0 +1,106 @@
...
+    // clicking Save and next of first slide moves to second slide
+    $('#settingsCarousel').carousel('next');
+  },
+  'click #prev-second-slide': function() {
+    // clicking Previous of second slide moves to previous slide
+    $('#settingsCarousel').carousel('prev');
+  },
+  'click #save-settings': function() {
+    // when configuration is done, call server method to set initialSetupComplete to true, so that the settings alert is no longer shown
+    Meteor.call("initialSetupCompleteTrue");
```
brylie: Move this method call to the `AutoForm.onSuccess` callback.

Alapan:
@brylie I made the changes you suggested. However, I made a slight change in the logic - I have introduced a new field `initialSetupComplete` in the `Settings` collection. Otherwise, a setting created through the program is not persisting across multiple templates. So if `initialSetupComplete` is set to `true` when the relevant method is called through Template 1, the value is becoming `undefined` when it is checked through Template 2. 

I have also removed the alert logic from the master layout file. Currently the alert is not being shown anywhere.

brylie:
Ok, make sure that the problem is not related to meteor restarting when editing the project sourcecode.

If you take the collection approach, be sure to create a publication that only publishes the `initialSetupComplete` field from the settings collection document. Otherwise, we may inadvertently publish sensitive information from our settings collection.

Alapan:
I made the changes.

brylie:
Great! :-) Is this ready for review/merge?

Alapan:
@brylie Yes, this is ready for review.

shaliko:
@Alapan project stop works for me if I do ```meteor reset``` and try open any page. Looks like infinite loop.

Alapan:
@shaliko I made a small change. Could you please test and let me know if you are still getting the same error?

brylie:
When starting the project without any settings, e.g. by running `meteor` without the `--settings` flag, I get the following error in the shell:

```
Error: After filtering out keys not in the schema, your object is now empty
```

brylie:
When starting Apinf *without* any settings, i.e. without the `--settings ...` flag, I get the following error:

```
Error: After filtering out keys not in the schema, your object is now empty
```

One goal of this task is that Apinf should start with no initial settings, so the Wizard can display.

brylie:
It seems like `/server/startup.js` may need significant re-work/review here. Specifically,

- Lines 11 and 24 attempt to insert the `Meteor.settings` object into the settings collection without verifying the structure of that object
  - initially, `Meteor.settings` has the following structure `{ public: {} }`, which doesn't contain any relevant settings
- There are two large sections of that file with duplicate functionality (lines 4-12 and lines 20-24)
  - refactor this code, so that it only runs once
- We should not call `Meteor.call("createApiUmbrellaWeb")` without first verifying that API Umbrella settings have been stored in the `Settings` *collection*
  - I.e. all of code may be refactored to first use the `Settings` collection as opposed to the `Meteor.settings` object
  - The `Meteor.settings` object will be a temporary placeholder for settings, so they can be persisted in the Settings collection
- Any API Umbrella related methods should only run if there are valid API Umbrella settings in the `Settings` collection
  - The *'Sync API Umbrella Users and API Backends'* `SyncedCron` job should only start if API Umbrella settings are present
    - the job, including any sub-methods,  should check the `Settings` collection, rather than the `Meteor.settings` object
  - `Meteor.call("syncApiUmbrellaUsers");` and `Meteor.call("syncApiBackends");` should also only run if API Umbrella settings are stored in the Settings collection.

brylie:
We cannot figure out how to get the settings wizard to display.

Alapan:
Currently, after running `meteor reset` and then `meteor` (without the settings file), when a user registers successfully, he is redirected to the wizard. However, currently I am facing a problem setting the Host name to `http://localhost:3000` in the wizard, as the application returns an `Unknown validation error`. Setting the host name to something like `http://example.com` works, though. 

Also, even after configuring the settings successfully, the following errors are being seen in the console: 
1. `[TypeError: Cannot read property 'clientId' of undefined]`
2. `[TypeError: Cannot read property 'username' of undefined]`

brylie:
Ah, yes. We had the same validation issue, possibly because of the port number.

As a second thought, it is probably unnecessary to provide the HOST value for the Apinf platform, since Meteor is already aware of this value, so lets deprecate that from the setup wizard/settings.

brylie:
@frenchbread, what is your recollection of why we add the `host` settings in `config.json`? I think it had something to do with the dashboard.

brylie:
@Alapan regarding the 'clientId' and 'username' errors, we need to check the existance of these fields before trying to use them. Try commenting out lines 11 and 24 of `server/startup.js`, as mentioned above.

bajiat:
@Alapan and @shaliko Could you do some pair programming on this one to try solve the issues Alapan is facing? Sometimes even just explaining the problem to someone else might already help.  Shaliko is not available 9 May, but should be back at work 10 May.
@brylie said he was just quessing to a certain extent what the problem was.

Alapan:
I tried commenting the lines as @brylie suggested, but it didn't help. Please note, there was no problem in running the application - just that those 2 TypeErrors are present after filling in the settings details.

brylie:
Ah, I understand now. Are they happening after you submit the form? It may still be a validation problem, where the schema is not passed in the fields it needs for validation.

brylie:
@shaliko are you available today to help troubleshoot this task?

shaliko:
@brylie Today joining to this task.

brylie:
Thanks @shaliko. @Alapan do you have time to work with @shaliko today?

Alapan:
I can work on this today late evening. I don't know if @shaliko will be available then. I will try to fix the error and update in a comment the status of my work.

shaliko:
@Alapan I will finish work in ~1 hour. Will work with you next time ;)

Alapan:
Along with the type errors, after entering the settings successfully, going to the dashboard page, waiting for the charts to appear, then logging out and logging back in, another error is seen on the console:
`Exception while invoking method 'getChartData' TypeError: Cannot read property 'host' of undefined`.

This is when setting the host name to something like `http://www.example.com`. Clearly, a proper host name is necessary. But entering `http://localhost:3000` is giving an unknown validation error

brylie:
Ah, yep. That is a known bug. We need to decouple the dashboard from elasticsearch, or make the elasticsearch settings required.

brylie:
For reference: https://github.com/apinf/api-umbrella-dashboard/issues/927

brylie:
Ping @ccsr @bajiat @ashakunt

ashakunt:
@Alapan issue #927 has been resolved by pull request #944 today. Can you check if this issue is still reproducible? 

Quote: https://github.com/apinf/api-umbrella-dashboard/pull/925#issuecomment-218619497

brylie:
The exception in `getChartData` may still be there. Please ignore, and finalize the wizard PR.

Alapan:
I didn't see the exception now, and the charts were showing after registering, logging out, and logging back in. Please review.

brylie:
Great work!

On the first page of the setup wizard, the 'choose file' buttons do not work in Firefox 46. It may be related to the following upstream bug:

https://github.com/yogiben/meteor-autoform-file/issues/104

brylie:
One small improvement would be to redirect the user to the dashboard after the wizard, rather than having them click a link. Our router package has a `Router.go()` method that would work here.

client/views/settings_wizard/settingsAlert.html
```diff
@@ -0,0 +1,9 @@
+<template name="settingsAlert">
+  <div class="alert alert-warning">
+    <a href="#" class="close" data-dismiss="alert" aria-label="Close">&times;</a>
+    <p>
+      <strong>Additional configuration required - </strong><br>
+        Hey there, this instance of <strong>APINF</strong> needs additional configuration. Click <a href="{{pathFor route='settingsWizard'}}">here</a> to continue. You must be an admin to do this.
```
brylie: Lets remove this text for a couple of reasons:

- it is no longer necessary, since the wizard automatically shows up without the user needing to click
- text like this, with mixed in markup and 'click here' links is not ideal to translate

client/views/settings_wizard/wizard/settingsComplete.html
```diff
@@ -0,0 +1,5 @@
+<template name="settingsComplete">
+  <div class="alert alert-success">
+    Settings configured successfully! Click <a href="{{pathFor route='dashboard'}}">here </a> to return to your dashboard.
```
brylie: Rather than asking the user to 'click here', lets simply redirect to the dashboard using `Router.go()`

client/views/settings_wizard/wizard/settingsWizard.html
```diff
@@ -0,0 +1,153 @@
...
+  <div class="container">
+    <h3 class="centered">APINF Configuration Wizard</h3>
+    <div id="settingsCarousel" class="carousel slide" data-ride="carousel" data-interval="false">
+      <div class="carousel-inner" role="listbox">
+        <div class="item active">
+          <div class="row">
+            <div class="col-xs-12">
+              <div class="panel panel-default">
+                <div class="panel-heading">
+                  <h3 class="panel-title">Step 1 - Project Branding</h3>
```
brylie: Add i18n token here

client/views/settings_wizard/wizard/settingsWizard.html
```diff
@@ -0,0 +1,153 @@
...
+                  {{> afQuickField name='siteTitle'}}
+                  {{> afQuickField name='siteAbbreviation'}}
+                  {{> afQuickField name='siteSlogan'}}
+                  {{> afQuickField name='siteFooter'}}
+                  {{> afQuickField name='color_theme' options='allowed'}}
+                  {{> afQuickField name='socialMedia'}}
+                  <div id="form-buttons">
+                    <div class="row">
+                      <div class="col-xs-12 col-sm-6 form-group">
+                        <button type="button" class="btn btn-default btn-block" id="prev-first-slide">Previous</button>
```
brylie: Add i18n token here

client/views/settings_wizard/wizard/settingsWizard.html
```diff
@@ -0,0 +1,153 @@
...
+                      </div>
+                      <div class="col-xs-12 col-sm-6 form-group">
+                        <button type="submit" class="btn btn-success btn-block" id="next">Save and Next</button>
```
brylie: Add i18n token here

client/views/settings_wizard/wizard/settingsWizard.html
```diff
@@ -0,0 +1,153 @@
...
+                  {{> afQuickField name='siteTitle'}}
+                  {{> afQuickField name='siteAbbreviation'}}
+                  {{> afQuickField name='siteSlogan'}}
+                  {{> afQuickField name='siteFooter'}}
+                  {{> afQuickField name='color_theme' options='allowed'}}
+                  {{> afQuickField name='socialMedia'}}
+                  <div id="form-buttons">
+                    <div class="row">
+                      <div class="col-xs-12 col-sm-6 form-group">
+                        <button type="button" class="btn btn-default btn-block" id="prev-first-slide">Previous</button>
```
brylie: Add i18n token here

client/views/settings_wizard/wizard/settingsWizard.html
```diff
@@ -0,0 +1,153 @@
...
+                      </div>
+                      <div class="col-xs-12 col-sm-6 form-group">
+                        <button type="submit" class="btn btn-success btn-block" id="next">Save and Next</button>
```
brylie: Add i18n token here

client/views/settings_wizard/wizard/settingsWizard.html
```diff
@@ -0,0 +1,153 @@
...
+          </div>        
+        </div>
+        <div class="item">
+          <div class="row" id="champion-messages">
+          {{#autoForm collection="Settings" doc=editDoc id="settings" type=formType}}
+            <div class="col-xs-12">
+              <div class="well">
+              <div class="panel panel-default">
+                <div class="panel-heading">
+                  <h3 class="panel-title">Step 2 - Settings for API Umbrella</h3>
```
brylie: Add i18n token here

client/views/settings_wizard/wizard/settingsWizard.html
```diff
@@ -0,0 +1,153 @@
...
+                <div class="panel-body">
+                {{> afQuickField name='apiUmbrella.host'}}
+                {{> afQuickField name='apiUmbrella.apiKey'}}
+                {{> afQuickField name='apiUmbrella.authToken'}}
+                {{> afQuickField name='apiUmbrella.baseUrl'}}
+                </div>
+              </div>
+              <div class="panel panel-default">
+                <div class="panel-heading">
+                  <h3 class="panel-title">Api Documentation Editor</h3>
```
brylie: Add i18n token here

client/views/settings_wizard/wizard/settingsWizard.html
```diff
@@ -0,0 +1,153 @@
...
+                <div class="panel-body">
+                {{> afQuickField name='apiDocumentationEditor.enabled'}}
+                {{#if afFieldValueIs name="apiDocumentationEditor.enabled" value=true}}
+                  {{> afQuickField name='apiDocumentationEditor.host'}}
+                {{/if}}
+                </div>
+              </div>
+              <div class="panel panel-default">
+                <div class="panel-heading">
+                  <h3 class="panel-title">Elasticsearch</h3>
```
brylie: Add i18n token here

client/views/settings_wizard/wizard/settingsWizard.html
```diff
@@ -0,0 +1,153 @@
...
+                </div>
+                <div class="panel-body">
+                {{> afQuickField name='elasticsearch.host'}}
+                </div>
+              </div>
+              <div class="panel panel-default">
+                <div class="panel-heading">
+                  <h3 class="panel-title">Mail</h3>
```
brylie: Add i18n token here

client/views/settings_wizard/wizard/settingsWizard.html
```diff
@@ -0,0 +1,153 @@
...
+                  <h3 class="panel-title">Additional</h3>
+                </div>
+                <div class="panel-body">
+                {{> afQuickField name='contactForm.toEmail'}}
+                </div>
+              </div>
+              <div id="form-buttons">
+                <div class="row">
+                  <div class="col-xs-12 col-sm-6 form-group">
+                    <button type="button" class="btn btn-default btn-block" id="prev-second-slide">Previous</button>
```
brylie: Add i18n token here

client/views/settings_wizard/wizard/settingsWizard.html
```diff
@@ -0,0 +1,153 @@
...
+                  </div>
+                  <div class="col-xs-12 col-sm-6 form-group">
+                    <button type="submit" class="btn btn-success btn-block" id="save-settings">Finish</button>
```
brylie: Add i18n token here

client/views/settings_wizard/wizard/settingsWizard.js
```diff
@@ -0,0 +1,104 @@
...
+      $('[data-schema-key],button').removeAttr("disabled");
+    }
+  }
+});
+
+AutoForm.addHooks(['settings'], {
+  onSuccess: function () {
+    // Call method to update Meteor.settings
+    Meteor.call('updateMeteorSettings');
+    FlashMessages.sendSuccess('Settings saved.');
```
brylie: Add i18n text here here:

```js
// Get success message
const successMessage = TAPi18n.__("setupWizard_successMessage");
```

client/views/settings_wizard/wizard/settingsWizard.js
```diff
@@ -0,0 +1,104 @@
...
+    // Check if we can create ApiUmbrellaWeb object
+    try {
+      Meteor.call("createApiUmbrellaWeb");
+      Meteor.call("initialSetupCompleteTrue", function() {
+        Router.go("settingsComplete");
```
brylie: Instead of routing to the `settingsComplete`, simplify our code/process by routing to the Dashboard and showing a success message.

lib/_config/router.js
```diff
@@ -68,6 +86,10 @@ onAfterAction = function() {
   }, 300);
 };
 
+// not sure how to check additionalSetupRequired for all routes
+// the below implementation causes an infinite redirect loop at sign-in
+Router.onBeforeAction(additionalSetupRequired, {except: []});
```
brylie: If this code works as expected, update the comment in the line above. Right now, the comment explains that things are not working as expected.

lib/router/main.js
```diff
@@ -20,6 +20,18 @@ Router.route("/api/:_id/edit", function () {
...
 
 Router.map(function() {
+
+  this.route("settingsWizard", {
+    path: "/settingsWizard",
+    layoutTemplate: "homeLayout",
+    render: "settingsWizard"
+  });
+
+  this.route("settingsComplete", {
```
brylie: Lets deprecate this route and the accompanying template. That will simplify the user-flow and our codebase.

server/methods/settings.js
```diff
@@ -12,5 +12,17 @@ Meteor.methods({
     var apiUmbrellaHost = apiUmbrellaBaseUrl.host();
 
     return apiUmbrellaHost;
+  },
+  'isSettingsObjectValidAndEmpty': function() {
```
brylie: Consider how this function name reads in an if statement, in plain english:

> if is settings object valid and empty, then do something

Consider how the function will be used, and update the name so it is grammatical. E.g. `settingsObjectIsValidAndEmpty`

> if settings object is valid and empty, then do something

server/publish/settings.js
```diff
@@ -14,3 +14,8 @@ Meteor.publish("singleSetting", function (setting) {
 
   return cursor;
 });
+
+Meteor.publish('initialSetupComplete', function() {
+  const settings = Settings.findOne();
+  return settings.initialSetupComplete;
```
brylie: Publications need to return a database cursor, as far as I recall. This publication returns a string, so may not work as expected. Remove if unused.

server/startup.js
```diff
@@ -2,60 +2,59 @@ Meteor.startup(function () {
...
   try {
 
-    // Store settings object
-    var settings = Meteor.settings;
-    // Check if something is already in collection
-    if ( ! Settings.findOne() ) {
-      // if not insert settings object
-      Settings.insert(settings);
-    }
+
```
brylie: Cleanup unused code.

server/startup.js
```diff
@@ -2,60 +2,59 @@ Meteor.startup(function () {
...
+        // if not insert settings object
+        Settings.insert(settings);
       }
-    });
-    Meteor.call("syncApiUmbrellaUsers");
-    Meteor.call("syncApiBackends");
+    }*/
+
+    // Check if Settings collection has existing API Umbrella settings
+    if (Settings.findOne) {
```
brylie: This looks like a syntax error. It should probably read:

```js
const settings = Settings.findOne();

if (settings) {}
```

brylie:
This is really close. Read through the comments, most of which are minor changes.

Be sure of the following:

- remove any unused code, comments, console.log, etc
- read your code in plain English, to check that it is grammatical
- make sure comments are meaningful and accurate; try to describe the intended goal rather than the code itself

Nazarah:
Now, if a mandatory field is not filled up while User configs the APIs, there are errors in loading information in the analytic view. Make sure the configuration fields are mandatory, so that the user does not leave them empty.

Nazarah:
If users are clicking save by leaving some config fields empty, make sure Save operation does not take place. Instead alert text appears under the fields which are left empty. (CSS alert class texts)

Nazarah:
@Alapan: can you please look at the last 2 comments of mine and give your opinion.
Would they be fixed in this PR or we should add them as enhancements?

Alapan:
@Nazarah I think it is necessary to let the user know which fields are mandatory (e.g. by having an asterisk besides the field name), not just for submitting the form but also for displaying analytics data. Since this issue is about creating a wizard and not really form validation, maybe it is better to have the latter as an enhancement.

bajiat:
@alapan It is ok to make the as enhancements. Can you still check Brylie's comments (2 days ago) so that the PR can be merged?

Alapan:
@bajiat @brylie The changes were made. Please let me know of further changes.

frenchbread:
@Alapan Please resolve the conflicts.

Alapan:
@frenchbread Done.

jykae:
@frenchbread As discussed I assign myself to check out this.

frenchbread:
Thanks @Alapan @jykae

jykae:
@Alapan Reviewing

jykae:
@Alapan This works and looks good. I am merging in.

Editor and Mail cannot be enabled because some validation bugs in the form. Discussed with @bajiat We have issue #961 for email and raised now in planning. Editor is then another case what we decide to do with that in future.

Issue #939, opened 2016-05-20
API Backend Feedback form not submitting (after 1.3 upgrade)


brylie:
While working on the Meteor 1.3.x upgrade, the API Backend Feedback form does not appear to be working.

Steps to reproduce:
1. Add an API backend
2. View the API backend page
3. Click the "Feedback" tab
4. Click "Add Feedback"
5. Fill in, and submit the form

# Related
Issue #935 - update Meteor

bajiat:
@frenchbread Can you take a look?

frenchbread:
@bajiat Self-assigned

frenchbread:
Reopening this issue since it still exists:
```
Exception while invoking method 'sendEmail' RecipientError: Can't send mail - no recipients defined
```

brylie:
@frenchbread were you submitting the 'Contact' form or 'API Backend -> Feedback' form? The above error seems more relevant to the 'Contact' form, which is on the front page near the bottom.

frenchbread:
@brylie The bug was for the "Contact" form submit. Wrong issue..
Anyway the issue above it resolved with #970

PR #970, opened 2016-05-26
Feature/cleanup config file

frenchbread:
### Closes: 
- #939 
- #816 

### Changes
- Remove config files
- Fix feedback submit form


brylie:
@frenchbread is this ready for review?

frenchbread:
@brylie Yes. Please review.

server/methods/contactForm.js
```diff
@@ -12,7 +12,7 @@ Meteor.methods({
 
     // Send the e-mail
     Email.send({
-      to: Settings.findOne().contactForm_toEmail,
+      to: Settings.findOne().contactForm.toEmail,
```
brylie: What happens if this setting is not present? E.g. if the contact form 'to email' has not been provided in the settings collection? I suspect we will get an exception, which needs to be prevented or handled here.

frenchbread: Should we check if there is a value set in settings and if not, use email address from `settings.json`

PR #967, opened 2016-05-25
Feature/documentation tab

elnzv:
Closes #929 
Closes #1006
Closes #871
Closes #663
Closes #948
Closes #887
Closes #938
Closes #367
Closes #465 
Closes #1042

PR #972, opened 2016-05-30
Validate swagger

### LINKED ISSUE #523: Validate Swagger files when they are uploaded
Definition of done
---
* [x] Choose package for Swagger validation.
* [x] Integrate the package to the project.
* [x] Create a function to run the validation.
* [ ] Work with the developer working with issue to make sure the validation is run when a Swagger document is being uploaded.

jykae:
Closes #523 

## Proposed changes
- add swagger-parser package
- add Meteor method to validate Swagger file

Related to #967 

### DIFF
--- a/server/methods/swagger.js
+++ b/server/methods/swagger.js
@@ -1,0 +1,19 @@
+import SwaggerParser from 'swagger-parser';
+
+Meteor.methods({
+  // Validate Swagger JSON/YAML
+  // Params: URL or file path to Swagger file
+  // Returns: true if valid else false
+  validateSwagger: function(url) {
+     SwaggerParser.validate(url)
+      .then(function(api){
+        console.log("Your API is valid!");
+        console.log("API name: %s, Version: %s", api.info.title, api.info.version);
+        return true;
+      })
+      .catch(function(err){
+        console.log(err);
+        return false;
+      });
+  }
+});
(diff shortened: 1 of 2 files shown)


server/methods/swagger.js
```diff
@@ -0,0 +1,19 @@
+import SwaggerParser from 'swagger-parser';
+
+Meteor.methods({
+  // Validate Swagger JSON/YAML
+  // Params: URL or file path to Swagger file
+  // Returns: true if valid else false
+  validateSwagger: function(url) {
+     SwaggerParser.validate(url)
+      .then(function(api){
+        console.log("Your API is valid!");
```
brylie: We probably do not need to show any console logs, since this code runs on the server. We will use the return value of this function to show an alert to the user.

server/methods/swagger.js
```diff
@@ -0,0 +1,19 @@
...
+        console.log("API name: %s, Version: %s", api.info.title, api.info.version);
+        return true;
+      })
+      .catch(function(err){
+        console.log(err);
+        return false;
```
brylie: Can we return `valid` or `invalid` from this function? Right now it is ambiguous how to use this function, or what `true` or `false` means given the method name. E.g.

```js
// From client code
Meteor.call("validateSwagger", function (error, response) {
  if (error) {
    // handle the error
    // e.g. we can pass the validation message here as an error
  } else {
    if (response === 'valid') {
      // Swagger file is valid
    }
  }
});
```

jykae: I would not use here strings cause we have boolean value: "valid" or "invalid". I could though change method name to "isValidSwagger" if that implies the functionality better.

PR #975, opened 2016-05-30
Feature/check api key

elnzv:


bajiat:
Requires more work with higher level API settings. Not to be finished in the sprint, but will be continued in connection of improving API Umbrella integration / decoupling. Closing the PR, but leaving the branch.

PR #1018, opened 2016-06-01
Hotfix/dynatable i18n text duplication

### LINKED ISSUE #785: Data table headers on dashboard render two translations when changing language
When viewing the dashboard (Analytics) and changing the language, the data table headers show translations for both languages. Determine why the initial language translation is not removed.

frenchbread:
Closes #785 

### Changes:
- Replace dynatable with datatables

### Fixes:
- Language change


jykae:
@frenchbread Reviewing

jykae:
Original issue seems to be solved. However rendering in UI, timestamps, etc. still not very nice. Should we work it on this PR or what are your thoughts @frenchbread ?

jykae:
Decided to work on the CSS layout issues on this PR with @frenchbread 
Marked WIP.

frenchbread:
@jykae I am using [this package](https://github.com/Jowin/Datatables-Bootstrap3) here. The thing is that those options for pagination does not work. E.g `${'.datatable').dataTable()` - works, `$('.datatable').dataTable({"sPaginationType": "bs_two_button"})` - does not, gives an error.

brylie:
@frenchbread the [last commit on the Datatables package](https://github.com/Jowin/Datatables-Bootstrap3/commits/master) was Dec 11, 2014 @Jowin Jowin. Is there a more active package?

frenchbread:
@jykae I've added changes. Please review. The only issue I'm getting now is when formatting unix timestamp it gives wierd year value: e.g: `9/08/48387 11:10:01`

jykae:
🆒 checking 

@brylie good note about the package health and possible future support.

frenchbread:
@brylie @jykae Latest commits contain different package. It's still datatables but official and native one for NPM.

jykae:
Good work so far a bit beyond than the explanation in the PR. 
Let's return to this tomorrow, I will check the server time settings and whether Umbrella has any issues related to this. For me it looks like you are doing the conversion anyway right with momentjs.

frenchbread:
Removing `WIP`. Please review.

jykae:
Checked in the review together, merging. Good work @frenchbread

client/views/dashboard/charts/charts.js
```diff
@@ -201,23 +202,50 @@ Template.chartsLayout.created = function () {
...
-        records: setUpDataTable(),
-        perPageDefault: 10,
-        perPageOptions: [10, 20, 50, 100]
-      }
-    }).data('dynatable');
+    // Init datatable
+    initDatatable();
+
+    // Init datatable function
+    function initDatatable () {
```
brylie: Define this function in the `onCreated` callback, then call it from within this `onRendered` callback.

client/views/dashboard/charts/charts.js
```diff
@@ -201,23 +202,50 @@ Template.chartsLayout.created = function () {
...
+      });
+
+      // Initialize data table
+      $('.datatable').dataTable({
+        pagingType: 'simple'
+      });
+    }
 
     // Listens to filtering event and refreshes the table on a change
     function refreshTable() {
```
brylie: Please move this function to the template `onCreated` callback. I.e. we can start defining template functions as soon as the template instance is created, and then call the function from within `onRendered`.

frenchbread:
@brylie I think there is no need to move those functions because `refreshTable()` is wrapped into `instance.renderCharts()` which is already located in `onCreated` callback for initialisation. `instance.renderCharts()` is called from `instance.getDashboardData()` also in `onCreated`. But the `intance.getDashboardData()` itself with the entire functions is called from `onRendered`. So there should not be any rendering problems.

brylie:
Part of modularizing the code will be to move nested functions to the template level. E.g. our code is hard to maintain and refactor, in part, because functions are defined *and* called inside of other functions.

frenchbread:
@brylie Agreed. Created #1040 for that - feel free to extend def. of done. I'm thinking that we should move more to object-oriented approach there (to be able to easily reuse functions and add functionality if needed)

PR #1043, opened 2016-06-06
Get settings from collection

### LINKED ISSUE #1044: baseUrl undefined when creating new apiBackend
https://github.com/apinf/api-umbrella-dashboard/blob/develop/server/methods/settings.js#L9 causes error when creating API backend, getting settings from Settings collection so that we don't get undefined

jykae:
Closes #1044 

## Proposed changes

- gets settings from collection to fix undefined baseUrl on client when creating API backend

### DIFF
--- a/server/methods/settings.js
+++ b/server/methods/settings.js
@@ -1,3 +1,5 @@
+import { Settings } from '/both/collections/settings';
+
 Meteor.methods({
   'updateMeteorSettings': function() {
     // Updating Meteor.settings from Settings collection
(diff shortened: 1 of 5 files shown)


server/methods/settings.js
```diff
@@ -6,7 +8,8 @@ Meteor.methods({
   },
   'getApiUmbrellaHost': function () {
     // Get API Umbrella base url from settings object
-    var apiUmbrellaBaseUrl = new URI(Meteor.settings.apiUmbrella.baseUrl);
+    const settings = Settings.findOne();
+    var apiUmbrellaBaseUrl = new URI(settings.apiUmbrella.baseUrl);
```
brylie: Add a try ... catch here. Just in case the settings.apiUmbrella object does not exist.

jykae:
@brylie Added error handling as suggested here: http://docs.meteor.com/api/methods.html#Meteor-Error

jykae:
@frenchbread @brylie if you get regression of settings fixed, you can probably dump this PR.

PR #1046, opened 2016-06-06
Card View Version 1

Nazarah:
Added One version of cards and associated functionality UI


elnzv:
Shouldn't we keep docs in https://github.com/apinf/docs rather than here?

elnzv:
PR moved to [apinf/docs](https://github.com/apinf/docs/pull/8)

PR #1062, opened 2016-06-07
Hotfix/api umbrella methods sync

### LINKED ISSUE #1044: baseUrl undefined when creating new apiBackend

### LINKED ISSUE #1068: Refactor settings code to fix startup errors
When starting Apinf without settings, we noticed a few errors. Specifically, the following settings need validation before we attempt to use them in our code:

- API Umbrella web
- email
- Github

Audit our code to resolve any issues caused by non-existant settings.

### LINKED ISSUE #1069: Cleanup JSON routes code, since we are not rewriting swagger baseUrl anymore
In Sprint 23, we decided not to rewrite swagger documentation files, but rather to render them directly in Swagger UI. This differs from a previous decision, where we thought it important to route API test calls from Swagger UI through our proxy.

In line with the new decision, remove any code related to rewriting swagger files, such as the JSON routes code.

frenchbread:
# Fixes
- Closes #1044 
- Closes #1068 
- Closes #1069

- Api backends sync from ApiUmbrella


frenchbread:
@apinf/developers Please review & test

jykae:
@frenchbread Overall code and solution looked good. Settings file usage won't be anymore supported I suppose?

frenchbread:
Yes. But basically, as I recall correctly there were some feature that transfers settings data from setting file to collection.

brylie:
@frenchbread yes. We should probably still allow settings to be provided by settings.json, so in startup we could write the Meteor.settings object to the Settings collection. Everywhere else in the app, though, we should be getting settings from the Settings collection.

jykae:
Not syncing, should I wait 5 mins, or still bug?

lib/helperFunctions/verifyApiUmbrellaSettings.js
```diff
@@ -0,0 +1,14 @@
+export function apiUmbrellaSettigsValid (settings) {
```
jykae: Add letter 'n' for everywhere this function gets called and function name.

apiUmbrellaSettigsValid => apiUmbrellaSettingsValid

jykae:
Generally:
On startup with fresh database I guess we don't have Settings collection, only possible is settings file that could have been passed.

lib/helperFunctions/verifyApiUmbrellaSettings.js
frenchbread: Good catch.

jykae:
Decided together to leave settings file case outside of this.

frenchbread:
Added some more refactoring & checks

jykae:
Tested again after meteor reset, sync runs now without errors.

jykae:
Added one more close to the description. Merging and closing. 

If one minute syncing interval causes issues for example in production when there might be more latency let's make it longer.

Great co-working guys @frenchbread @brylie

frenchbread:
@jykae @brylie 👍

PR #1070, opened 2016-06-07
Hotfix/Contact form fix

frenchbread:
Closes #961 

## Fixes:
- Contact form is hidden if the email address is not provided

PR #1076, opened 2016-06-08
Remove custom validation, toggle mail fields

### LINKED ISSUE #961: Contact form email is not working in nightly
If a user fills the contact form on nightly.apinf.io and presses submit, the message never gets sent to the relevant email address specified in the settings form. If you try to enable the email in the settings form, the form complains about mailgun username and password.

To do: enable email in the settings, investigate what the form is expected for mailgun fields.

Also, check apinf.io settings as well.

Also, it would make sense to disable the contact form from landing page, if settings have not been provided.

jykae:
Closes #961 

Continue previous mail PR #1070 

## Proposed changes

- Remove custom validation that is not working
- Toggle field visibility on enabled


brylie:
Rather than removing custom validation, we need to make the field conditionally required. See the following example from the SimpleSchema docs:

- [Making a field conditionally required](https://github.com/aldeed/meteor-simple-schema#make-a-field-conditionally-required)

client/views/settings/settings.html
```diff
@@ -45,11 +45,13 @@ <h3 class="panel-title">Elasticsearch</h3>
         <div class="panel panel-default">
           <div class="panel-heading">
             <h3 class="panel-title">Mail</h3>
-            {{> afQuickField name='mail.enabled'}}
           </div>
           <div class="panel-body">
-            {{> afQuickField name='mail.username'}}
-            {{> afQuickField name='mail.password'}}
+            {{> afQuickField name='mail.enabled'}}
```
brylie: One design idea is for this toggle field to be in the panel header, and for the panel body to display only if this field is checked. E.g. if the mail is enabled, show the quick fields in the panel body. Otherwise, show only the panel header.

brylie:
@jykae see the following wireframe for the intended design:

single setting panel wireframe

jykae:
@brylie Thanks for the link and clarification. I will come back to this tomorrow.

jykae:
@brylie Got conditional check working, but cannot get design get behave that nicely.

brylie:
Try adding `style="display:inline"` to the checkbox element. You might need a parent div. Also, the Bootstrap `pull-right` task might be relevant here.

jykae:
Ok, I tried that "pull-right" but did not have "display:inline" rule. Let's see..

jykae:
Almost there with previous commit. @elnzv could you help with CSS in this PR when you have moment?

nayttokuva 2016-06-09 kello 11 44 59

client/views/settings/settings.less
```diff
@@ -24,3 +24,9 @@
     padding: 0;
   }
 }
+
+.settings-fields {
```
jykae: Great @elnzv it was exactly 4 LOC, if line breaks are not counted :D

jykae:
@brylie I would say this is ready to review now with changes.

brylie:
Lets also move the contact form email field to the Mail settings panel.

jykae:
@brylie ok, checkout

PR #1083, opened 2016-06-08
Add "Auth error" status code

### LINKED ISSUE #918: API status check giving false error message
## Behaviour
After adding API that requires authorization, API status shows: 
"API backend is down for some reason. Please contact support." 

Example: 
"Yle API uses app_id and app_key URL parameters for authentication, Apinf should be flexible for different kind of API authorizations. It's not good to show "API backend is down for some reason. Please contact support.", if we are getting 401 not authorized."

Above comment from issue: https://github.com/apinf/api-umbrella-dashboard/issues/917

## Expected
When getting 401, show info "API requires authentication" or allow 401 to be successful status.

frenchbread:
Closes #918 

# Changes
- Add `401` status code check


frenchbread:
@jykae Please review since you reported the bug.

jykae:
@frenchbread reviewing

jykae:
@frenchbread I think this still needs some more.. Committing soon to this PR, and plelase comment what you think.

frenchbread:
@jykae Agree but not sure what else we can add.. Some more status codes?

jykae:
@frenchbread what do you think? Not sure what to show on client side when getting 401

jykae:
Put WIP, so we are not merging yet, did not test each if-else statement that it works as intended.

frenchbread:
@jykae We could show different message depending on what status code is received but we are basically checking the status of the host. Do you think we should add custom message output depending on status code?

frenchbread:
@jykae Based on our discussion, I tried showing up the status code within the status panel but all the cases I've tested it returned "0" value so I decided to leave it as it is.

brylie:
There seems to be several categories of [HTTP status codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes).

We could show, for instance, red, yellow, or green based on the status code category and display the specific status code message on hover. One idea would be the following:

- 200s - green
- 300s - yellow
- 400s - yellow
- 500s - red

brylie:
@frenchbread where is the line of code where you are handling the server response? What does the response object contain?

frenchbread:
@brylie Status object returned from the server looks like [this](https://github.com/apinf/api-umbrella-dashboard/blob/ae6ee785fbe3b62fe50392ddec8a2a12bf1c94c6/client/views/api_backends/view_api_backend/details/status/status.js#L10).

I've tested displaying different kind of info depending on status code but the in most cases it has zero value.

brylie:
What are some example cases where you are getting a zero value? What value is zero?

brylie:
```js
HTTP.call("GET", "http://domain.com");
```

The above should return an object with a status code. That should work for non-existing domains as well, since it will throw an error.

brylie:
Is there a better JS utility for checking a website? E.g. one that would pass us any relevant errors?

brylie:
Also, check where you are trying to use the return value. Meteor methods do not return a value as you might expect from normal funcitons. Method values are returned in a callback funciton, wherein you can set a template reactive variable.

brylie:
Lets clean up the `isUp` code, since we can check the status code directly in the template.

jykae:
I was just about to write to check if result exists, and maybe do things on client in callback like @brylie says above.

brylie:
Also, we might not need a server method here, since we are not storing the data.

jykae:
That's also a good point :)

brylie:
Right, I just tried it, but it has some difficulties about cross-origin requests.

frenchbread:
Added some refactoring, cleanups & modified the layout of status:

screen shot 2016-06-09 at 13 37 06

So not it's super easy to reuse the status indicator, e.g. :
```
{{> viewApiBackendStatus apiBackend=apiBackend width="0.4"  }}
```
Since this is a circle, width==heigh. In `em`s.

@apinf/developers Please review/test/merge.

jykae:
All right, reviewing :)

jykae:
Great now it works and is gives feedback for user 👍 
Suggestion for text "Code 401. API requires authorization." 
I think "client error" is not that clear what it means if 401 is not clear at first for the user.

nayttokuva 2016-06-09 kello 13 51 31

apis/client/api_backends/view_api_backend/details/status/status.js
```diff
@@ -1,40 +1,61 @@
...
       //   errorMessage    : <String>
       // };
 
-      if (status.isUp) {
-        // updates layout with success status
-        $('#apiState')
+      // Init indicator element
+      const apiStatusIndicator = $('#api-status-indicator');
+
+      // Init regEx for status codes
```
jykae: Oh, wow, you use regexes for status codes, cool

brylie:
We would have to generate a help text for each status code, since the response object only contains the number.

frenchbread:
Added internationalisation.

jykae:
Cool. Tested, still works, merging.

PR #1088, opened 2016-06-08
Get back forgotten function from PR history

### LINKED ISSUE #1087: Add Api Backend wizard causes always error after recent methods changes for settings handling
After previous hotfixing for Settings handling and setup wizard, add API backend wizard causes always error because functions it tries to call were nested under one syncApiBackends server method.

Move functions needed to callable place and get Api Umbrella host from settings.

Check out this: https://github.com/apinf/api-umbrella-dashboard/blob/develop/client/views/api_backends/add/wizard.js#L80

jykae:
Closes #1087 

## Proposed fix

- get back forgotten function that add API backend wizard needs

### DIFF
--- a/server/methods/settings.js
+++ b/server/methods/settings.js
@@ -1,6 +1,16 @@
 import { apiUmbrellaSettingsValid } from '/lib/helperFunctions/validateSettings';
 
 Meteor.methods({
+  'getApiUmbrellaHostName': function () {
+    // Get API Umbrella base url from settings object
+    const settings = Settings.findOne();
+    var apiUmbrellaHostUrl = new URI(settings.apiUmbrella.host);
+
+    // Get host part of API Umbrella host URL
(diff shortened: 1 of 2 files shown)


jykae:
@frenchbread Went with smaller effort than I thought, only one forgotten function needed return from past :) 

Whoever reviews check if we would need IFs or such to avoid "undefined" cases.

server/methods/settings.js
```diff
@@ -1,6 +1,16 @@
 import { apiUmbrellaSettingsValid } from '/lib/helperFunctions/validateSettings';
 
 Meteor.methods({
+  'getApiUmbrellaHostName': function () {
+    // Get API Umbrella base url from settings object
+    const settings = Settings.findOne();
+    var apiUmbrellaHostUrl = new URI(settings.apiUmbrella.host);
```
brylie: What happens here if `apiUmbrella` is undefined?

PR #1131, opened 2016-06-13
Feature/api rating bookmark data model

### LINKED ISSUE #1130: Denormalize averageRating and bookmarksCount for API Backends
In order to sort the API Catalogue by average rating and bookmarks count, we may need to store the corresponding values in the ApiBackends collection. Create helpers and methods to calculate the average rating and bookmarks count. Make sure the helpers are called when a user bookmarks or rates an API Backend.

brylie:
Closes #1130 

# Proposed changes
- add `averageRating` and `bookmarkCount` fields to ApiBackends
- add `setAverageRating` collection helper for ApiBackends
- add `getAverageRating` and `getBookmarkCount` helpers for ApiBackends
- add `setAllApiBackendAverageRatings` method


catalogue/tabular.js
```diff
@@ -17,9 +17,11 @@ TabularTables.ApiTable = new Tabular.Table({
     },
     {tmpl: Meteor.isClient && Template.favourite, title: "Bookmark"},
     {
-     tmpl: Meteor.isClient && Template.apiBackendRating,
-     title: "Rating"
-    }
+      data: "averageRating",
```
jykae: Why is Tabular tables still hanging around here? You said to me that we switch it away and I have been creating table view on my PR. This does not make sense that you continued working with it and told me to do other way.

apis/server/methods/ratings.js
```diff
@@ -0,0 +1,20 @@
...
+    // Get all API Backends
+    const apiBackends = ApiBackends.find().fetch();
+
+    // Update the average rating value for each API Backend
+    apiBackends.forEach(function (apiBackend) {
+      // Set average rating value for current API Backend
+      apiBackend.setAverageRating();
+    });
+  },
+  setApiBackendAverageRating (apiBackendId) {
```
jykae: @brylie method name here could be something like set ApiBackendBookmarkCount

brylie: Which method? The `setApiBackendAverageRating` is related to the average rating field.

jykae:
@brylie Overall nice work, other thing to fix before I merge this, 
we need to set new field values on probably on startup, possibly by doing it as migration (running once). Or any other idea for that?

apis/server/methods/ratings.js
jykae: @brylie Ok, I confused this because console.log says about bookmarks. Could you also cleanup/fix console.logs?

jykae:
Ok, merging in, so I can integrate this to my table view. Thanks @brylie

PR #1140, opened 2016-06-14
Move API Status Indicator to right of header text

### LINKED ISSUE #1112: Move API status indicator to the right side of API Name
Referring the issue #918
There will be API Logo on left side of API name.
The current position of API status indicator would not be visible enough for this.

brylie:
Closes #1112 

# Changes
- move API Status Indicator to right of header text


jykae:
Looks really nice, do you agree @Nazarah ? I think we could merge this.

nayttokuva 2016-06-14 kello 11 49 25

PR #1185, opened 2016-06-20
Bugfix/documentation editor host undefined

### LINKED ISSUE #1057: Documentation Editor 'host' not defined
When updating the Documentation Editor setting, in either the wizard or settings page, the setting is saved. However, when viewing the documentation editor route, or checking for the documentation editor setting in helpers, we get an error in the console:

```
cannot read 'host' of undefined
```

Make sure we are looking for the right object in the settings collection, and that we have the proper subscription available when using the api documentation editor setting.

jykae:
Closes #1057 

## Proposed changes
- fix subscriptions for documentation editor
- show "Create documentation" section with link on API "Documentation" tab when Editor is enabled and host settings is set

### DIFF
--- a/both/collections/settings.js
+++ b/both/collections/settings.js
@@ -1,5 +1,11 @@
 Settings = new Mongo.Collection('Settings');
 
+SimpleSchema.messages({
+  regEx: [
+    {exp: SimpleSchema.RegEx.Url, msg: "[label] must be a valid URL"},
+  ]
+});
+
 Schemas.SettingsSchema = new SimpleSchema({
   apinf: {
     type: Object,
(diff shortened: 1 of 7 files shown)


jykae:
For reviewer: 
Documentation editor section would be available on "Documentation" tab of API in "Manage" modal when Editor settings are ok.

jykae:
Further notice: 
I will put "Enabled" style the same way as it is with "Mail", so that these settings look consistent. Marking WIP..

jykae:
@apinf/developers ready for review

frenchbread:
Reviewing this PR.

While testing, I noticed that when incorrect url is provided the modal window instantly hides and the error text is not visible. Should we continue showing modal if there were a validation error?

jykae:
@frenchbread Incorrect URL for editor in settings? I think it is admin responsibility to set up Apinf correctly. Or what incorrect URL you mean, something in modal?

frenchbread:
Works good. Merging.

As discussed, potentially we can provide a select box with existing and know editors rather than making a user fill the url by himself. Also currently, any url, not related to editor can be stored.

PR #1226, opened 2016-06-28
Fix empty API key for first user

### LINKED ISSUE #1216: Empty API key right after registration first user
First user get empty API key because API-Umbrella settings does not available when APInf try create user and get API key.

How to reproduce:

1) Make new installation of api-umbrella-dashboard without settings.json
2) Register first user and fill data in wizard.
3) Visit https://apiumbrellahost/profile

https___apinf_usubov_com_profile

Possible fix move "create user and get API key after user have done setup wizard".

jykae:
Closes #1216 

## Proposed changes
- change routing to show setup wizard as the first thing, onSuccess route first user to signUp
- change allow/deny rules for Branding, ProjectLogo & CoverPhoto (fixes Branding not saved issue)

### DIFF
--- a/both/collections/branding.js
+++ b/both/collections/branding.js
@@ -82,8 +82,17 @@ Schemas.BrandingSchema = new SimpleSchema({
 Branding.attachSchema(Schemas.BrandingSchema);
 
 Branding.allow({
-  insert: function (userId) {
-    return Roles.userIsInRole(userId, ['admin']) && Branding.find().count() === 0;
+  insert: function () {
+    // get branding
+    var brandingCount = Branding.find().count();
+    // if no branding exist
+    if ( brandingCount > 0 ) {
+      // don't allow insert
+      return false;
+    } else {
+      // insert
+      return true;
+    }
   },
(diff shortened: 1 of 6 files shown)


brylie:
The branding problem might be related to collection allow/deny rules. E.g. the Branding collection probably only allows administrator users to insert/update.

jykae:
@brylie Oh, good note, thanks. I'll check..

jykae:
Fixed issue with permissive allow/deny rules.

@brylie Checking "admin" role is good though, do you know if we can update allow rules for collection after setup wizard so after that it would allow only admins to manipulate Branding & Settings?

jykae:
Based on discussion after daily, I make new PR with with existing flow + add "Get Umbrella API key" button to "User profile" page to get key if it doesn't exist. This way we can keep database rules for Branding & Settings collections to allow only admin to insert/update for better security.

jykae:
@brylie This PR can be closed without merging, when new solution #1247 is accepted.

PR #1247, opened 2016-06-30
Another solution for missing API keys

### LINKED ISSUE #1216: Empty API key right after registration first user

jykae:
Closes #1216 

## Proposed changes
- add Meteor method for creating API Umbrella user & API key
- modify user profile page
- add i18n strings


jykae:
Based on discussion after today's daily:
This new solution does not change "setup wizard flow", but only proposes method for getting API key if missing.

On profile page I propose separate section for each proxy, as we are going towards multiproxy support, currently only API Umbrella section.

lib/i18n/en.i18n.json
```diff
@@ -16,6 +16,9 @@
   "signUp": "Sign Up",
   "apiCatalogue": "API Catalogue",
   "profile": "Profile",
+  "profile_usernameIsTaken": "Username is already taken",
+  "profile_UmbrellaHeader": "API Umbrella",
```
brylie: Camel case starts with lowercase in most other keys. Change for consistency.

server/methods/apiUmbrella.js
```diff
@@ -80,5 +80,62 @@ Meteor.methods({
...
+            "first_name": "-",
+            "last_name": "-",
+            "terms_and_conditions": true
+          }
+        };
+
+        // Try to create user on API Umbrella
+        try {
+          // Add user on API Umbrella
+          response = apiUmbrellaWeb.adminApi.v1.apiUsers.createUser(apiUmbrellaUserObj);
```
brylie: This is defined without `let` or `const`. Define with `const` if the reference will not change later.

server/methods/apiUmbrella.js
```diff
@@ -80,5 +80,62 @@ Meteor.methods({
...
+  // Might throw errors, catch on client callback
+  createAPIkeyForCurrentUser () {
+    // Get logged in user
+    const currentUser = Meteor.user();
+    // Check currentUser exists
+    if(currentUser) {
+      // Check apiUmbrellaWeb global object exists
+      if (apiUmbrellaWeb) {
+        // Create API Umbrella user object with required fields
+        apiUmbrellaUserObj = {
```
brylie: Add `const` here.

client/views/profile/profile.js
```diff
@@ -12,35 +12,61 @@ AutoForm.hooks({
...
+        copyButton.on("click", function () {
+          copyButton.text("Copied!");
+        });
+      }
+    });
+  }
 };
+
+Template.profile.events({
+  'click .get-umbrellaApiKey'(event) {
```
brylie: Use an ID rather than class, since this button is unique on the page.

client/views/profile/profile.html
```diff
@@ -7,13 +7,27 @@
...
       {{/autoForm}}
+      <div class="panel panel-default">
+        <div class="panel-heading profile-header">{{_ "profile_UmbrellaHeader"}}</div>
+        <div class="panel-body">
+          {{#if currentUser.profile.apiKey }}
+            <b>API Key:&nbsp;</b>{{ currentUser.profile.apiKey }}
+          {{ else }}
+            <!-- Get API key button -->
+            <p>
+              <button class="btn btn-primary get-umbrellaApiKey">
```
brylie: Use an ID rather than class, since this button is unique on the page.

server/methods/apiUmbrella.js
```diff
@@ -80,5 +80,62 @@ Meteor.methods({
         }
       });
     }
+  },
+  // Create API key & attach it for given user,
+  // Might throw errors, catch on client callback
+  createAPIkeyForCurrentUser () {
```
brylie: *Consistency*: rename to `createApiKeyForCurrentUser

We generally don't capitalize acronyms, such as API, in our function/variable names. Just camel case for first letters, where applicable.

server/methods/apiUmbrella.js
```diff
@@ -80,5 +80,62 @@ Meteor.methods({
...
+          // Update currentUser
+          Meteor.users.update({_id: currentUser._id}, {$set: fieldsToBeUpdated});
+
+          // Insert full API Umbrella user object into API Umbrella Users collection
+          ApiUmbrellaUsers.insert(response.data.user);
+        } catch (error) {
+          // Meteor Error (User create failed on Umbrella)
+          throw new Meteor.Error(
+            "umbrella-createuser-error",
+            "Failed to create user on API Umbrella."
```
brylie: Can we add i18n token here?

server/methods/apiUmbrella.js
```diff
@@ -80,5 +80,62 @@ Meteor.methods({
...
+          );
+        }
+      } else {
+        // Meteor Error (apiUmbrellaWeb not defined)
+        throw new Meteor.Error(
+          "umbrella-notdefined-error",
+          "API Umbrella not defined."
```
brylie: Add i18n token here, if possible.

server/methods/apiUmbrella.js
```diff
@@ -80,5 +80,62 @@ Meteor.methods({
...
+        );
+      }
+    } else {
+      // Meteor Error (User not logged in)
+      throw new Meteor.Error(
+        "apinf-usernotloggedin-error",
+        "Could not find logged in user."
```
brylie: Add i18n token here, if possible.

client/views/profile/profile.html
```diff
@@ -7,13 +7,27 @@
...
             {{_ "profile_UpdateButton"}}
           </button>
         </div>
       </div>
       {{/autoForm}}
+      <div class="panel panel-default">
+        <div class="panel-heading profile-header">{{_ "profile_UmbrellaHeader"}}</div>
+        <div class="panel-body">
+          {{#if currentUser.profile.apiKey }}
+            <b>API Key:&nbsp;</b>{{ currentUser.profile.apiKey }}
```
brylie: Add i18n token here.

client/views/profile/profile.js
```diff
@@ -12,35 +12,61 @@ AutoForm.hooks({
...
+    // Ask user to set username if it is not set.
+    if(!currentUser.username) {
+      const setUsernameMsg = TAPi18n.__("profile-setUsername");
+      sAlert.info(setUsernameMsg);
+    }
+    instance.autorun(function(){
+      // Show clipboard button only if API key exists
+      if(currentUser.profile.apiKey) {
+        // initializes button
+        var copyButton = $("<a class=\"btn btn-default btn-xs\" id=\"copyApi\"> Copy API key to clipboard</a>");
```
brylie: Add i18n token here.

jykae:
@brylie Requested fixes done. Could you help with the clipboard, thanks.

brylie:
When clicking 'Finish', the settings wizard has a noticeable delay, e.g. 2-3 seconds. Has the form submit hook logic changed? Perhaps route the user before doing API Umbrella related steps.

brylie:
When clicking "Get API Key for Umbrella" there is a significant delay with no indication that anything is happening. Lets add a spinner or some other indicator, e.g. grey out the button, so that the user does not grow impatient or become confused.

jykae:
Haven't touched it, could be another thing

brylie:
Since we are working with the Profile component, it is a good opportunity to move it to the root level. This is what we have done with the catalogue, bookmarks, ratings, and API Backends. E.g. we are incrementally refactoring our project to have components at the root level.

brylie:
Please pull changes before making any other changes. I refactored the code by moving the profile code to a 'users' component.

lib/i18n/en.i18n.json
```diff
@@ -16,6 +16,12 @@
   "signUp": "Sign Up",
   "apiCatalogue": "API Catalogue",
   "profile": "Profile",
+  "profile_apiKey_copyButtonText": "Copy API key",
+  "profile_apiKey_copySuccessful": "API key copied to clipboard.",
+  "profile_usernameIsTaken": "Username is already taken",
+  "profile_umbrellaHeader": "API Umbrella",
```
brylie: We should probably make this text more generic. E.g. API Key(s). This is similar to how we don't use the word "Swagger" when referring to API Docs.

users/profile/client/profile.js
```diff
@@ -0,0 +1,54 @@
+AutoForm.hooks({
+  updateProfile: {
+    onSuccess: function(operation, result, template) {
+      return sAlert.success('Profile updated');
```
brylie: Lets add i18n tokens here, since we are editing this file.

jykae:
@brylie we are done.

PR #1298, opened 2016-07-06
Feature/analytics drilldown

### LINKED ISSUE #1000: Create option(s) for analytics drilldown
# User story

> As an API owner 
> I want to drill down to the results of the dashboard
> so that I can narrow down the results shown in charts based on my needs.
# Design
- Drilldown information to show by table. 2 styles, but same functionality

apitable1

apitable2-drodown

apitable2-hovered
# Definition of done

_to be further defined_
- [ ] API owner can view analytics 
  - [ ] per IP address
  - [x] per HTTP method 
  - [ ] per IP location (country, city?)
- [ ] API owner can apply more than one rule to the custom query
- [ ] API owner can return to a situation where  custom rules / no rules are applied with one single action

frenchbread:
Closes #1000


frenchbread:
Partly moved to #1329.

PR #1334, opened 2016-07-19
Feature/catalog pagination

### LINKED ISSUE #1196: Add infinite scrolling or pagination to Catalog view
Add functionality that only publishes (?) a performance-wise feasible number of API backends to API catalog container. More/Next backends are published when user takes an action. Most likely options for the user action are infinite scrolling or catalog pagination.

# User story
>  As an Apinf architect
I want catalog view to load only a reasonable amount of APIs when user goes into the view
so that the performance of the platform does not suffer from too many APIs being published to front-end

# Definition of done
* [x] Select either infinite scrolling or pagination as the solution
  * [x] Document the pros and cons of each option in the comments for this issue
* [x] Implement the solution so that UI does not get stuck even if the number of APIs stored in database would be high

# Wireframe

* Pagination for Catalog and Table View

paginationcardentries

paginationtable entries

* Infinite Scrolling

infinitescrollingcard

infinitescrollingtable

frenchbread:
Closes #1196

### DIFF
--- a/catalogue/client/catalogue.css
+++ b/catalogue/client/catalogue.css
@@ -21,3 +21,16 @@
 .catalogue-main-title {
   margin-left: 0.63em;
 }
+
+/* Class to make pagination buttons inactive */
+.inactive a {
+  color: #d3d3d3 !important;
+  cursor: default;
+  background-color: #fafafa !important;
+}
+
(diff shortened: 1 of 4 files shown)


frenchbread:

PR #1335, opened 2016-07-19
Feature/unique api backend

### LINKED ISSUE #1200: Allow only unique proxy backend base paths
In order to avoid having the same API added to the catalog more than once, we should only allow unique API backends. (As a follow-up, we may need to give an option to API owners to report, if someone else have added their API)

# Definition of done
* [ ] User can only add an API with Add API Wizard, if the front-end prefix (proxy backend base path) is unique?
* [ ] If the user tries to add a pre-existing URL (with pre-existing frontend prefix?), they get an error saying that this API backend already exists.

nnn:
Closes #1200 
For now it's using description field as an example.
PR is WIP.


apis/client/add/wizard.js
```diff
@@ -48,6 +48,7 @@ Template.addApiBackendWizard.helpers({
       },
       "url_matches.frontend_prefix": {
         optional: true,
+        unique: true,
```
brylie: We probably don't need to check the uniqueness in this schema, since it is not attached to the collection. This means the Wizard may not have a 'sense' of what has already been entered into the collection.

brylie:
I can submit the wizard twice using the same frontend prefix for both APIs. The second API is not saved, even though a status message appears saying that the API was created successfully.

Basically, the Add API Backend wizard needs to perform validation before submitting.

brylie:
The Edit API Backend form works as expected, displaying a validation error when attempting to use an existing frontend prefix.

```
 This value is already in use, please make it unique.
```

brylie:
Also, when submitting the Add API Backend wizard form twice with the same frontend prefix, the API is saved both times on API Umbrella, even though it is rejected from Apinf. Double check that API Umbrella does not receive API backends with duplicate frontend prefixes, when submitting the Add API Backend wizard.

nnn:
We have changed the way we keep API schemas, closing this PR.
Will open new PR for #1200.

PR #1339, opened 2016-07-21
Fix references in docker-compose setup instructions.

shaliko:


jykae:
@shaliko This is obviously instructions fix, I merge in.

PR #1342, opened 2016-07-25
Fix disable API key field

### LINKED ISSUE #622: Ensure api managers can disable API key check on Backends
Original report (by @kyyberi)
----
Even if I selected that API key is optional (and also selected unlimited requests), when testing the API response was: 
`API_KEY_MISSING`

Definition of done (by @bajiat, by @brylie modified due to Multiproxy work)
---
* [ ] Make sure the disable api key field in added to and stored in Proxy backend collection
* [ ] Make sure the properties of the field are aligned with Api Umbrella
* [ ] Make sure the field is passed correctly to the Admin REST API
* [ ] Make sure it gets stored properly in API Umbrella
* [ ] There needs to be setting in the Proxy tab in API profile

jykae:
Closes #622 

## Proposed changes
- fix form to have "disable_api_key" under Settings object & correct field type (Boolean -> String)
- allow 3 options that Umbrella expects for "disable_api_key" field

### DIFF
--- a/apis/collection/backend.js
+++ b/apis/collection/backend.js
@@ -5,6 +5,17 @@ SimpleSchema.RegEx.Port = new RegExp(/^[0-9]{1,5}$/);
 SimpleSchema.RegEx.Prefix = new RegExp(/^\/[a-z0-9A-Z_\-\/]*$/);
 
 Schemas.Settings = new SimpleSchema({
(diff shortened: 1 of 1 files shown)


jykae:
@nnn please review, field should work now, place in form & style possibly need some fix.

PR #1372, opened 2016-08-09
Feature/api visibility

### LINKED ISSUE #1020: Setting for API visibility
Add a setting / option for API owner so that they can hide their API backend from catalog. This is one of the first steps from supporting not only open APIs but also closed APIs.

_Open issue: Should this setting be available in API profile view or in Edit API settings view?_

Related issue: #1207

# User story
> As an API owner 
I want to be able to select whether my API is publicly visible in the catalog or not

# Definition of done
* [ ] API owner has an option to select API visibility in catalog: visible or hidden (UI widget)
* [x] The API owner and Apinf administator(s) can easily see what is the current/selected visibility setting of API 
* [x] If API is hidden by owner, only administrators and the particular owner can see it in the catalog. It is hidden for all other users.
* [x] Not authorized users cannot view  the data: they will get a not authorized message

# Wireframe

apivisibilityvisible

apivisibilityhidden

jykae:
Closes #1020 

## Proposed changes
- add visibility field
- add visibilitySwitch to API view

Also fixes undefined initial value for bookmarkCount


jykae:

jykae:
Found issue with @brylie in router "apiBackend.findOne" call, not yet able to solve. Happens when creating Api Backend.

apis/client/latest_api/latestApi.html
```diff
@@ -7,6 +7,7 @@ <h2 class="feature-section-heading">{{_ "latestApiBackends_Heading" }}</h2>
     </div>
   </div>
   {{#each latestBackends }}
```
nnn: We need to change here to publishing only public apis.

apis/collection/helpers.js
```diff
@@ -83,6 +83,15 @@ ApiBackends.helpers({
     // Otherwise, get average rating
     return this.averageRating;
   },
+  currentUserCanView () {
+    // Check visibility
+    if(this.visibility) {
```
nnn: Should be isPublic

brylie:
This is ready for review.

apis/client/latest_api/latestApi.html
brylie: I edited the publications so that only public APIs are published.

bajiat:
@frenchbread Could you review this PR?

frenchbread:
Code looks good!

Tested & noticed that API is not showing up on `My bookmarked APIs` tab. (I think it's logical to show to a user his API wether it's private or not)

frenchbread:
Also FYI, was not able to create new user to test with different permissions. Tested with `no logged in user`

brylie:
@frenchbread, I am able to add multiple users. Perhaps do `meteor reset`? What are steps to reproduce the second user issue?

brylie:
> Tested & noticed that API is not showing up on My bookmarked APIs tab. (I think it's logical to show to a user his API wether it's private or not)

I have added a change that shows private bookmarked APIs, if the user is a manager of the API and has also bookmarked it. Private APIs should only show up for managers, regardless of whether a user has previously managed to bookmark the API.

frenchbread:
>  I am able to add multiple users. Perhaps do meteor reset? What are steps to reproduce the second user issue?

@brylie Ran `meteor reset` & even got new keys, did not solve the issue.

frenchbread:
Thanks for fixes. This looks good. Merging.

Issue #1375, opened 2016-08-09
Data migration strategy


philippeluickx:
How do we handle data migrations when moving forward? Do we still make breaking changes? Are migration files needed when changing data models?

# Definition of done
* [x] Options for migration have been studies and the optimal solution has been selected
  * [x] Options are documented in the comments for this issue
*  [x] There is a Meteor or npm package or migration script that handles database migrations

bajiat:
Multi-proxy is a prime example of a disruptive, backwards incompatible change on the database level. This is why I am labeling this issue with Multiproxy label.

jykae:
Option 1: General mongodb-migrations framework, not only for Meteor platform. https://www.npmjs.com/package/mongodb-migrations

philippeluickx:
@jykae Is there any reason not to take the meteor standard package?

jykae:
@philippeluickx well, I understood that Meteor is not providing maintained package. They let community to maintain a solution, https://github.com/percolatestudio/meteor-migrations <-- Checkout the beginning of README. 
I think taking general use migration tool is more likely to stand the test of time. Meteor itself is moving to npm just because Node.js/NPM is larger community.

jykae:
Basically, they are probably trying the same Django was doing years, before they got lots of feedback about missing migration solution & noticed that they must provide maintained solution for that. With this I mean moving from South to Django migrations in 1.7

philippeluickx:
@jykae  Good arguments, thanks for sharing. I am ok with mongodb-migrations.

brylie:
@jykae great work! Lets see what we can do with mongodb-migrations.

bajiat:
@jykae Thanks for the research. I will close this issue and let's discuss concrete steps to take the mongodb-migrations into use,

jykae:
@bajiat reopening because had troubles get mongodb-migrations working, percolate:migrations might still be best bet for migrations. And selecting package is not yet strategy imho.

bajiat:
@jykae I'm fine with reopening. Please investigate to come up with the best option for database migration.

jykae:
I propose we take percolate:migrations back. It's easy to use and works also for Meteor 1.4. Example migration for adding isPublic field if it's missing, https://github.com/apinf/api-umbrella-dashboard/pull/1411

Strategy requires also a mutual decision of process how we handle writing of required migration function when a new feature requires it. It should not be one person fight, who manages the version upgrade into production.

[...]

PR #1411, opened 2016-08-19
Add migration for isPublic field

jykae:
## Proposed changes

- add percolate:migrations package
- add isPublic field migration
- run migrations on startup

This is for preview related #1375 
To be updated based on new schemas.


server/startup.js
```diff
@@ -1,6 +1,8 @@
 Meteor.startup(function () {
   try {
-
+    // Run migrations
```
brylie: This file should go in `core/server/startup.js`. We are trying to deprecate the `/server/` folder.

jykae:
@brylie Moved file to requested location.

jykae:
@bajiat @brylie As discussed in daily today I propose closing this, schemas change and imho it's not worth effort to do migration for old architecture design. As noted above, this is preview of using percolate:migrations

PR #1429, opened 2016-08-22
Integrate sdk generator

### LINKED ISSUE #1376: Integrate SDK download to Documentation tab
Integrate the SDK generator UI created in #1225 into the Documentation tab as a section of its own, thus allowing API consumers to download an SDK in the selected programming language, if a valid Swagger file has been updated for an API

# Definition of done
* [ ] SDK UI is only available in Documentation tab, if a valid Swagger file has been uploaded for an API. 
* [ ] The SDK UI passes the Swagger file to the Codegen API without the user having to select the file (In other words, the particular Swagger file on this tab gets passed to the Codegen API)

# Wireframe

sdkgenerator
documentationpage2

marla-singer:
add modal form, add button "SDK Generator", add list of language

Closes #1376

### DIFF
--- a/documentation/codegenerator/client/codegenerator.js
+++ b/documentation/codegenerator/client/codegenerator.js
@@ -4,7 +4,8 @@ import { Session } from 'meteor/session';
 
 import _ from 'lodash';
 
-import { DocumentationFiles, languageHashName } from '/documentation/collection/collection';
+import { DocumentationFiles } from '/documentation/collection/collection';
+import { languageHashName } from './hashName'
 
 Template.sdkCodeGeneratorModal.onCreated(function () {
   const instance = this;
--- a/documentation/codegenerator/client/codegenerator.js
+++ b/documentation/codegenerator/client/codegenerator.js
@@ -1,15 +1,16 @@
+import { Template } from 'meteor/templating';
+import { ReactiveVar } from 'meteor/reactive-var';
+import { Session } from 'meteor/session';
 import { HTTP } from 'meteor/http';
+
 import { LanguageParams, LanguageList, DocumentationFiles } from '/documentation/collection/collection';
-import { ApiBackends } from '/apis/collection/backend';
 
 Template.sdkCodeGeneratorModal.onCreated(function () {
   const instance = this;
 
-  instance.autorun(function () {
-    const apiBackend = ApiBackends.findOne(instance.data.apiBackend._id);
-    // Save apibackend id
-    Session.set('currentApiBackend', apiBackend);
+  instance.callRequest = new ReactiveVar(false);
 
+  instance.autorun(function () {
     // Get documentation file id
     const documentationFileId = instance.data.apiBackend.documentationFileId;
 
@@ -23,13 +24,12 @@ Template.sdkCodeGeneratorModal.onCreated(function () {
 
 Template.sdkCodeGeneratorModal.onDestroyed(function () {
   // Unset session
-  Session.set('currentApiBackend', undefined);
   Session.set('currentDocumentationFileURL', undefined);
 });
 
 Template.sdkCodeGeneratorModal.helpers({
   // Schema for SDK Code Generator form
-  generateSDK: function () {
+  generateSDK () {
     var sdkSchema = new SimpleSchema({
       selectLanguage: {
         type: String,
@@ -42,11 +42,27 @@ Template.sdkCodeGeneratorModal.helpers({
       }
     });
     return sdkSchema;
+  },
+
+  // Give variable callRequest to template
+  statusRequest () {
+    // Get reference to template instance
+    const instance = Template.instance();
+
+    return instance.callRequest.get();
+  },
+  // Give variable callRequest to autoform as a parameter
+  getCallRequest () {
+    // Get reference to template instance
+    const instance = Template.instance();
+
+    return instance.callRequest;
   }
+
 });
 
 AutoForm.addHooks('downloadSDK', {
-  onSubmit: function (formValues) {
+  onSubmit: function (formValues, updateDoc, callRequest) {
     // Prevent form from submitting
     this.event.preventDefault();
 
@@ -74,6 +90,9 @@ AutoForm.addHooks('downloadSDK', {
       'swaggerUrl': pathToFile
     };
 
+    // Start spinner when send request
+    callRequest.set(true);
+
     // Send POST request
     HTTP.post(url, { data: options }, function (error, result) {
       // Get information from Swagger API response
@@ -91,6 +110,8 @@ AutoForm.addHooks('downloadSDK', {
         // Otherwise show an error message
         FlashMessages.sendError(response.message);
       }
+      // Finish spinner
+      callRequest.set(false);
     });
   }
 });
--- a/documentation/codegenerator/client/codegenerator.js
+++ b/documentation/codegenerator/client/codegenerator.js
@@ -1,11 +1,11 @@
 import { Template } from 'meteor/templating';
 import { ReactiveVar } from 'meteor/reactive-var';
-import { Session } from 'meteor/session';
+import { HTTP } from 'meteor/http';
 
 import _ from 'lodash';
 
 import { DocumentationFiles } from '/documentation/collection/collection';
-import { languageHashName } from './hashName'
+import { specificLanguageNames } from './codgeneratorSpecificLanguages';
 
 Template.sdkCodeGeneratorModal.onCreated(function () {
   const instance = this;
@@ -16,11 +16,8 @@ Template.sdkCodeGeneratorModal.onCreated(function () {
   // Get documentation file id
   const documentationFileId = instance.data.apiBackend.documentationFileId;
 
-  // Get documentation file URL
-  const documentationFileURL = Meteor.absoluteUrl().slice(0, -1) + DocumentationFiles.baseURL + '/id/' + documentationFileId;
-
   // Save documentation file URL
-  instance.documentationFileURL = documentationFileURL;
+  instance.documentationFileURL = Meteor.absoluteUrl().slice(0, -1) + DocumentationFiles.baseURL + '/id/' + documentationFileId;
 
   /* Get list of an available languages from Codegen server */
 
@@ -40,7 +37,7 @@ Template.sdkCodeGeneratorModal.onCreated(function () {
 
     _.forEach(response, function (language) {
       // Check on specific name
-      let newLanguageName = languageHashName[language];
+      let newLanguageName = specificLanguageNames[language];
 
       // Convert name by standard method if it isn't specific name
       if (_.isUndefined(newLanguageName)) {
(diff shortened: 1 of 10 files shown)


bajiat:
@marla-singer Is this PR ready for review?

marla-singer:
@bajiat Yes

marla-singer:
@bajiat Now the file is downloaded slow because of the cross domain request. I suggest to add a spinner for user's usability

brylie:
@marla-singer that sounds like a good enhancement to add user feedback. Will you include it in this PR?

marla-singer:
@brylie yes, sure

bajiat:
@marla-singer I suggest that we add a Work in progress (WIP) label to this PR until the spinner has been added. Please remove the label when you are done with adding the spinner and the PR is ready for review.

documentation/view/client/documentation.css
```diff
@@ -12,3 +12,7 @@ iframe {
 .s-alert-box {
   z-index: 10000
 }
+
+button#sdk-code-generator {
+  margin-right: 10px
```
brylie: Please use relative values for layout: `em`, `rem`, or `%`

documentation/view/client/documentation.html
```diff
@@ -2,13 +2,24 @@
...
-          class="btn btn-xs btn-info pull-right">
-          <i class="fa fa-pencil"></i>&nbsp;
-          Manage
-        </button>
-      {{/if}}
+          <button
+            id="manage-api-documentation"
+            class="btn btn-xs btn-info pull-right">
+            <i class="fa fa-pencil"></i>&nbsp;
+            Manage
```
brylie: Please add i18n token here. E.g.

```js
{{_ "documentation_manageDocumentation_buttonText" }}
```

documentation/codegenerator/client/codegenerator.js
```diff
@@ -0,0 +1,124 @@
...
+        // Otherwise show an error message
+        FlashMessages.sendError(response.message);
+      }
+      // Finish spinner
+      callRequest.set(false);
+    });
+  }
+});
+
+FlashMessages.configure({
```
brylie: Remove this configuration, as it duplicates the global FlashMessages configuration.

documentation/codegenerator/client/codegenerator.js
```diff
@@ -0,0 +1,124 @@
...
+  getCallRequest () {
+    // Get reference to template instance
+    const instance = Template.instance();
+
+    return instance.callRequest;
+  }
+
+});
+
+AutoForm.addHooks('downloadSDK', {
```
brylie: Please move AutoForm hooks to their own file. Typically, we are calling this file `autoform.js`.

documentation/collection/collection.js
```diff
@@ -14,4 +14,21 @@ const DocumentationFiles = new FileCollection('DocumentationFiles', {
   ]
 });
 
-export { DocumentationFiles };
+// List of popular language for SDK Code Generator
```
brylie: How difficult would it be to get this list dynamically from the Codegen server? E.g. we wouldn't have to maintain a static list. Also, it is possible that different instances of the Codegen server would support different languages, e.g. an older version of Codegen might not support as many languages as a newer version.

documentation/codegenerator/client/codegenerator.js
```diff
@@ -0,0 +1,124 @@
...
+    const documentationFileURL = Meteor.absoluteUrl().slice(0, -1) + DocumentationFiles.baseURL + '/id/' + documentationFileId;
+
+    // Save documentation file URL
+    Session.set('currentDocumentationFileURL', documentationFileURL);
+  });
+});
+
+Template.sdkCodeGeneratorModal.onDestroyed(function () {
+  // Unset session
+  Session.set('currentDocumentationFileURL', undefined);
```
brylie: Try to use template instance variables rather than Session variables. You can add an instance variable like so:

```js
// Inside of Template lifecycle callback
const instance = this;

instance.variableName = "value";
```

Then you can use the instance variable inside of Template events and helpers. Use the [ReactiveVar package](https://docs.meteor.com/api/reactive-var.html), if you need the instance variable to be reactive.


```js
// inside of instance lifecycle callback
const instance = this;

instance.someReactiveVariable = new ReactiveVar();

instance.someReactiveVariable.set("value");
```

documentation/codegenerator/client/codegenerator.js
```diff
@@ -0,0 +1,124 @@
...
+    let languageHashName = LanguageParams[selectedLanguage];
+
+    if (_.isUndefined(languageHashName)) {
+        // Use standart method for creating parameter
+      languageHashName = selectedLanguage.toLowerCase();
+      languageHashName = languageHashName.replace(/\s+/g, '-');
+    }
+
+    // Create URL to send request
+    const url = 'https://generator.swagger.io/api/gen/clients/' + languageHashName;
```
brylie: This will need to be added to the `Settings` collection as part of this, or a later, pull request. We need to be able to configure the Codegen url through the user interface.

brylie:
When submitting the [petstore swagger example](http://petstore.swagger.io/v2/swagger.json), I get the following error in Chrome console:

```sh
POST https://generator.swagger.io/api/gen/clients/python 400 (Bad Request)
```

For clarification, I have downloaded the petstore example and uploaded it via the Manage documentation modal.

brylie:
Can you please provide an example Swagger file, so that we can test using the same file?

brylie:
When using the [swagger petstore example file](http://petstore.swagger.io/v2/swagger.json) we are also getting file validation errors, which may be incorrect:

screenshot from 2016-08-23 13-01-37

brylie:
Overall, this pull request looks good. The user interface is nice, and the code is easy to read. I have made some points about code structure and conventions, which hopefully won't add too much more work.

marla-singer:
@brylie thanks for you points! I fixed some remarks.

I think the problem with 400 request happens on localhost. 
You can test SDK codegen with swagger petstore example file in [this page](https://apinf.2glab.com/api/v6ANXuvsKHpWcrDhi) 
Before I used  [this standard example](http://editor.swagger.io/#/) both JSON and YAML format

brylie:
That makes sense, the localhost path to the swagger file wouldn't be accessible to the Codegen server. I didn't think about that before.

brylie:
The error message, in any case, should be more specific to the type of error.

marla-singer:
Okay, i will think about error message
I am implementating dynamically list  from the Codegen server. I tried to convert the response to more friendly names of languages and now list of languages, which users will see in dropdown list, looks like this 
```
["Akka scala", "Android", "Async scala", "Clojure", "Cpprest", "Csharp", "Csharpdotnet2", "Cwiki", "Dart", "Dynamic html", "Flash", "Go", "Groovy", "Html", "Html2", "Java", "Javascript", "Javascript closure angular", "Jmeter", "Objc", "Perl", "Php", "Python", "Qt5cpp", "Ruby", "Scala", "Swagger", "Swagger yaml", "Swift", "Tizen", "Typescript angular", "Typescript angular2", "Typescript fetch", "Typescript node"]
``` 
Is it okay?

bajiat:
@marla-singer @brylie What if you get the list dynamically and then have a function/method that replaces the names to a more user friendly name? If the list available from Codegen server has changed and there is no corresponding name in your function, then it will be displayed as is (= as retrieved from server)?

marla-singer:
@bajiat Allright, I've got it

marla-singer:
The task can be tested on [this page](https://apinf.2glab.com/api/v6ANXuvsKHpWcrDhi)

brylie:
@marla-singer looks good. We will demonstrate it briefly at our meeting this morning.

documentation/collection/collection.js
```diff
@@ -15,3 +17,52 @@ const DocumentationFiles = new FileCollection('DocumentationFiles', {
...
+// Get list of an available languages from Codegen server
+HTTP.get(url, {}, function (error, result) {
+  // Get information from Swagger API response
+  const response = JSON.parse(result.content);
+
+  // Save response to use it like url parameter in POST request
+  urlParameters = response;
+
+  // Create list of friendly language names
+  LanguageList = [];
```
marla-singer: @brylie I used global variables in the file collection.js (LanguageList, urlParameters) But I noticed you have done a restructure project in pull request #1437, where it is necessary to use export/import in the collection files. I can't use import because of the async callback in the GET request that returns data.
Now the data processing from Swagger is executed in main file (codegenerator.js) instead of collection.js and  Modal form is waiting for data from the callback before displaying a dropdown list with the languages. I added a spinner, because it is not immediately

Can you please give advice to improve this solution?

brylie:
@marla-singer what import do you need? Is it a collection that you cannot import?

documentation/codegenerator/client/codegenerator.js
```diff
@@ -1,9 +1,8 @@
 import { Template } from 'meteor/templating';
 import { ReactiveVar } from 'meteor/reactive-var';
 import { Session } from 'meteor/session';
-import { HTTP } from 'meteor/http';
 
-import { LanguageParams, LanguageList, DocumentationFiles } from '/documentation/collection/collection';
+import { DocumentationFiles } from '/documentation/collection/collection';
```
marla-singer: @brylie File codegenerator.js 
If I do import the variables 'LanguagesList' and 'LanguageParams' from collection.js, this variables wil be undefined.

brylie:
Try placing them in their own file(S). Collection.js should only contain code related to the collection.

marla-singer:
Then the code is ready for reviewing

documentation/collection/collection.js
```diff
@@ -14,4 +14,21 @@ const DocumentationFiles = new FileCollection('DocumentationFiles', {
   ]
 });
 
-export { DocumentationFiles };
+// The most part of list items can convert to user friendly name by standard method
+// The method is to replace dash on space and to do the capital letter of each word
+// Some language names can't be convert by standard method.
+// Variable 'languageHashName' keeps specific names
+const languageHashName = {
```
brylie: Move this to a different file. The only purpose of the `collection.js` file is to define and export the collection.

marla-singer:
I should add this constant to **codegenerator.js** in part .onCreated or like helpers?

brylie:
It might be a good idea to keep it in a separate file. E.g. `codegenarator_languages.js`

brylie:
All other code here looks good, and we can merge these changes once the codegenerator languages file/import is working.

documentation/codegenerator/client/hashName.js
```diff
@@ -0,0 +1,17 @@
+// The most part of list items can convert to user friendly name by standard method
+// The method is to replace dash on space and to do the capital letter of each word
+// Some language names can't be convert by standard method.
+// Variable 'languageHashName' keeps specific names
+const languageHashName = {
```
brylie: Keep this in the same file as the language names. E.g. `codegenerator_languages.js`

brylie:
Just checking on this PR. We can merge it if the files are in the desired location. E.g. hash names and language strings. Is `hashName.js` the desired file name for the hash lookup map?

marla-singer:
I renamed the file and variable and gave more clearly names. Another files are in the desired location

PR #1397, opened 2016-08-16
Feature/simplify api backends collection

brylie:
Closes #1091 
Closes #1379 

# Changes
- Rename `ApiBackends` collection to `Apis`
  - rename file to `index.js` for clean imports
- Remove API Umbrella fields from Apis schema
- Create Add API form
- Remove Add API Wizard
- Remove Initial Setup Wizard
- Edit basic settings/branding check to display modal dialogue to admin
- Cleanup unused packages
- Cleanup/rename/reorganize files along the way
- fix bugs related to user profile and user menu

PR #1431, opened 2016-08-22
Feature/apis schema add form

### LINKED ISSUE #1091: Design API backend schema
In order to decouple API backends from Api Umbrella, we will need a collection for API backend that is separate from the proxy settings. In other words, the schema needs to be changed and separated from proxies and their settings.

Idea: Align with Swagger specification or Open API specification (depending on the schedule of the latter). This might enable importing API backend configuration by importing Swagger (or Open API) file. Only after adding the API backend, the owner would assign a proxy to it.

Potential user story related to the Swagger idea:
>  As an API owner
I want to add my API by importing a Swagger file
so that I can reuse my Swagger file and do not need to type in the API details

As a minimum requirement, all fields that exist both in our collection and in Swagger spec should match.

# Definition of done
* [ ] Api Backend schema exists 
* [ ] Api Backend does not contain API Umbrella related settings
* [ ] When applicable, schema field(s) confirm to Swagger / Open API specification

### LINKED ISSUE #1379: Update Add API wizard
Update Add API wizard to add an API without proxy settings. (Proxy should be added - or in the future selected - in separate UI.) This UI will add an API with basic details to Catalog.

# Definition of done
* [x] User can enter all details related purely to  API backend such as name, description, base URL with Add API wizard 
  * [x] Add API wizard does not contain any details or settings related to proxy or selecting proxy
* [x] When user presses Publish API backend, the API will be added to API backends collection
  * [x] User get redirected to single API view for the newly created API

jykae:
## Proposed changes

- add Apis schema, closes #1091  
- add new add API form, replace wizard, closes #1379  
- update all related features to use Apis collection

Alternative step by step solution. Rename & clean in separate PR.

Known issues:
- Analytics not functional, it is Umbrella/Elasticsearch related package. Proposing this to be solved in separate task when new proxy schema is available and binding can be done.


brylie:
Clean up commit message 304912c. Also, consider merging these fixes into my working branch (PR #1397), since that is the designated branch for the Apis collection refactoring.

jykae:
@brylie I close this as duplicate effort

PR #1438, opened 2016-08-24
Meteor update to 1.4.1

jykae:
## Proposed changes

- upgrade Meteor platform to 1.4.1


jykae:
```
W20160824-16:21:00.278(3)? (STDERR) Note: you are using a pure-JavaScript implementation of bcrypt.
W20160824-16:21:00.553(3)? (STDERR) While this implementation will work correctly, it is known to be
W20160824-16:21:00.554(3)? (STDERR) approximately three times slower than the native implementation.
W20160824-16:21:00.554(3)? (STDERR) In order to use the native implementation instead, run
W20160824-16:21:00.555(3)? (STDERR)
W20160824-16:21:00.555(3)? (STDERR)   meteor npm install --save bcrypt
W20160824-16:21:00.556(3)? (STDERR)
W20160824-16:21:00.556(3)? (STDERR) in the root directory of your application.
```
Adds also native bcrypt as suggested after upgrade.

PR #1440, opened 2016-08-24
Add link to favicon

### LINKED ISSUE #1235: favicon
Console complains a favicon is missing. Should be a quick win. We can use a simple tool like http://realfavicongenerator.net/

jykae:
Closes #1235 

## Proposed changes

- add html head link to already existing favicon.png

### DIFF
--- a/core/client/shared/head.html
+++ b/core/client/shared/head.html
@@ -1,0 +1,3 @@
+<head>
+  <link rel="icon" sizes="16x16 32x32" href="/favicon.png?v=2">
+</head>


core/client/shared/head.html
```diff
@@ -0,0 +1,3 @@
+<head>
+  <link rel="icon" sizes="16x16 32x32" href="/favicon.png?v=2">
```
bajiat: What does the ?v=2 mean? There seems to be one favicon in public folder that hasn't changed in a year.

philippeluickx: If I remember well, this will force the browser to update the favicon in case there is a cached one. It's a good practice

PR #1480, opened 2016-09-02
Feature/swagger

### LINKED ISSUE #1476: Implement Swagger viewer as npm module
Swagger Viewer is currently displayed in iFrame. The solution is very brittle and for example inserting base URL and user's API for making calls through apinf would require a hacky implementation.

The iFrame solution should be replaced by implementing Swagger viewer as npm module.

# Definition of done
* [ ] Suitable, actively maintained npm package for Swagger UI has been selected
* [ ] Package has been integrated into Apinf
* [ ] Swagger viewer module is displayed in the same place as the current iFrame solution
* [ ] Swagger viewer is only displayed in documentation tab, if a Swagger file has been uploaded for that particular API
* [ ] Swagger viewer displays the Swagger file that has been uploaded for the particular API

# Candidate npm packages
- [swagger-ui](https://www.npmjs.com/package/swagger-ui)
- [swagger-ui-lite](https://www.npmjs.com/package/swagger-ui-lite)
- [Swagger UI Browserify](https://www.npmjs.com/package/swagger-ui-browserify)
- [Swagger UI Headless](https://www.npmjs.com/package/swagger-ui-headless)

## Example of integrating Swagger UI into templating languages
- [Ember Swagger UI](https://www.npmjs.com/package/ember-swagger-ui)

marla-singer:
Closes #1476


brylie:
Remember to add an ignore rule to our `.eslintignore` so that Swagger UI code is not checked.

brylie:
Also, try importing the Swagger UI CSS directly from the NPM package:

- https://guide.meteor.com/using-npm-packages.html#npm-styles

marla-singer:
@brylie thanks for link!
Whether it is necessary to display buttons "Authorize" and "Explore" or these buttons don't need in apinf project?

bajiat:
@marla-singer The buttons are needed. Although, I think the presence of the Authorize button depends on the Swagger file - it is only available if the authorization has been defined in the Swagger file to be displayed.

marla-singer:
@brylie @bajiat I come in to collision:
1) I use package "swagger-ui-browserify" that base on 2.1.4 version of "swagger-ui". It works, but Authorization button is not avaible because this feature was appeared in ^2.2 Swagger version. 
2) So I try to use package "swagger-ui" in new empty meteor project and have error on console
```
Uncaught TypeError:template is not a function
```
which i can't fix it quickly and can't find solution on the Enthernet.

marla-singer:
@brylie @bajiat Well, I did fork of swagger-ui-browserify (SUB) repo, updated swagger-ui to latest version and fixed errors. After I created new meteor project with my SUB package version and run it. You can look sources [here](https://github.com/marla-singer/empty-project/tree/implement-swagger).  Also I added some instruction to run project. 

Can I integrate this private SUB package into apinf project? Package is available [here ](https://github.com/marla-singer/swagger-ui-browserify/releases/tag/v2.2.3) and PR was created [too   ](https://github.com/achingbrain/swagger-ui-browserify/pull/4)

brylie:
@marla-singer good work. We basically need to be able to duplicate what you did in the 'empty-project', but within the Apinf project. E.g. we want to be able to call the following code from within Apinf:

```js
const swagger = window.swaggerUi = new SwaggerUi({
    url: '/swagger.json',
    dom_id: 'swagger-ui-container',
    useJQuery: true,
    supportHeaderParams: true,
    onComplete: function (swaggerApi, swaggerUi) {
      console.log('Loaded SwaggerUI')
    },
    onFailure: function (error) {
      console.error('Unable to Load SwaggerUI', error)
    },
    docExpansion: 'none'
  });
  
swagger.load()
```

Except, we do not want to attach `swaggerUi` to `window`. We would rather `import { SwaggerUi } from 'swagger-ui-browserify'`.

marla-singer:
@brylie Yes, I see that attach swaggerUi to window is a bad idea but Swagger developers use swaggerUi variable in library like global. It won't work if I create variable without `window.swaggerUi`. I need more time to explore this problem and find some solution.

marla-singer:
@brylie Official npm package "swagger-ui-browserify" bases on swagger-ui version 2.1.4 and doesn't provide an implementation of Authorization button

brylie:
@marla-singer your [pull request](https://github.com/achingbrain/swagger-ui-browserify/pull/4) was merged into *swagger-ui-browserify*, updating the package to *2.2.3*.

brylie:
> I see that attach swaggerUi to window is a bad idea but Swagger developers use swaggerUi variable in library like global. It won't work if I create variable without window.swaggerUi.

Do you have an upstream support request regarding this issue?

brylie:
@bajiat should we go ahead and use the [`swagger-ui-browserify`](https://www.npmjs.com/package/swagger-ui-browserify) package that @marla-singer has contributed to, with the [expectation that @marla-singer's changes will be published soon](https://github.com/achingbrain/swagger-ui-browserify/pull/4#issuecomment-245222726)?

This will mean that the Authorization button may not be in @marla-singer's PR, but should appear when the upstream package is updated.

bajiat:
@brylie @marla-singer It is ok to continue and use the swagger-ui-browserify particularly if they are planning to publish it in the next couple of days. If we need to upgrade the package, we can do that in connection with the Swagger UI continuation task(s) we have in backlog.

marla-singer:
@brylie Please review.
I used swagger-ui-browserify v2.2.3 where don't resolve problem with attach swagger to window. So I suggest to improve it after developer will publish patch.

brylie:
Cool. For now we can attach it to window. Can we remove the SwaggerUi object from window when the documentation browser is not visible?

brylie:
I went ahead and merged this PR. We can come back to the `window` cleanup or find an alternative approach as a follow-up task.

PR #1491, opened 2016-09-06
Refactor apiUmbrellaWeb usage

### LINKED ISSUE #1485: Refactor APIUmbrellaWeb object usage
Refactor createApiUmbrellaWeb function and apiUmbrellaSettingsValid functions.

createApiUmbrellaWeb returns API wrapper object. To be called when needed. Switching away from global object usage.

jykae:
Closes #1485 

## Proposed changes
- refactor createApiUmbrellaWeb to return wrapper object for Umbrella Admin API
- refactor apiUmbrellaSettingsValid function based on new field names in Proxies

Please review @apinf/developers

### DIFF
--- a/proxies/server/methods.js
+++ b/proxies/server/methods.js
@@ -24,7 +24,7 @@ Meteor.methods({
     }
   },
   'createApiUmbrellaWeb': function () {
-    // Assume one proxy only
+    // TODO: Assuming one proxy only
     const proxy = Proxies.findOne();
 
     // Check if API Umbrella Web settings are valid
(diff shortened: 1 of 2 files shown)


proxies/server/methods.js
```diff
@@ -23,34 +24,25 @@ Meteor.methods({
     }
   },
   'createApiUmbrellaWeb': function () {
-    const settings = Settings.findOne();
+    // Assume one proxy only
```
brylie: Add a TODO here mentioning that we need to refactor this method for multi-proxy support. That way we will remember what files to edit.

brylie:
@jykae this looks good. Please add the TODO comment, and we can merge the PR.

jykae:
@brylie TODO added

brylie:
I added a commit to clarify the comment. I.e. a TODO should be a verb, something that we have *to do*.

jykae:
Ok, thanks. Anything else or could we merge this so I could use it in apiKey PR ?

PR #1497, opened 2016-09-07
feature/api-settings-export-fixes

### LINKED ISSUE #1385: Change API settings export to match collection change
Make sure that both API details and proxy backend settings can be exported from Export tab. (If possible, tab should be available only for admin and API owner)

Note that API settings can already be exported.

# Definition of done
* [x] User has possibility to download both API backend details and proxy API backends configurations (can be separate files, in which case user needs to be able to select what they want)
* [x] Minimum requirement is that user can download json
* [x] If possible, Export tab is only available for admins and API managers

nnn:
Closes #1385


apis/client/export/export.html
```diff
@@ -6,13 +6,13 @@ <h2 class="panel-title">
       </h2>
     </div>
     <div class="panel-body">
-      <h3>Export API configuration</h3>
-      <p>Export API configuration</p>
+      <h3>{{_ "view_ApiBackend_Export_API_Title"}}</h3>
```
brylie: This i18n key should start with the template name as it appears in the `name` attribute:

```html
{{_ "viewApiBackendExport_api_title" }}
```

apis/client/export/export.html
```diff
@@ -6,13 +6,13 @@ <h2 class="panel-title">
...
+      <p>{{_ "view_ApiBackend_Export_API_Description"}}</p>
```
brylie: This i18n key should start with the template name as it appears in the `name` attribute:

```html
{{_ "viewApiBackendExport_api_description" }}
```

apis/client/export/export.html
```diff
@@ -6,13 +6,13 @@ <h2 class="panel-title">
...
       <a class="btn btn-success" id="exportJSONConfig">{{_ "view_ApiBackend_ExportJson"}}</a>
       <a class="btn btn-success" id="exportYAMLConfig">{{_ "view_ApiBackend_ExportYaml"}}</a>
       <hr class="panel-hr" />
-      <h3>Export proxy configuration</h3>
-      <p>Export proxy configuration</p>
+      <h3>{{_ "view_ApiBackend_Export_Proxy_Title"}}</h3>
```
brylie: This i18n key should start with the template name as it appears in the `name` attribute:

```html
{{_ "viewApiBackendExport_proxy_title" }}
```

apis/client/export/export.html
```diff
@@ -6,13 +6,13 @@ <h2 class="panel-title">
...
+      <p>{{_ "view_ApiBackend_Export_Proxy_Description"}}</p>
```
brylie: This i18n key should start with the template name as it appears in the `name` attribute:

```html
{{_ "viewApiBackendExport_proxy_description" }}
```

apis/client/export/export.js
```diff
@@ -1,5 +1,19 @@
...
+import { ProxyBackends } from '/proxy_backends/collection';
+
 import jsyaml from 'js-yaml';
 
+Template.viewApiBackendExport.onCreated(function () {
+  // Get reference to template instance
+  const instance = this;
+
+  // Get the API Backend ID from the route
+  instance.apiId = Router.current().params._id;
```
brylie: Get the API ID from the template data context, rather than the router. This is for two reasons:

- the API template already passes the `api` into child templates
  - `const apiId = instance.data.api._id` should be easy to get from the template data context
- we will have less router code to convert when porting to flow:router

apis/client/export/export.html
```diff
@@ -2,19 +2,19 @@
...
       </h2>
     </div>
     <div class="panel-body">
-      <h3>{{_ "view_ApiBackend_Export_API_Title"}}</h3>
-      <p>{{_ "view_ApiBackend_Export_API_Description"}}</p>
-      <a class="btn btn-success" id="exportJSONConfig">{{_ "view_ApiBackend_ExportJson"}}</a>
-      <a class="btn btn-success" id="exportYAMLConfig">{{_ "view_ApiBackend_ExportYaml"}}</a>
+      <h3>{{_ "view_ApiBackendExport_api_title"}}</h3>
+      <p>{{_ "view_ApiBackendExport_api_description"}}</p>
+      <a class="btn btn-success" id="exportJSONConfig">{{_ "view_ApiBackendExport_json_button"}}</a>
```
brylie: The template name is `viewApiBackendExport`. The general pattern for i18n tokens is:

> "templateName_pageElement_additionalText"

So, this token would be:

`viewApiBackendExport_api_title`

apis/client/export/export.html
```diff
@@ -2,19 +2,19 @@
...
+        {{_ "view_ApiBackendExport_title"}}
       </h2>
     </div>
     <div class="panel-body">
-      <h3>{{_ "view_ApiBackend_Export_API_Title"}}</h3>
-      <p>{{_ "view_ApiBackend_Export_API_Description"}}</p>
-      <a class="btn btn-success" id="exportJSONConfig">{{_ "view_ApiBackend_ExportJson"}}</a>
-      <a class="btn btn-success" id="exportYAMLConfig">{{_ "view_ApiBackend_ExportYaml"}}</a>
+      <h3>{{_ "view_ApiBackendExport_api_title"}}</h3>
+      <p>{{_ "view_ApiBackendExport_api_description"}}</p>
```
brylie: The template name is `viewApiBackendExport`. The general pattern for i18n tokens is:

> "templateName_pageElement_additionalText"

So, this token would be:

`viewApiBackendExport_api_description`

apis/client/export/export.html
```diff
@@ -2,19 +2,19 @@
...
-        {{_ "view_ApiBackend_Export_Title"}}
+        {{_ "view_ApiBackendExport_title"}}
       </h2>
     </div>
     <div class="panel-body">
-      <h3>{{_ "view_ApiBackend_Export_API_Title"}}</h3>
-      <p>{{_ "view_ApiBackend_Export_API_Description"}}</p>
-      <a class="btn btn-success" id="exportJSONConfig">{{_ "view_ApiBackend_ExportJson"}}</a>
-      <a class="btn btn-success" id="exportYAMLConfig">{{_ "view_ApiBackend_ExportYaml"}}</a>
+      <h3>{{_ "view_ApiBackendExport_api_title"}}</h3>
```
brylie: The template name is `viewApiBackendExport`. The general pattern for i18n tokens is:

> "templateName_pageElement_additionalText"

So, this token would be:

`viewApiBackendExport_api_title`

brylie:
This is almost ready. There are still some issues with the i18n keys, so please observe how other i18n keys are structured.

Also, make use of nesting to improve readability. Add a newline after HTML tags, so that child elements are indented one level.

nnn:
@brylie added suggested changes.

apis/client/view/header/header.html
```diff
@@ -72,14 +72,15 @@ <h1 class="api-name">
               {{_ "viewApiBackendNavigationMenu_backlog" }}
             </a>
           </li>
+          {{ # if currentUser }}
+          {{ # if api.currentUserCanEdit }}
           <li>
-            <a href="#api-export" data-toggle="tab">
+            <a href="#api-backend-export" data-toggle="tab">
```
brylie: Small detail, but the word 'backend' was removed on purpose, please revert -- and if possible, remove Backend from the template name too. 

Basically, the word 'backend' is appearing in our code hundreds of times, and is not really helping improve readability. We have been removing 'backend' and 'frontend' from our user interface and code, to the extent possible.

I can remove the 'backend' from `viewApiBackend` and all related i18n strings, if you will just clean up this part of the code.

brylie:
This is *really* close.

One thing I have been trying to do is remove 'backend' from our code and user interface. It was used in a confusing manner for both end users and developers. We are switching to the following terminology for simplicity:

- API - an API somewhere on the web, added to our catalogue
- Proxy - an API Proxy solution, such as API Umbrella
- ProxyBackend - an API that has been configured behind a proxy, hence 'backend'

We are only using 'backend' related to the third case, in order to distinguish it from a regular API, and to make it easier to describe our software.

brylie:
I can make the above change, but it is important that we all know to limit the word 'backend'.

nnn:
@brylie, got it, thanks for the explanation.

brylie:
@nnn I made the changes, renaming *API* related templates for consistency. Please review.

brylie:
@nnn please review.

Issue #1510, opened 2016-09-08
Edit Proxy backend and store the changes in API Umbrella


bajiat:
If user edits the Proxy backend settings in te Proxies tab in API profile view, the changes will need to be reflected in API Umbrella.

# Definition of done
* [x] The edited proxy backend data/settings gets properly updated in API Umbrella

bajiat:
@nnn Would you be interested in this task? May require some pair programming.

bajiat:
@nnn What is the status of this task?

nnn:
@bajiat started working on it, have a PR opened. Planning on finishing it during Tuesday.

PR #1518, opened 2016-09-09
Feature/disable apikey

### LINKED ISSUE #622: Ensure api managers can disable API key check on Backends

nnn:
Closes #622


nnn:
Blocked by #1520

bajiat:
Status update? Can you start by adding UI and only then store to UI?

nnn:
@bajiat, yes, the problem is that I couldn't submit Proxy form. Trying it again today.
UI is in a place.

nnn:
Everything is in place. Since Proxy settings can be saved to API Umbrella, can't test it. 
Waiting for #1510 to be merged.

proxy_backends/collection/apiUmbrellaSchema.js
```diff
@@ -1,6 +1,16 @@
 // Utility import
 import { proxyBasePathRegEx, apiBasePathRegEx } from './regex';
 
+
+const Settings = new SimpleSchema({
+  'disable_api_key': {
+    type: Boolean,
+    optional: true,
+    label: 'Disable API Key',
```
brylie: We need to internationalize this label.

brylie:
Add internationalization.

brylie:
This is really close. Just switch identifiers to ID rather than class, when there is only one 'instance' or a particular element on a page.

PR #1524, opened 2016-09-09
Enhance API key feature 

### LINKED ISSUE #1483: Enhance Get API key
Enhancement to #1384.

Get API key and Copy API key buttons should only be available in the API profile page, if the API is connected to a proxy.

# Definition of done
* [x] Get API key is only available for APIs that are connected to a proxy.
* [x] Copy API key is only available for APIs that are connected to a proxy.
* [x] When clicking 'Get API Key' button, indicate that something is happening

Note that Get API key is shown before user has pressed that key and has received an API key. After that, Copy API key is shown to that user for all APIs connected to the proxy.

jykae:
Closes #1483 

- add processing state for usability, reason: Umbrella Admin API call to create user takes some time, avoid multiple clicks from user and provide UI feedback


jykae:
Small improvement for previous sprint work. @apinf/developers please review

api_keys/client/api_key.js
```diff
@@ -6,6 +6,10 @@ Template.apikey.onCreated(function() {
 
 Template.apikey.events({
   'click #getApiKeyButton' (event) {
+    // Set bootstrap loadingText
+    $('#getApiKeyButton').button({loadingText: TAPi18n.__('apiKeys_getApiKeyButton_processing')});
+    // Set button to processing state
```
brylie: **Lint:** put line breaks between comments and the preceding line.

apis/client/view/view.html
```diff
@@ -1,6 +1,6 @@
 <template name="viewApiBackend">
   {{#if Template.subscriptionsReady }}
-  {{> viewApiBackendPageHeader api=api proxyIsConfigured=proxyIsConfigured }}
+  {{> viewApiBackendPageHeader api=api proxyIsConfigured=proxyIsConfigured proxyBackend=proxyBackend }}
```
brylie: Remove the `proxyIsConfigured` helper, as previously discussed, since it is redundant. I.e. `proxyBackend` wouldn't exist of the proxy wasn't configured.

brylie:
This looks good, and we can merge it today. Please lint the code and remove the redundant template helper `proxyIsConfigured`.

brylie:
@jykae also, please lint your code. This is easily done by enabling `linter-eslint` in Atom IDE. If you use a different IDE, see the relevant [eslint installation instructions](http://eslint.org/docs/user-guide/integrations).

apis/client/view/view.html
jykae: @Are you sure? You are showing the "Proxy" tab based on that. I think proxy settings can exist before proxy configuration has been done for API.

jykae:
Btw, what is purpose of having Linthub checks that usually shows green if we need to do linting ourselves anyway? I am getting quite a lot errors and warnings using linter-eslint

jykae:
I am not going to get away all those linting errors that eslint is nagging. This is way too strict for our code :)

brylie:
@jykae which errors are you having trouble with?

brylie:
Some of the linting is done automatically when you save the file in Atom.

brylie:
In general, we all need to lint code while we develop. 

I am trying to figure out why Linthub does not seem to detect our ESLint rules.

api_keys/client/api_key.js
```diff
@@ -44,14 +52,13 @@ Template.apikey.helpers({
...
       // Get API Key
-      const apiKey = ApiKeys.findOne({'userId': currentUserId});
+      const apiKey = ApiKeys.findOne({ userId: currentUserId });
 
       // Check that Umbrella API key exists
-      if(apiKey && apiKey.apiUmbrella) {
+      if (apiKey && apiKey.apiUmbrella) {
         // Shorten key to be shown in UI
-        return apiKey.apiUmbrella.apiKey.substring(0,15) + "...";
+        return apiKey.apiUmbrella.apiKey.substring(0, 15) + '...';
```
brylie: Use CSS to truncate the API key, rather than JavaScript. That way we don't have to maintain this `apiKeyShort` function.

apis/client/view/view.html
brylie: OK, I forgot the purpose of the `proxyIsConfigured` helper, it seems.

apis/client/view/header/header.html
```diff
@@ -34,9 +34,11 @@ <h1 class="api-name">
         </div>
         {{> visibilitySwitch api=api }}
       </div>
+      {{#if proxyBackend }}
       <div class="row">
```
brylie: Move the {{> apiKey }} template into the row above, rather than creating a new row for this one page element. It uses unnecessary vertical space at the top of the page.

brylie:
@jykae I removed the unused `apiKeyShort` template helper and linted the code slightly. Please review.

jykae:
@brylie Ok to merge.

PR #1527, opened 2016-09-09
Hide status indicator when URL not available

### LINKED ISSUE #1526: API Status indicator throwing errors related to URL
When trying to render API Backends that have been synced from API Umbrella, the status indicator is looking for a missing URL property. This property was introduced in a recent pull request, and is not present on APIs from API Umbrella. As a stop gap, write a check for this URL property, and hide the status indicator for APIs without the url property.

brylie:
Closes #1526 

# Changes
- check api object for `url`property
- hide status indicator for APIs without `url` property


jykae:
Good check.

PR #1537, opened 2016-09-11
Update docker image for api-umbrella

### LINKED ISSUE #1249: Upgrade API Umbrella to 0.13
API Umbrella 0.13 was recently released. Upgrade our nightly, and possibly production, to use the latest version.

shaliko:
* Update docker image for api-umbrella to version 0.13.0
* Split ElasticSearch fro api-umbrella to separate container.
* Make ElasticSearch available only from private docker network (between containers). 

Closes #1249

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,8 +11,11 @@ services:
       - "3000:80"
     depends_on:
       - mongo
+      - ssl
+    links:
+      - elasticsearch:elasticsearch.docker
   apiumbrella:
-    image: nrel/api-umbrella:0.12.0
+    image: nrel/api-umbrella:0.13.0
     restart: always
     volumes:
       - ./docker/api-umbrella/config:/etc/api-umbrella
@@ -20,9 +23,10 @@ services:
     ports:
       - "3001:3001"
       - "3002:3002"
-      - "14002:14002"
     depends_on:
       - ssl
+    links:
+      - elasticsearch:elasticsearch.docker
   mongo:
     restart: always
     image: mongo:3.2.6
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -41,9 +45,13 @@ services:
       - letsencrypt:/etc/letsencrypt
       - letsencrypt_backups:/var/lib/letsencrypt
       - dhparam_cache:/cache
-    depends_on:
-      - apinf
+  elasticsearch:
+    image: elasticsearch:1.7.5
```
brylie: Should we also be depending on Elasticsearch 2.x.x?

shaliko: ElasticSearch 2 allowed, but by default api-umbrella use 1.7.5. Because we will migrate on use API instead of direct call to Elasticsearch, I decide keep default version.

But if you see any scenario where it will be useful I can migrate on 2.x.x

brylie:
Well, lets go ahead and migrate to Elastic 2 now. It changes our analytics schema slightly, so we can make the changes sooner, rather than later.

shaliko:
@brylie After migrate to Elastic 2, api-umbrella dashboard works OK (Analytics), but apinf-umbrella-dashboard dose not show data. Tomorrow will check with @marla-singer

brylie:
OK, thanks @shaliko. We may need to make an update to our dashboard code, to match the Elasticsearch data structure.

brylie:
Please add the **Ready for revies** tag once you have spoken with @marla-singer

shaliko:
@brylie In debug session with @marla-singer we found the root case of issue:

```
"reason": "Result window is too large, from + size must be less than or equal to: [10000] but was [50000]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level parameter."
```

For work with old releases (docker images) I will add increase "max_result_window" on start ElasticSearch. But I am not sure that we really need [50000 value](https://github.com/apinf/platform/blob/56b538e3c999618708f53844c5c51b25d70ec6a9/dashboard/client/dashboard.js#L63).

PR #1555, opened 2016-09-12
Feature/remove both lib server folders

brylie:
Closes #936 
Closes #1556 
Closes #1557
Closes #1558
Closes #1559 
Closes #1575 
Closes #1577
Closes #1578
Closes #1579

# Changes
- Make final changes to modularize project package structure
  - move files from `/both` folder into relevant locations
  - move files from `/lib` folder into relevant locations
  - move files from `/server` folder into relevant locations
- Add missing i18n tokens where found in cleanup process above
- Adjust import/exports
- Deprecate FlashMessages package
- Add missing sAlert i18n functionality
- Automatic lint
- remove `editApi` component
- remove `manageApis` component

PR #1563, opened 2016-09-12
Try-out button for Swagger file

### LINKED ISSUE #1490: API owner to be able to control 'Try it out!' button
In swagger documentation it's possible t try to make real calls. Some customers want to allow try out for all methods, some for only get methods, some customers don't want it at all.

Controlling 'Try it out!' can be done when Swagger Ui is launched in Javascript with supportedSubmitMethods option. With supportedSubmitMethods option it's possible to 

To serve customer, should give customer an option to choose for each api documentation, which are the methods that support 'Try it out!'.

     window.swaggerUi = new SwaggerUi({
        url: url,
        dom_id: "swagger-ui-container",
        supportedSubmitMethods: ['get', 'post', 'put', 'delete'],

# Wireframe

joxi_screenshot_1473677587247

marla-singer:
Closes #1490 
- Add checkboxes on the Manage Documentation modal
- Add autorun to render Swagger Ui after changing methods

### DIFF
--- a/apis/collection/schema.js
+++ b/apis/collection/schema.js
@@ -25,6 +25,7 @@ Apis.schema = new SimpleSchema({
     type: String,
     optional: true
   },
+  
   apiLogoFileId: {
     type: String,
     optional: true
@@ -33,6 +34,11 @@ Apis.schema = new SimpleSchema({
     type: String,
     optional: true,
     regEx: SimpleSchema.RegEx.Url
+  },
+  submit_methods: {
+    type: [String],
+    label: 'Allow try-out for following methods:',
+    optional: true
   },
   created_at: {
     type: Date,
--- a/apis/collection/schema.js
+++ b/apis/collection/schema.js
@@ -37,7 +37,6 @@ Apis.schema = new SimpleSchema({
   },
   submit_methods: {
     type: [String],
-    label: 'Allow try-out for following methods:',
     optional: true
   },
   created_at: {
@@ -93,5 +92,8 @@ Apis.schema = new SimpleSchema({
     defaultValue: true
   }
 });
+
+// Enable translations (i18n)
+Apis.schema.i18n('schemas.apis');
 
 Apis.attachSchema(Apis.schema);
--- a/documentation/client/manage/autoform.js
+++ b/documentation/client/manage/autoform.js
@@ -3,7 +3,7 @@ AutoForm.hooks({
     onSuccess () {
       // Get success message translation
       const message = TAPi18n.__('manageApiDocumentationModal_LinkField_Updated_Message');
-
+      
       // Alert user of success
       sAlert.success(message);
     },
--- a/documentation/client/manage/autoform.js
+++ b/documentation/client/manage/autoform.js
@@ -1,7 +1,11 @@
 AutoForm.hooks({
   apiDocumentationForm: {
     onSuccess () {
-      sAlert.success(TAPi18n.__('manageApiDocumentationModal_LinkField_Updated_Message'));
+      // Get success message translation
+      const message = TAPi18n.__('manageApiDocumentationModal_LinkField_Updated_Message');
+
+      // Alert user of success
+      sAlert.success(message);
     },
   }
 });
--- a/documentation/client/manage/autoform.js
+++ b/documentation/client/manage/autoform.js
@@ -1,5 +1,7 @@
-AutoForm.addHooks(['apiDocumentationForm'], {
-  onSuccess () {
-    sAlert.success(TAPi18n.__('manageApiDocumentationModal_LinkField_Updated_Message'));
-  },
+AutoForm.hooks({
+  apiDocumentationForm: {
+    onSuccess () {
+      sAlert.success(TAPi18n.__('manageApiDocumentationModal_LinkField_Updated_Message'));
+    },
+  }
 });
--- a/documentation/client/manage/manage.html
+++ b/documentation/client/manage/manage.html
@@ -11,25 +11,25 @@
         <div class="modal-body">
           {{#autoForm collection=apisCollection doc=api id="apiDocumentationForm" type="update" }}
 
-              <legend>
+            <legend>
               {{> showHelp 'uploadApiDocumentation'}}
               {{_ "import_ApiDocumentation_Title"}}
-              </legend>
+            </legend>
             <div class="documentation-file">
               {{#unless documentationFile }}
-              {{> manageApiDocumentationModalUploadButton }}
+                {{> manageApiDocumentationModalUploadButton }}
               {{else}}
-              <ul class="list-group">
-                <li class="list-group-item">
-                  <i class="fa fa-file-text-o" aria-hidden="true"></i>
-                  {{ documentationFile.filename }}
-                  <span class="pull-right">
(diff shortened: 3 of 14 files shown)


marla-singer:
@preriasusi Could you please review?

brylie:
This looks good, but is behaving very strangely.

E.g. After uploading a Swagger document, the documentation viewer renders, but none of the sub sections will open when clicking **show/hide** or the endpoint name.

I notice the following console errors:

```
modules.js:41892 Uncaught TypeError: $.param.fragment is not a function
modules.js:41929 Uncaught TypeError: Cannot read property 'pushState' of undefined
```

brylie:
I am also noticing multiple sAlerts when clicking the 'choose file' button:

> Documentation link successfully updated!

Check the event handlers for the template to see if they are overlapping.

apis/collection/schema.js
```diff
@@ -34,6 +35,11 @@ Apis.schema = new SimpleSchema({
     optional: true,
     regEx: SimpleSchema.RegEx.Url
   },
+  submit_methods: {
+    type: [String],
+    label: 'Allow try-out for following methods:',
```
brylie: We need to internationalize this label. It can be done via schema i18n. See, for example:

- https://github.com/apinf/platform/blob/develop/metadata/collection/schema.js#L63-L64
- https://github.com/apinf/platform/blob/develop/lib/i18n/en.i18n.json#L297-L298

marla-singer:
>modules.js:41892 Uncaught TypeError: $.param.fragment is not a function

I suppose this error shows when npm packages installed/ran in the wrong order. If you delete folder 'node_modules' and reinstall packages it should work correctly

marla-singer:
@brylie Done. I also update swagger-ui-browseify package version and I don't use window variable. But I  will push it after merging #1555

brylie:
@marla-singer is this ready for review? Please remember to toggle the 'ready for review' label, to make it clear when a PR is ready.

I will go ahead and review, but wanted to make sure the PR is ready.

marla-singer:
@brylie Yes, PR is ready

documentation/client/manage/autoform.js
```diff
@@ -1,5 +1,7 @@
-AutoForm.addHooks(['apiDocumentationForm'], {
-  onSuccess () {
-    sAlert.success(TAPi18n.__('manageApiDocumentationModal_LinkField_Updated_Message'));
-  },
-});
+AutoForm.hooks({
+  apiDocumentationForm: {
+    onSuccess () {
+      sAlert.success(TAPi18n.__('manageApiDocumentationModal_LinkField_Updated_Message'));
```
brylie: Please separate this line into two parts:

```js
// Get success message translation
const message = TAPi18n.__('manageApiDocumentationModal_LinkField_Updated_Message');

// Alert user of success
sAlert.success(message);
```

It is a bit 'cleaner' this way, and shortens the line length.

documentation/client/manage/manage.html
```diff
@@ -9,48 +9,56 @@
...
+              {{> showHelp 'uploadApiDocumentation'}}
+              {{_ "import_ApiDocumentation_Title"}}
+              </legend>
+            <div class="documentation-file">
+              {{#unless documentationFile }}
+              {{> manageApiDocumentationModalUploadButton }}
+              {{else}}
+              <ul class="list-group">
+                <li class="list-group-item">
+                  <span class="glyphicon glyphicon-file"></span>
```
brylie: Consider using FontAwesome icon here, since we have been using FontAwesome more often than Glyphicon. If you like the Glyphicon better, it can stay. Just check FontAwsome in case it has a nicer icon.

http://fontawesome.io/icons/

marla-singer: It's old code, I didn't use glyphicon. Replace?

documentation/client/manage/manage.html
```diff
@@ -9,48 +9,56 @@
...
+                  {{ documentationFile.filename }}
+                  <span class="pull-right">
+                    <button type="button" class="btn btn-xs btn-danger delete-documentation">
+                      <span class="glyphicon glyphicon-trash"></span>
```
brylie: Compare Glyphicon with available FontAwesome fonts. If Glyphicon is 'better' keep it. If FontAwesome is nicer, use FA here instead.

documentation/client/manage/autoform.js
marla-singer: Okay

documentation/client/manage/manage.js
```diff
@@ -110,4 +114,13 @@ Template.manageApiDocumentationModal.helpers({
     // Return a reference to Apis collection, for AutoForm
     return Apis;
   },
+  submitMethods () {
+    return [
```
brylie: Add comment, and/or semantic variable name, describing this return object. E.g. what is it used for?

documentation/client/manage/uploadButton/uploadButton.html
```diff
@@ -1,3 +1,3 @@
 <template name="manageApiDocumentationModalUploadButton">
-  <button class="btn btn-primary fileBrowse">{{_ "manageApiDocumentationModalUploadButton" }}</button>
+  <button type="button" class="btn btn-primary fileBrowse">{{_ "manageApiDocumentationModalUploadButton" }}</button>
```
brylie: Break this into multiple lines, to improve readability.

```html
<button type="button" class="btn btn-primary fileBrowse">
  {{_ "manageApiDocumentationModalUploadButton" }}
</button>
```

Also, consider using an ID, instead of class, for the `fileBrowse` label.

This reason is twofold:
1. ID is used for unique page elements
  - there is only one `fileBrowse` button on the page, so the `fileBrowse` label should be ID
2. moving `fileBrowse` to an `id` attribute makes the code a little easier to read:
```html
<button type="button" id="file-browse" class="btn btn-primary">
  ...
</button>
```

Notice also that we are trying to use hyphens instead of camel case for HTML attributes. This arbitrary decision is a bit more inline with other class declarations, e.g. Bootstrap classes.

documentation/client/swaggerUI/swaggerUi.html
```diff
@@ -16,7 +16,7 @@
         <div id="message-bar" class="swagger-ui-wrap message-success" data-sw-translate=""></div>
         {{#if dataFetched}}
             {{#if documentationValid}}
-                {{>swaggerUiContent apiDocumentation=apiDocumentation}}
+                {{>swaggerUiContent apiDoc=apiDoc api=api}}
```
brylie: Check the indentation on this file. It may be sufficient to use two space per nested level when indenting HTML files. This also matches our JS indentation convention.

brylie:
I have done a review of the code. There are some 'lint' issues, but overall this is ready to merge soon.

documentation/client/manage/manage.html
brylie: It is up to you. Which do you like better?

One thing to note, we can clean up/improve/lint old code when working nearby or refactoring related parts.

marla-singer: I'd like to use FontAwesome

marla-singer:
@brylie Done. Thanks for you recommendations!

PR #1574, opened 2016-09-14
username validation errors

### LINKED ISSUE #1507: Unable to edit 'Username' in 'Profile' view
Resolution: Make the username field read-only.

Steps:  
1.Visit https://nightly2.apinf.io  
2.Create account by clicking on 'Sign up'  
3.Navigate to 'Profile'  view  
4.Edit username  
5.Click on 'Update' button   

Result: 
- Highlights 'Username' field with error message 'profile-usernameIsTaken'   
- Enter the previously given username and click on 'update' also shows the same error message.  
- Signout->Relogin->Go to profile view->shows the same error message.  

Browser:   
1. Mozilla Firefox 48.0  
(Mozilla Firefox for Ubuntu Canonical – 1.0 )    
2. Google Chrome 52.0.2743.116    

Operating System:  
Ubuntu 16.0 LTS    

Screenshot:  
edit_username_error  

selection_022

nnn:
Closes #1507


brylie:
This looks good, and is a simple fix.

However, we realized there may be times when users don't have username, such as when registering with Github. We need to figure out what to do if this is still possible, so that all users have a username.

nnn:
Fixed. Ready for review.
Now it's possible to change username with RegEx validation and uniqueness.
If user is registering with Github it takes Github's username.

jykae:
Reviewing

brylie:
Make sure we handle the situation where a person might register using their Github account and there is already an Apinf user with that username. I think we already have code to cover that situation, but double check.

jykae:
I am having issue with Github settings & registration. It is not basically related to this PR, but makes it hard to test. Github settings did not seem to effect when I changed them on Settings, I had them first wrong (using nightly keys). After that did meteor reset I did the settings for Github and signup page keeps saying that "Configure Github". I will make issue about this.

jykae:
Opened issue https://github.com/apinf/platform/issues/1591

brylie:
Can we merge this pull request, or is there some work to be done?

brylie:
@nnn why did we change the username RegEx in this pull request? What was the source from which the current regular expression comes? Why is the current regex so complicated?

## Before
```js
const UsernameRegEx = /^[a-z0-9A-Z_\-]{3,15}$/;
```

## After
```js
const UsernameRegEx = /^(?!\d)(?!.*-.*-)(?!.*-$)(?!-)[a-zA-Z0-9-]{3,15}$/;
```

/cc @bajiat @jykae

PR #1593, opened 2016-09-15
Bugfix/enable GitHub settings

### LINKED ISSUE #1591: Github settings are not enabled when editing Settings
This function https://github.com/apinf/platform/blob/develop/settings/server/github_configuration.js#L4 does not get called when Settings are saved.

I think it should be called in autoform hook for Settings form https://github.com/apinf/platform/blob/develop/settings/client/autoform.js

This leads to behaviour where Sign Up form always shows "Configure Github".

jykae:
Closes #1591 

## Proposed changes
- make Github settings function Meteor method
- call method on settings form onSuccess hook to update Github settings


jykae:
@apinf/developers please review, fixes regression

One small cave-eat, that has been probably earlier also. 
If Github setting fields are saved blank, ServiceConfiguration does not get removed. But that could be enhancement if we make it possible to enable/disable also this Github login.

PR #1597, opened 2016-09-16
Feature/edit proxy form fix

brylie:
Closes #1532 
Closes #1596 

# Changes
- clean code
  - unneeded helpers
  - added comments
  - unneeded data context
- add tmeasday:publish-counts
- add `proxiesCount` publication
- use `proxiesCount` publication to control item visibility
  - navigation menu `proxyIsConfigured` helper
  - API view `proxyIsConfigured` helper

PR #1604, opened 2016-09-20
Disable proxy button when proxy is added

### LINKED ISSUE #1484: Disable Add proxy button if one proxy exists
While we are currently only supporting one proxy (one API Umbrella per deployment), disable the Add proxy button from Settings / Proxies if administrator has added one proxy to the deployment.

# Definition of done
* [x] Admin users can see the Add proxy button only if no proxy has been added to the platform
* [x] Collection allow/deny rule has been added
  * [x] `insert` is only allowed when no Proxies exist

frenchbread:
Closes #1484


frenchbread:
TODO: add help text to disabled button

bajiat:
@frenchbread Why not hide the button instead of disabling?

frenchbread:
@bajiat IMHO Hiding the "Add proxy" can confuse users. I have no idea where, in that case, to put a message saying e.g. "You have already added one proxy. Can't add more for now."

Disabling the button was also suggested by @Nazarah [here](https://github.com/apinf/platform/issues/1484#issuecomment-246261613).

What's left is attaching a tooltip with helper text to a button. Open to suggestions.

proxies/client/proxies.js
```diff
@@ -20,12 +20,15 @@ Template.proxies.onCreated(function(){
...
+  },
 });
 
 Template.proxies.helpers({
   proxies () {
     const instance = Template.instance();
     return instance.proxies.get();
-  }
+  },
+  biggerThan (a, b) {
```
nnn: Please use more meaningful name variables here, didn't quite get what is a and what is b.
Also comment will help.

core/lib/i18n/en.i18n.json
```diff
@@ -273,6 +273,7 @@
   "proxies_pageHeader": "Proxies",
   "proxies_addProxy": "Add proxy",
   "proxies_noProxiesFound": "No proxies found.",
+  "proxies_cannotAddMoreProxies": "You already have a Proxy available to attach your API. Please select the Proxy. You can add a new Proxy only after your have deleted the existing one.",
```
brylie: This help text is unnecessary, if we simply **hide the Add Proxy button**.

proxies/client/proxies.js
```diff
@@ -2,30 +2,41 @@ import { Proxies } from '../collection';
...
+
+      // Set button disabled if at least one proxy is already added
+      instance.isDisabled.set(proxiesCount >= 1);
     }
   });
 });
 
 Template.proxies.events({
   'click #add-proxy': function (event, instance) {
     Modal.show('addProxy', { proxy: {}, isEdit: false });
```
brylie: Be aware of changes to the `addProxy` form. Specifically, the template has been renamed `proxyForm` and used for both adding and editing proxies.

https://github.com/apinf/platform/pull/1597

proxies/collection/permissions.js
```diff
@@ -2,6 +2,11 @@ import { Proxies } from './';
 
 Proxies.allow({
   insert (userId, proxy) {
+    // Check if proxy is already exist
+    if (Proxies.find().count() >= 1) {
+      return false;
+    }
+
     // Check if user has admin role
     return Roles.userIsInRole(userId, ['admin']);
```
brylie: Add import for  `Roles` package (lint).

brylie:
This is really close. Just clean up the help text, which is not necessary. Also, check linter errors for missing imports, etc.

frenchbread:
Changes are done

PR #1617, opened 2016-09-23
Fix help text on Documentation tab

### LINKED ISSUE #1611: Documentation viewer help text i18n token mismatch
The documentation viewer help text is not rendering properly. 
screenshot from 2016-09-22 12-14-35

There is probably a mismatch with the i18n key, from the language files. Fix the documentation viewer help text i18n keys/translation strings.

marla-singer:
Closes #1611


marla-singer:
@brylie Could you please review?

PR #1638, opened 2016-09-26
Feature/only admins can add apis option

### LINKED ISSUE #1616: Allow only administrators to add APIs to a deployment
The point to limit that anyone can't add any kind of API can be in other way also: To limit only Admin to be able to add API. 

# Definition of done
* [x] There is a setting where admins can select that only admins can add APIs in this deployment.
* [x] When admin has selected this admin only setting
  * [x] The link to Add API is not visible to other type of users than admins.
  * [x] Other users than admins don't have insert privileges on the APIs collection.

# Wireframe

addapiadminprivilege

One more suggestion if there are future needs to implement more privileges based on user roles.
There can be a separate module named "User Role and Privileges"
This view would contain access privileges to different APINF users (Admin, API Owner, API Consumer) as radio buttons.
N - non accessible
R - Only view mode
W - View and Edit

Once a new privilege is introduced, this can be added in all three role panes.

addapiadminprivilege2

frenchbread:
Closes #1616 

## Changes:
- Added UI panel on `/settings` page
- Added schema field
- Permission checks when adding an API


frenchbread:

jykae:
Reviewing

jykae:
@frenchbread Works. i18n on place & linted 👍 Merging

frenchbread:
@jykae Sweet!

apis/collection/helpers.js
```diff
@@ -48,18 +49,17 @@ Apis.helpers({
     const managerIds = this.managerIds;
 
     // Create API managers array with usernames
-    const apiManagers = _.map(managerIds, function (id) {
+    const apiManagers = _.map(managerIds, (id) => {
       let userById;
       if (id) {
         userById = Meteor.users.findOne(id);
```
brylie: This code may not work on the client side, since there is likely no publication sending all user details. It may be related to other bug(s) currently being investigated by @marla-singer. All we need to check here is that the userId is in the managerIds array. Fetching the user object and username may be too complicated.

Issue #1618, opened 2016-09-23
Include star rating widget in API profile and Catalog


bajiat:
Reintroduce star rating into platform. Needs to be visible and editable in API catalog and API profile.

# Definition of done
* [x] Bookmarking icon and terminology needs to be distinct from star rating
  * [x] Bookmarking is called bookmark, icon can be freely chosen
  * [x] Rating uses star metaphor in icons, terminology refers to rating
* [x] Star rating is visible in Catalog (both card and table view)
* [x] Star rating is visible in API profile
* [ ] User can add and edit their rating 
  * [ ] in Catalog  (both card and table view)
  * [x] in API profile

# Wireframes

1. API Card View

![apicardnewbookmark-1](https://cloud.githubusercontent.com/assets/17976243/19039600/56f29ba2-8989-11e6-9f78-1f1d78f32f17.png)

2. API Table View

![apitablenew](https://cloud.githubusercontent.com/assets/17976243/19039623/683fedec-8989-11e6-92bc-24c51253f477.png)

3. API Profile

![apiprofilestar bookmark](https://cloud.githubusercontent.com/assets/17976243/19039667/8f3b248e-8989-11e6-99b5-0d1e89c6aca9.png)

bajiat:
@Nazarah: Wireframe needed refers to bookmark icon and placement of bookmark and star rating in catalog and API profile

bajiat:
@nnn Would you be interested in this task?

nnn:
@Nazarah when will you have time for this wireframe?
We can work together on ideas.

nnn:
Also #1625 is related to this task, and kinda depends.

Nazarah:
@nnn lets work together.

PR #1649, opened 2016-09-27
Feature/apiprofile header improvements

### LINKED ISSUE #1625: Organize API Profile header
I notice that we are adding new components to API Profile header and it starts to look clumsy.
### Proposition:

- [ ] Organize existing components, if needed move some components out from header
- [ ] Plan (ex. wireframe) position of future components based on backlog

nnn:
Closes #1625 

### Proposed changes
- Moved tour button
- Moved private/public button
- Cleaned up


nnn:
Partly blocked by #1618, will be ready for a review when we decide on rating position.

nnn:

apis/client/view/header/header.html
```diff
@@ -2,43 +2,37 @@
...
+          {{> viewApiStatus api=api width="0.4" }}
+        </h1>
+        <ul class="list-inline view-api-meta-details">
+          <li>
+            <i class="fa fa-user"></i>
+            {{ api.getApiManagersByName }}
+          </li>
+        </ul>
+      </div>
+      <div class="page-head-starsblock pull-right">
```
brylie: Lets rename this from 'starsblock' to something more general, since it contains more page elements than the star rating widget. E.g. page-head is fine.

apis/client/view/header/header.html
```diff
@@ -2,43 +2,37 @@
...
+          {{> favourite api=api }}
+        {{else}}
+          <a href="#" class="star-button">
+            <div
+                class="api-card-bookmark"
+                data-toggle="tooltip"
+                data-placement="bottom"
+                title="{{_ 'catalogue_bookmarkTooltip_nonUser'}}">
+              <span class="api-card-bookmark-number">{{ api.bookmarkCount }}</span>
+              <i class="fa fa-star bookmark-default"></i>
```
brylie: This should no longer be a star, since it is related to bookmarking. Also, the page text and button class should be changed to use the word 'bookmark' instead of 'star'. Will these changes be part of a different PR?

apis/client/view/visibility/visibility.js
```diff
@@ -1,36 +1,35 @@
...
+    const instance = this;
+
+    // Get API Backend from data context
+    const api = this.api;
+
+    // Get ID of current service
+    const apiId = api._id;
+
+    // Check visibility status
+    const status = Apis.findOne(apiId).isPublic;
```
brylie: This can be shortened, since the API is already part of the template data context. In order to get the template data context in a way that is reactive, use `Template.currentData()`, rather than `instance.data`.

http://blazejs.org/api/templates.html#Template-currentData

apis/client/view/header/header.html
```diff
@@ -2,43 +2,37 @@
...
-                  data-placement="bottom"
-                  title="{{_ 'catalogue_bookmarkTooltip_nonUser'}}">
-                <span class="api-card-bookmark-number">{{ api.bookmarkCount }}</span>
-                <i class="fa fa-star bookmark-default"></i>
-              </div>
-            </a>
-          {{/if}}
-        </div>
-        {{#if api.currentUserCanEdit }}
-          {{> visibilitySwitch api=api }}
```
brylie: I think this reduces the usablility for the *visibility toggle* feature for several reasons:
- We have lost the explicit text saying whether the API is *public* or *private*
- The *visibility toggle* has been moved to the settings page, and nested among other fields
- The *visibility toggle* button on the settings page does not say what the current state is
  - it rather says the *opposite* of what the current stat is, as a verb

jykae:
Reviewing

brylie:
I think we need to review this PR as a group. Ping @bajiat @Nazarah

apis/client/view/header/header.html
nnn: Yes, different PR.

brylie:
I will defer to @jykae as to whether this gets merged.

However, I would like to have more clear discussion about changes like this, including defining related task more specifically in the future.

nnn:
@jykae, please review.

apis/client/view/header/header.html
jykae: @brylie I agree, though visibility feature is going to have third option "partly private".
@Nazarah has wireframe for new UI.

jykae:
@nnn This does not look too good in my eye, could API key component be on left under API logo, name, owner information? @Nazarah opinion?

nayttokuva 2016-10-03 kello 15 40 02

brylie:
@jykae is this task still under review, or are you waiting for follow-up from @nnn?

I removed the 'ready for review', so that conflicts could be resolved.

jykae:
Changes discussed and approved with @nnn @brylie

PR #1659, opened 2016-09-29
Feature/delete api enhancement

### LINKED ISSUE #1627: Deleting an API needs to delete also Proxy backend
If owner deletes an API from Apinf, proxy backends needs also to be deleted. Also, the corresponding API backend needs to be deleted from API Umbrella.

# Definition of done
- [x] Clicking 'Delete' on the API Profile: Settings tab deletes the Api from API Umbrella

jykae:
Closes #1627


brylie:
The code looks good. However, I am having trouble accessing our Nightly API Umbrella instance to verify that the delete works.

PR #1670, opened 2016-09-30
Update copyright notice

bajiat:


bajiat:
@brylie Can you check and merge?

PR #1675, opened 2016-10-03
Old PR: add api monitoring

### LINKED ISSUE #1512: Allow owner to switch on API monitoring
Currently the API monitoring in APInf is providing false results. See error #1386. We should allow the owners to select whether the API is being monitored or not and to choose the URL. At first phase, we are not providing authentication for monitoring.

See also comments and screenshots from Monitoring EPIC #1183. 

# User story
> As an API owner I want to be able to switch on API monitoring

> As an API owner I want to be able to select the URL that is being monitored

# Definition of done
* [x] API owner can enable the API monitoring either from API settings or from Monitoring tab..
* [x] API owner can select / give the URL to be monitored. 
  * [ ] Optionally, give a default for URL and allow them to change/edit. 
* [x] Monitoring happens with http request instead of pinging.
* [x] If monitoring is enabled, the monitoring call is made once in hour(?)
* [x] Only show the monitoring icon in API profile and catalog, if API is being monitored (owner has switched that on)

# Wireframe

When Monitoring is not enabled
apimonitoring

On enabling API monitoring, the URL field gets activated. Owner is allowed to use only GET method for API monitoring. Owner should add path to the base URL of the API to get valid results about API status. info icon should describe how the API process works. 
apimonitoring2

For APIs with Monitoring Enable, small circles indicating API health should appear in API profile (see wireframe2), card view and table view in API catalog (indicated with squares)
apimonitoringcatalogview

apimonitoringtableview

### LINKED ISSUE #1515: Schema and collection for storing API monitoring status
In order to notify the API owners only when their API monitoring status changes, we need to store the API monitoring statuses in a collection. This collection would also allow us later to create chart/analytics about API monitoring.

# Definition of done
* [x] Schema and collection for API monitoring status exists.
* [x] The response for every monitoring call/request is stored in the collection along with the time and date of request.
* [x] The collection is linked to the API collection.

marla-singer:
## Changes
- Create a new collection Monitoring: schema, permissions, publications
- Linked the Monitoring and APIs collections
- Create a new template, realize the layout, add this on a new section on API Settings tab
- Create the new server methods for manipulations with Cron
- When meteor server is restarted, cron is restarted too
- Update status indicator icon:
  * nothing icon: monitoring is disable
  * blue:  status code is 1xx
  * green: status code are 2xx or 3xx
  * orange: status code is 4xx
  * red: status code is 5xx
  * grey: monitoring is enable but status code is unknown
- Change APIs collection schema: add field status_code for saving server status code
-  Add auto-value field  on template 'Add api' with status_code = -1 by default
- Upgrade status.js: checking on status and returning className and statusText have own function

## Testing
If you don't want to wait until the cron will work for an hour, you can replace the value 'every 1 hour' on the 'every 1 minute' in this line:

https://github.com/apinf/platform/pull/1675/files#diff-1a813c094792511f986d28c7e58890bbR18


marla-singer:
@brylie Could you please review it?

brylie:
@marla-singer please fix conflicts.

brylie:
One general request would be that you make smaller commits. E.g. touching only a couple of files and/or relevant lines within a file. This makes it easier to provide meaningful commit messages.

Commit https://github.com/apinf/platform/pull/1675/commits/933a0e5f72e0aefa75afdd7d251119db61060ad0 contains the bulk of code changes for this PR, and is quite large.

brylie:
I have tried adding the following GET endpoints to an API:

- https://apinf.io
- https://google.com

However, the status icon does not update from the 'Loading...' state.

brylie:
As @bajiat has previously suggested, we may only need to default to GET requests for now. So, we can remove the 'Method' field.

monitoring/client/monitoring.html
```diff
@@ -0,0 +1,48 @@
...
+        {{> afQuickField name='enabled'}}
+
+        {{#if afFieldValueIs name='enabled' value=true}}
+          <label for="endpoint-monitor-field">
+            {{ afFieldLabelText name='url'}} &nbsp;
+          </label>
+          <div class="form-group">
+            <div class="row">
+              <div class="col-sm-1">
+                <input type="text" placeholder="GET" class="form-control" readonly>
```
brylie: Lets remove this placeholder for now, since it is unnecessary. It is a bit confusing, e.g. I was trying to change the value without realizing what I was doing.

The help text is already quite informative.

monitoring/client/monitoring.js
```diff
@@ -0,0 +1,47 @@
...
+
+  // Get api id
+  const apiId = instance.data.api._id;
+
+  // Subscribe on Monitoring collection
+  instance.subscribe('monitoring', apiId);
+});
+
+Template.apiMonitoring.onRendered(function () {
+  $('[data-toggle="popover"]').popover();
```
brylie: Add comment describing why we are showing a popver. What does it do?

monitoring/collection/schema.js
```diff
@@ -0,0 +1,50 @@
...
+      let validation;
+
+      // Require editor host if apiDocumentationEditor.enabled is checked
+      if (monitoringUrlEnabled === true && !monitoringUrl) {
+        validation = 'required';
+      }
+      return validation;
+    },
+  },
+  requests: {
```
brylie: We may need to move the 'requests' into their own collection. E.g. if we want to 'get only the latest request for a given API' we would have to get a (potentially) huge document containing all requests and then fiddle with the 'requests' field array to find the latest one.

brylie: The 'requests' collection would look something like this:

- monitoringId
- date
- statusCode
- apiStatus

Then, we could do a `MonitoringRequests.findOne({}, {sort: {date: -1});` to get only one most recent response.

monitoring/collection/schema.js
```diff
@@ -0,0 +1,50 @@
...
+  },
+  'requests.$.date': {
+    type: String,
+    optional: true,
+  },
+  'requests.$.status_code': {
+    type: String,
+    optional: true,
+  },
+  'requests.$.apiStatus': {
```
brylie: What is the diffference between `status_code` and `apiStatus`? Do we need both fields, or would a collection and/or template helper suffice?

monitoring/collection/server/publications.js
```diff
@@ -0,0 +1,9 @@
+// Meteor packages import
+import { Meteor } from 'meteor/meteor';
+
+// APINF import
+import { Monitoring } from '/monitoring/collection';
+
+Meteor.publish('monitoring', function (apiId) {
+  return Monitoring.find({ apiId });
```
brylie: This could potentially return a *huge* document, depending on how long the API has been under monitoring. Consider the refactoring suggestions above, so that the Monitoring document does not grow to be too large.

monitoring/server/cron.js
```diff
@@ -0,0 +1,74 @@
...
+import { Monitoring } from '/monitoring/collection';
+import { Apis } from '/apis/collection';
+
+// NPM packages import
+import _ from 'lodash';
+
+Meteor.methods({
+  startCron (apiId, url) {
+    // Create unique name for Cron job
+    const uniqueName = 'Monitoring: ' + apiId;
```
brylie: Us ES2015 template string. 
```js
  `Monitoring: ${apiId}`
```

monitoring/server/cron.js
```diff
@@ -0,0 +1,74 @@
...
+
+    // Set time of the monitoring call
+    const cronTime = 'every 1 hour';
+
+    // Update api status to 'Loading...'
+    Apis.update(apiId, { $set: { status_code: 0 } });
+
+
```
brylie: Remove extra line

monitoring/server/cron.js
```diff
@@ -0,0 +1,74 @@
...
+      name: uniqueName,
+      schedule (parser) {
+        // parser is a later.parse object
+        return parser.text(cronTime);
+      },
+      job () {
+        // Call HTTP request
+        Meteor.http.get(url, {}, (error, result) => {
+          // Set status code
+          const serverStatusCode = result ? result.statusCode : 500;
```
brylie: A 500 error means there was a server side problem. We should probably not assume a 500 error. A 404 might be more 'correct', but it is hard to anticipate why this call might fail.

apis/collection/schema.js
```diff
@@ -25,11 +25,17 @@ Apis.schema = new SimpleSchema({
     type: String,
     optional: true
   },
-  
+  status_code: {
```
brylie: Lets not store the `status_code` in the API document itself. Rather, a collection helper, such as `latestMonitoringStatusCode` could return the latest value. Follow the example of the Ratings module to see how storing individual 'monitoring requests' could be done.

brylie: Alternatively, rename this field to be `latestMonitoringStatusCode`.

monitoring/client/autoform.js
```diff
@@ -0,0 +1,72 @@
...
+
+          // Link Monitoring collection with Apis collection
+          Apis.update(apiId, { $set: { monitoringId: result } });
+
+          // Start Cron
+          Meteor.call('startCron', apiId, monitoring.url);
+        }
+      },
+    },
+    onSuccess () {
```
brylie: In addition to starting the Cron job, we need to make a 'GET' request to the monitoring backend immediately. This way the user can see the API status immediately.

monitoring/server/cron.js
```diff
@@ -0,0 +1,74 @@
...
+          const monitoringData = {
+            date: new Date(),
+            status_code: serverStatusCode,
+          };
+
+          // Update an api status
+          Apis.update({ _id: apiId }, { $set: { status_code: serverStatusCode } });
+
+          // Add the monitoring data in Collection
+          Monitoring.update({ apiId }, { $push: { requests: monitoringData } });
```
brylie: This array could grow quite large. Hence, the suggestion to store 'monitoring requests' in their own collection. That would allow easier querying and granular data control, without over-publishing data to the client.

monitoring/server/cron.js
```diff
@@ -0,0 +1,74 @@
...
+        });
+      },
+    });
+  },
+  stopCron (apiId) {
+    // Create unique name for Cron job
+    const uniqueName = 'Monitoring: ' + apiId;
```
brylie: Use ES2015 template syntax:

```js
  `Monitoring: ${ apiId }`
```

marla-singer:
> However, the status icon does not update from the 'Loading...' state.
The icon is updated when template is rendered. Try to refresh the api page. Also make sure that the apis collection has relevant structure with new filed "statsu_code"

brylie:
@marla-singer I have refreshed the page, and the indicator did not update. It may have been that the Cron job does not automatically run.

marla-singer:
@brylie If cron is running, Meteor console must contain log about it as `SyncedCron: Scheduled "Monitoring: TFL2EE76A5ptgDkMj" next run @Mon Oct 03 2016 15:34:33 GMT+0300 (MSK)`

brylie:
In any case, the icon should be made to update reactively (requiring no page refresh). This can be done by accessing the 'monitoring' template data context using the reactive `Template.currentData()` method. From `Template.currentData()` you can access the `api.latestMonitoringRequest` field, to update the template display.

brylie:
@marla-singer cron begins, but does not appear to run the job. I will try in Chrome.

marla-singer:
@brylie Did you have logs like `SyncedCron: Starting "Monitoring: TFL2EE76A5ptgDkMj".`
and `SyncedCron: Finished "Monitoring: TFL2EE76A5ptgDkMj"`?

brylie:
Well, the template icon works in both Chrome and Firefox. There seems to be a big delay between adding the API for monitoring and when the Cron runs. I suggested calling the Monitoring method one time in the AutoForm.onSuccess callback, to immediately update the API status.

brylie:
Yes, the cron starts, but takes a while to run. I can see the console messages.

brylie:
I just added a cron, and see the following message:

> SyncedCron: Scheduled "Monitoring: 5PDxGw3xa857KNsjW" next run @Mon Oct 03 2016 15:38:38 GMT+0300 (EEST)

brylie:
However, when refreshing the page, the grey 'Loading...' icon appears.

marla-singer:
If you don't change the value 
```javascript 
const cronTime = 'every 1 hour';
```
on new value
```javascript 
const cronTime = 'every 1 minute';
```
in /monitoring/server/cron.js file in 18 line, the icon will appear only within an hour

brylie:
Right. We can improve that later. However, please also add a single call to the Monitoring method in the onSuccess callback. That way, the API status will return immediately, no matter how often the Cron job is configured to run.

monitoring/server/cron.js
```diff
@@ -0,0 +1,74 @@
...
+    // Create cron working
+    SyncedCron.add({
+      name: uniqueName,
+      schedule (parser) {
+        // parser is a later.parse object
+        return parser.text(cronTime);
+      },
+      job () {
+        // Call HTTP request
+        Meteor.http.get(url, {}, (error, result) => {
```
brylie: Move this code into a Meteor method, so that it can be called from outside of the Cron job, as well as used within the Cron job.

marla-singer: When we disabled monitoring, form is submited and onSucces is running too. Is it a good idea to run Meteor method in the onSuccess callback?

brylie: You can check the form ID from within the onSuccess callback.

brylie: Alternatively, you might hook into a different AutoForm hook, or consider using collection hook(s).

brylie: From within your onSuccess hook, how can you check the `enabled` field value? E.g. if `enabled` is true, you can call the Meteor method.

Basically, the end goal here is to call the HTTP request once when enabling or updating monitoring settings for a given URL.

marla-singer: Aha, I think I got what do you mean. You suggest to call the HTTP request on Success hook, got the api status, update the indicator but don't start cron. Cron will be start on insert and update as before. Right?

monitoring/collection/schema.js
marla-singer: Well, sum up, Monitoring collection is needed to refactor. There will be two collections for api monitoring storage:
- **Monitoring collection** stores monitoring settings like api id, url, header(s), monitoring frequency, http method;
- **MonitoringRequests** collection stores _only requests_ for each api like date and server status code;

monitoring/server/cron.js
brylie: Sure, the cron can start wherever you decide it should. I just want to make sure that we get a status message back, if possible, right away. This way, the API manager won't have to wait for the cron to run.

brylie: Also, moving the status check function into its own 'Meteor method' could allow us to re-use the method elsewhere. E.g. we could trigger it from a button click.

apis/collection/schema.js
marla-singer: Is it a good practice to do request to database every hour for each api with an enable monitoring ?

marla-singer: Also these can be situation when we go to Catalogue page, app does a request to database to show an available apis plus does N requests to database to know status for these N apis?

marla-singer:
@brylie Changed the small remarks. In work:
- Refactor Monitoring collection
- Upgrade the api indicator: do icon reactively (requiring no page refresh).

marla-singer:
This isn't relevant PR.
That's new #1702

brylie:
@marla-singer, should we close this and delete the branch?

marla-singer:
@brylie yeah, of course!

PR #1680, opened 2016-10-03
Feature/clarify disable api key check

### LINKED ISSUE #1679: API Key Check field logic is inverted
The 'Disable API Key check' field on Proxy Backend has inverted logic. The Bootstrap Switch logic did not align with the underlying schema, nor the final setting on API Umbrella.

Remove the Bootstrap switch element and clarify the UI text to match the expected behavior.

brylie:
Closes #1679 

# Changes
- Remove Bootstrap Switch from Proxy Backend field
- Update i18n texts to match intended field behavior


brylie:
@nnn I removed the Bootstrap Switch, since I had not labeled the states correctly. There is also a UI glitch in Firefox, where the Bootstrap Switch button has strange padding.

screenshot from 2016-10-03 14-12-37


Please reveiw.

nnn:
Works as expected. Merging.

PR #1682, opened 2016-10-03
Feature/fix i18n token usage

### LINKED ISSUE #1681: Improve consistency of i18n token usage
We have small  inconsistencies in how we use i18n tokens. Improve i18n token usage by, e.g., removing spaces and using consistent quoting.

brylie:
Closes #1681 

# Changes
- Remove unnecessary spaces from before `_` in template helpers
- import TAPi18n

### DIFF
--- a/apis/flags/client/flag.html
+++ b/apis/flags/client/flag.html
@@ -5,7 +5,7 @@
         {{ # autoForm collection="ApiFlags" id="insertApiFlag" type=formType doc=apiFlag }}
         <div class="modal-header">
           <h4 class="modal-title">
-            {{ _ "flagApiModal_titleText" }}
+            {{_ "flagApiModal_titleText" }}
           </h4>
         </div>
         <div class="modal-body">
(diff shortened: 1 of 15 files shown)


brylie:
@frenchbread please review.

Ping @bajiat

frenchbread:
@brylie Shouldn't we move towards common approach we are taking with templates like:
 ```
{ > templateName }
```
And have the same pattern with i18n strings?
```
{ _ 'template_headerText' } 
```

brylie:
A quick search of our code found the following counts
- 139 results for `{{>`
- 22 results for `{{ >`

I think we should prefer the first version for two reasons:

- it is more common in our code, currently
- the `>` character is not an English word, so is a bit noisy

E.g. we tend to keep the `#` character next to the `{{` in template block helpers:

- 107 results for `{{#`
- 41 results for `{{ # `

frenchbread:
> both cases are more common in our code, currently

Before using `eslint` for js we did not care about spacing in the code. After we started using it we refactored all the code to fit it, and to have spaces.

> combining the {{> and {{# into single symbols becomes a semantic unit

Disagreed here, since `{{ }}` are used to tell parser to operate differently than to HTML. `>`, `#` and `_` are kind of operators here for specific cases.

I think it's better to keep `{{ > someThing }}` "with spaces" approach here, since it is:
- Consistent
- Easy to read

In terms of rewriting all the code to the selected approach, is just a matter of opening search bar in atom with `Cmd+Shift+F` and replacing one peace of the code to another.

brylie:
> {{ }} are used to tell parser to operate differently than to HTML. >, # and _ are kind of operators here for specific cases.

 Right, both are used primarily to help the computer (parser) to do its job. Combining them makes it easier for the human to read, and still accomplishes the same. I.e. `{{_`,  `{{>` and `{{#` are easy to read and distinguish (without spaces), and the parser can still do its job just fine.

brylie:
@frenchbread I realize our opinions are different, and there is no *right* or *wrong* here.

I would like to fix the other handlebars helpers, for consistency.

In a nutshell, I am proposing we remove the space infix because:

- the spaces are unnecessary
- spaces do not necessarily increase readability
  - i.e. template helpers without spaces between the prefix are easy to read, since the last character is easy to distinguish
- most template helpers in our code *do not* separate the handlebars (`{{`) from the prefix (`_`, `>`, `#`)

brylie:
If it would help, we could do a quick vote/poll regarding this issue. Otherwise, I can just do the search/replace as part of this PR.

Ping @bajiat

PR #1702, opened 2016-10-06
Api monitoring v 2

### LINKED ISSUE #1512: Allow owner to switch on API monitoring

### LINKED ISSUE #1515: Schema and collection for storing API monitoring status

marla-singer:
Closes #1512 
Closes #1515

## Changes
- Create a new collection Monitoring: schema, permissions, publications
- Linked the Monitoring and APIs collections
- Create a new template, realize the layout, add this on a new section on API Settings tab
- Create the new server methods for manipulations with Cron
- When meteor server is restarted, cron is restarted too
- Update status indicator icon:
  * nothing icon: monitoring is disable
  * blue:  status code is 1xx
  * green: status code are 2xx or 3xx
  * orange: status code is 4xx
  * red: status code is 5xx
  * grey: monitoring is enable but status code is unknown
- Change APIs collection schema: add field status_code for saving server status code
-  Add auto-value field  on template 'Add api' with status_code = -1 by default
- Upgrade status.js: checking on status and returning className and statusText have own function

## After reviewing
- Remove methods field
- Add comment
- Change default status code fom 500 to 404
- Moving the status check function into its own Meteor method
- Add a single call to the Monitoring method in the onSuccess callback
- Rename field in apis collection
- Refactoring of monitoring collection
- Do an icon indicator reactive
- Fix markup after rename class for indicator

### DIFF
--- a/apis/client/add/add.html
+++ b/apis/client/add/add.html
@@ -2,7 +2,7 @@
   <div class="container">
     {{# autoForm collection=ApisCollection id="addApiForm" type="insert" }}
       <!--invisible auto-value field-->
-      {{> afQuickField name='status_code' value="-1" type="hidden"}}
+      {{> afQuickField name='latestMonitoringStatusCode' value="-1" type="hidden"}}
       <fieldset>
         <legend>Add an API</legend>
         {{> afQuickField name='name' }}
--- a/apis/client/add/add.html
+++ b/apis/client/add/add.html
@@ -1,6 +1,8 @@
 <template name="addApi">
   <div class="container">
     {{# autoForm collection=ApisCollection id="addApiForm" type="insert" }}
+      <!--invisible auto-value field-->
+      {{> afQuickField name='status_code' value="-1" type="hidden"}}
       <fieldset>
         <legend>Add an API</legend>
         {{> afQuickField name='name' }}
--- a/apis/client/view/settings/settings.html
+++ b/apis/client/view/settings/settings.html
@@ -30,6 +30,7 @@
       {{/ autoForm }}
     </div>
   </div>
+  {{> apiMonitoring api=api}}
   <div class="panel panel-danger">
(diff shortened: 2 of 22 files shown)


brylie:
Great work! :smile:

PR #1722, opened 2016-10-10
Feature/unique prefix

### LINKED ISSUE #1200: Allow only unique proxy backend base paths

nnn:
Closes #1200

### DIFF
--- a/core/client/style/global.less
+++ b/core/client/style/global.less
@@ -48,7 +48,3 @@ hr {
   margin-bottom: 5px;
   font-weight: 700;
 }
-
-.panel-body p {
-  color: @gray;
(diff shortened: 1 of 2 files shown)


nnn:

PR #1745, opened 2016-10-10
Hotfix/dashboard not rendering

frenchbread:


brylie:
This appears to fix the errors. @frenchbread what is the status of this PR?

brylie:
Also, have you tested that the charts work when analytics data is available? I am not able to render any charts. What proxy backends have you tested with?

frenchbread:
@brylie I did not see any charts rendered either, but I assume that there is no data in ES.

brylie:
Ok, I have tested the charts to work.

One other thing, how can we get analytics for the `/api-umbrella/` endpoints? I.e. Admin users might want to see analytics for the Admin API, and this gives us quite a bit of data to render our charts.

brylie:
I have added the following option group to the select menu, but it does not work as expected:

```html
<!-- Admin API -->
<optgroup label="Admin" >
  <option value="api-umbrella" >
    Admin API
  </option>
</optgroup>
```

frenchbread:
Probably we should hardcode the `/api-umbrella/` in prefixesQuery variable to have it there.

frenchbread:
Please review

brylie:
Sounds good. We need to make sure the user has an Admin role, before adding the `/api-umbrella/` to the prefixes query. Can you do that as part of this PR please?

frenchbread:
This PR only shows a message to the user if proxy backends are not added, I think it should be that way

brylie:
Actually, I will just review, since this is a hotfix. Can you consider opening a feature branch so that admins can access the Admin API analytics?

frenchbread:
@brylie I got it working. I'll create a new issue & PR.

brylie:
Yay! :smile:

PR #1752, opened 2016-10-11
Hotfix/api umbrella analytics data

### LINKED ISSUE #1751: Allow Admin users to see Admin API analytics on the Dashboard
Currently, there is no way for Admin users to view usage statistics about the Admin API. This may be useful for diagnostic purposes, such as detecting performance bottlenecks.

Add an option to the *Select API* dropdown, that only Admin users can see, so that we can display analytics from the `/api-umbrella/` endpoint on the dashboard:

screenshot_20161011_104652

# Example code

```html
<!-- Admin API -->
<optgroup label="Admin" >
  <option value="/api-umbrella" >
    Admin API
  </option>
</optgroup>
```

frenchbread:
Closes #1751 

### Changes
- Added `/api-umnrella/` prefix to ES query
- Added UI option for `API Umbrella Logs`
- Permission checks
- Fix publication
- i18n strings


frenchbread:

brylie:
@frenchbread this works, but there are some issues:

- we need to let the admin user know what they are seeing
  - to this end, we should provide an option for "Admin API" in the dropdown menu
- If possible, the "Admin API" should include the `proxy.name` field, so the Admin user knows what they are viewing

# Example screenshot

brylie:
Ping @bajiat @philippeluickx

brylie:
@frenchbread basically, I think you can just wrap the above code example in an `if user is in role Admin` block, and we can merge this PR.

brylie:
@frenchbread after reviewing with @bajiat we think it would be good to remove the *API Umbrella Logs* option from the *My APIs* section,

screenshot_20161011_115844


The menu contains an *Administrator* section with *Proxy Admin API* section:

screenshot_20161011_115147

brylie:
Ping @bajiat is this your understanding of our discussion? If so, this PR is really close to ready.

frenchbread:
So how much sections are we having in the end and why removing MyApis ?

bajiat:
@frenchbread We are not suggesting to remove MyAPIs section. @brylie just had not added any APIs when he took the screenshot. 
Keep current sections. Just add an Administrator section and move the Proxy Admin API there. Also, rename "API Umbrella Logs" as "Proxy Admin API"

brylie:
Right, just remove **API Umbrella Logs** from the *My APIs* section:

screenshot_20161011_115844

frenchbread:
Done

frenchbread:
Please review

PR #1766, opened 2016-10-12
Hotfix/email verification

### LINKED ISSUE #1728: Allow user with admin role to login without email verification
To prevent further so called "black box lock-out" happening, allow users with admin role to log in without verified email address

Related: #1725 #1726 #1727

marla-singer:
Closes #1728 (not releated)
### Changes
- Update users collections when  updateMailConfiguration is called
- Lint updateMailConfiguration methods

### DIFF
--- a/settings/server/methods.js
+++ b/settings/server/methods.js
@@ -60,7 +60,8 @@ Meteor.methods({
         // Update admin account. Set 'verified: true' for all admin user
         Meteor.users.update(
           { roles: { $in: ['admin'] }, 'emails.0.verified': false },
-          { $set: { 'emails.0.verified': true } }
+          { $set: { 'emails.0.verified': true } },
+          { multi: true },
         );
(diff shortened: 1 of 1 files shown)


marla-singer:

brylie:
Excellent!

settings/server/methods.js
```diff
@@ -56,6 +57,12 @@ Meteor.methods({
           ${encodeURIComponent(smtpPort)}
         `;
 
+        // Update admin account. Set 'verified: true' for all admin user
```
jykae: @marla-singer I think better place for this could https://github.com/apinf/platform/blob/24a1f161a862a02e02419201307b8af8e681fa32/core/helper_functions/login_verify.js#L4

You could do also that like
```
Roles.userIsInRole(userId, ['admin']);
```

PR #1809, opened 2016-10-24
Allow only one primary rate limit for response headers

### LINKED ISSUE #1684: Rate limiting response_headers should work like radio button
Left as enhancement:
User should be able to select only one **'show rate limit in response headers'** value when multiple rate limit rules are defined. This is used as the primary `response_headers` value on API Umbrella, and there can be only one primary rule.

nayttokuva 2016-10-03 kello 17 59 02

jykae:
Closes #1684 

- onChange event handler to allow only one primary rate limit for response headers

### DIFF
--- a/proxy_backends/client/form/form.js
+++ b/proxy_backends/client/form/form.js
@@ -163,7 +163,7 @@ Template.apiProxy.events({
     const changedCheckbox = $(event.target);
 
     const checkboxesArray = $("input:checkbox[name*='response_headers']");
-    // If Primary limit checked, uncheck other
+    // If changedCheckbox is checked, uncheck other
     if (changedCheckbox[0].checked) {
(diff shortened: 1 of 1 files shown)


proxy_backends/client/form/form.js
```diff
@@ -156,4 +159,19 @@ Template.apiProxy.events({
       });
     }
   },
+  'change input:checkbox[name*="response_headers"]': (event) => {
+    const changedCheckbox = $(event.target);
+
+    const checkboxesArray = $("input:checkbox[name*='response_headers']");
```
marla-singer: Please add a comment for this line to describe an operation

marla-singer:
@jykae Works good. It's almost ready to merge

jykae:
@marla-singer comment added

marla-singer:
@jykae Merge it

PR #1825, opened 2016-10-26
Add condition if response from API umbrella has errors

### LINKED ISSUE #1822: Proxy settings never save when submitting invalid host
When submitting proxy settings for an API with an invalid host, the Proxy Settings form never submits. The end-user is not informed of any problem, even though API Umbrella likely rejected the submission since the host does not exist.

# Steps to reproduce

1. Add an API
  - enter any title and description
  - for the host, **enter a domain name that does not exist**
2. Go to the API details page
3. Click the Proxy tab
4. Enter proxy details
5. Click 'Save and Publish'

# Expected result
The proxy form should indicate to the user that the submission has failed, exiting the 'pending' state.

# Observed result
The proxy form remains in the 'pending' state, never completing the submission.

# Screencast
peek 2016-10-26 15-33

marla-singer:
Closes #1822

Test it for insert proxy backend configuration, not update

### DIFF
--- a/proxy_backends/client/form/autoform.js
+++ b/proxy_backends/client/form/autoform.js
@@ -38,11 +38,21 @@ AutoForm.hooks({
                 // Throw a Meteor error
                 Meteor.error(500, error);
                 return false;
-              } else if ( // If success, attach API Umbrella backend ID to API
+              }
+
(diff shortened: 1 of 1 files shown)


jykae:
@marla-singer nice addition

PR #1827, opened 2016-10-26
Replace condition for feedback button

### LINKED ISSUE #1824: Allow anonymous user to View feedback
When fixing #1817, anonymous user being able to remove feedback, also View feedback button was removed. As a quick fix, reintroduce View button for anonymous users.

I will create another  enhancement task for showing feedback e.g. as accordeon or panel and also checking editing options for feedback. This is just a quick fix.

marla-singer:
Closes #1824

### DIFF
--- a/feedback/client/single/single.html
+++ b/feedback/client/single/single.html
@@ -17,7 +17,7 @@
         {{ topic }}
       </span>
     </div>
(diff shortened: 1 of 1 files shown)


jykae:
Works nice 👍

PR #1828, opened 2016-10-27
Bugfix/searchpage fix

### LINKED ISSUE #1820: In Search APIs page, searchbox doesn't function
# Steps to reproduce:
* go to nightly.apinf.io
* click search button from topbar
* click search button in searchbox
* in Search APIs page, type "api" and click on search button in the searchbox

# Result
* no search for specific API doesn't take place
searchboxbug

# Expected Result
* Search button should be functioning
* should return a new search result

# Environment
* OS: Windows 8
* Browser: any

jykae:
Closes #1820 

## Proposed changes
- prevent usual submit with enter press in live query
- use template level jQuery as we have duplicate ID "search-text" (navbar search & search page input)

### DIFF
--- a/search/client/search.html
+++ b/search/client/search.html
@@ -7,7 +7,7 @@
     </div>
 
     <div class="row">
-      <form id="search-form">
+      <form id="search-form" onsubmit="return false;">
         <div class="input-group">
           <span class="input-group-addon">
             <i class="fa fa-search"></i>
--- a/search/client/search.html
+++ b/search/client/search.html
@@ -7,7 +7,7 @@
     </div>
 
     <div class="row">
-      <form id="search-form" onsubmit="return false;">
+      <form id="search-form">
         <div class="input-group">
           <span class="input-group-addon">
             <i class="fa fa-search"></i>
@@ -16,7 +16,7 @@
           type="text"
           placeholder="{{_ 'searchPage_searchForm_placeholderText' }}"
           class="form-control"
-          id="search-text" />
+          id="search-page-text" />
         </div>
       </form>
       <p class="help-block">
--- a/search/client/search.html
+++ b/search/client/search.html
@@ -16,7 +16,7 @@
           type="text"
           placeholder="{{_ 'searchPage_searchForm_placeholderText' }}"
           class="form-control"
-          id="search-page-text" />
+          id="search-field" />
         </div>
       </form>
       <p class="help-block">
--- a/search/client/search.js
+++ b/search/client/search.js
@@ -73,16 +73,16 @@ Template.search.onRendered(function () {
   const instance = this;
 
   // Update search field with current search value
-  instance.$('#search-page-text').val(instance.searchValue.get());
+  instance.$('#search-field').val(instance.searchValue.get());
 
   // Check if search parameter is set
   if (instance.searchValue.get()) {
     // Update search field with search value provided in the URL
-    instance.$('#search-page-text').val(instance.searchValue.get());
+    instance.$('#search-field').val(instance.searchValue.get());
   }
 
   // Put focus of search field on a page
-  instance.$('#search-page-text').focus();
+  instance.$('#search-field').focus();
 });
 
 Template.search.helpers({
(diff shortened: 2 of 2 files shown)


jykae:
For reviewer: should we use different IDs? Template level jQuery is anyway safer (and possibly even faster than whole page query)

search/client/search.html
```diff
@@ -7,7 +7,7 @@
     </div>
 
     <div class="row">
-      <form id="search-form">
+      <form id="search-form" onsubmit="return false;">
```
brylie: This may not need to be a form. E.g. just an `input` tag is necessary, and button to trigger the event.

brylie: So, we can remove the `form` element, and not worry about preventing the `submit` event.

brylie: Otherwise, the event handlers are typically defined in the related JavaScript file, under `Template.templateName.events`

search/client/search.js
```diff
@@ -71,16 +73,16 @@ Template.search.onRendered(function () {
...
 
   // Update search field with current search value
-  $('#search-text').val(instance.searchValue.get());
+  instance.$('#search-text').val(instance.searchValue.get());
 
   // Check if search parameter is set
   if (instance.searchValue.get()) {
     // Update search field with search value provided in the URL
-    $('#search-text').val(instance.searchValue.get());
+    instance.$('#search-text').val(instance.searchValue.get());
```
brylie: Rather than use `instance.$` lets use the Blaze [`instance.find()`](http://blazejs.org/api/templates.html#Blaze-TemplateInstance-find) method.

search/client/search.js
```diff
@@ -71,16 +73,16 @@ Template.search.onRendered(function () {
...
   }
 
   // Put focus of search field on a page
-  $('#search-text').focus();
+  instance.$('#search-text').focus();
```
brylie: Rather than use `instance.$` lets use the Blaze [`instance.find()`](http://blazejs.org/api/templates.html#Blaze-TemplateInstance-find) method.

search/client/search.js
```diff
@@ -71,16 +73,16 @@ Template.search.onRendered(function () {
   const instance = this;
 
   // Update search field with current search value
-  $('#search-text').val(instance.searchValue.get());
+  instance.$('#search-text').val(instance.searchValue.get());
```
brylie: Also, add a different ID to the `search-text` field, since IDs should be unique on a page.

search/client/search.js
```diff
@@ -112,14 +114,14 @@ Template.search.helpers({
 });
 
 Template.search.events({
-  'keyup #search-text': function (event, template) {
+  'keyup #search-text': function (event) {
```
brylie: Add a different ID to the `search-text` field, since IDs should be unique on a page.

> The id global attribute defines a unique identifier (ID) which must be unique in the whole document.
>
> Source: MDN  [id - HTML](https://developer.mozilla.org/en/docs/Web/HTML/Global_attributes/id)

search/client/search.js
```diff
@@ -112,14 +114,14 @@ Template.search.helpers({
...
     event.preventDefault();
 
     // Get reference to Template instance
     const instance = Template.instance();
 
     // Get search text from a text field.
-    const searchValue = $('#search-text').val();
+    const searchValue = instance.$('#search-text').val();
```
brylie: Rather than use `instance.$` lets use the Blaze [`instance.find()`](http://blazejs.org/api/templates.html#Blaze-TemplateInstance-find) method.

search/client/search.html
brylie:
```js
Template.searchForm.events({
  'submit #search-form' (event, instance) {
    // Prevent the 'submit' event
    event.preventDefault();
  }
});
```

jykae:
@brylie Requested changes done.

search/client/search.js
```diff
@@ -73,16 +73,16 @@ Template.search.onRendered(function () {
   const instance = this;
 
   // Update search field with current search value
-  instance.$('#search-text').val(instance.searchValue.get());
+  instance.$('#search-page-text').val(instance.searchValue.get());
```
brylie: Sorry, this ID is slightly confusing. 'Search page text' sounds like text appearing on a search page. `search-page-input`, even though this is already an `input` element.

jykae:
@brylie anything else, ready to merge?

brylie:
Looks good.

PR #1829, opened 2016-10-27
Hotfix/fix settings subscriptions

brylie:


jykae:
@brylie nice solution for settings publications 👍

jykae:
@brylie could you make PR against develop also? I don't delete branch yet.

PR #1830, opened 2016-10-27
Hotfix/fix settings subscriptions

brylie:


jykae:
merging security fix for develop also

PR #1837, opened 2016-10-28
Feature/api 404 redirect

### LINKED ISSUE #1261: TypeError: Cannot read property '_id' of undefined
### Observed
- On API Backend View page which should not exist - `TypeError: Cannot read property '_id' of undefined` error is shown. 
### Expected:
- 404 page is rendered
### Steps to reproduce:
- Open APInf on nighlty or locally
- Go to catalog
- Click on any API -> API backends page opens up
- Modify API's ID in the page URL to smth random & hit enter
- Open browser console -> `TypeError: Cannot read property '_id' of undefined`

brylie:
Closes #1261 
# Changes
- Show 404 error when API not found
  - Add server method to check for API, so that user doesn't get data before authorization method runs
- Lint other files/methods

### DIFF
--- a/apis/client/lib/router.js
+++ b/apis/client/lib/router.js
@@ -24,7 +24,6 @@ Router.route('/api/:_id/', function () {
 
   // Check if API exists
   Meteor.call('checkIfApiExists', apiId, function (error, apiExists) {
-    console.log(apiExists);
     // Check if API exists
     if (apiExists === false) {
       // If API doesn't exist, show 'Not Found'
--- a/apis/client/lib/router.js
+++ b/apis/client/lib/router.js
@@ -25,22 +25,22 @@ Router.route('/api/:_id/', function () {
   // Check if API exists
   Meteor.call('checkIfApiExists', apiId, function (error, apiExists) {
     // Check if API exists
-    if (apiExists === false) {
-      // If API doesn't exist, show 'Not Found'
-      route.render('notFound');
-    } else {
+    if (apiExists) {
       // Ensure current user has permissions to view backend
       Meteor.call('currentUserCanViewApi', apiId, (error, userIsAllowedToViewApi) => {
         if (userIsAllowedToViewApi) {
           route.render('viewApi');
           route.layout('masterLayout');
         } else {
+          // User is not allowed to view API
           Router.go('forbidden');
         }
       });
+    } else {
+      // If API doesn't exist, show 'Not Found'
+      route.render('notFound');
     }
   });
-
 }, {
   name: 'viewApi',
 });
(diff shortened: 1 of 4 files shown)


marla-singer:
@brylie Self assigned

marla-singer:
@brylie 
Something is wrong with navbar and Latest APIs section. Also my catalog is empty too but I have the apis. 
joxi_screenshot_1477655641159

marla-singer:
I watched MiniMongo on Catalog page, these doesn't have any subscribed apis item

marla-singer:
@brylie Sorry, looks like I have problem from my side

apis/server/methods/apiBackends.js
```diff
@@ -1,26 +1,20 @@
...
+    if (api) {
+      apiExists = true;
     } else {
-      // User is NOT authorized to view this API
-      return false;
+      apiExists = false;
     }
-  }
+
+    return apiExists;
```
marla-singer: I think it can be easy:
```javascript
if (api) {
  return true
} else {
  return false
}
```
And placeholder doesn't need

apis/server/methods/authorization.js
```diff
@@ -19,4 +19,40 @@ Meteor.methods({
...
+    } else {
+      // User is NOT authorized to view this API
+      userAuthorized = false;
+    }
+
+    return userAuthorized;
+  },
+  currentUserCanEditApi (apiId) {
+    // Placeholder for 'user can edit'
+    let userCanEdit;
```
marla-singer: You can also simplify this one:
```javascript 
 // Get API
// Check if user can view
return api && api.currentUserCanEdit()
```

brylie:
Ping @marla-singer. Changes made.

apis/client/lib/router.js
```diff
@@ -20,17 +20,27 @@ Router.route('/api/:_id/', function () {
...
 
-  // Ensure current user has permissions to view backend
-  Meteor.call('currentUserCanViewApi', apiBackendId, (error, userIsAllowedToViewApi) => {
-    if (userIsAllowedToViewApi) {
-      route.render('viewApi');
-      route.layout('masterLayout');
+  // Check if API exists
+  Meteor.call('checkIfApiExists', apiId, function (error, apiExists) {
+    // Check if API exists
+    if (apiExists === false) {
```
marla-singer: Edit this condition to 
```javascript 
apiExists === undefined 
```

brylie:
@marla-singer I inverted the conditional, to make it easier to read and avoid the `false` vs `undefined` choice.

marla-singer:
@brylie Nice solution :+1:  Works as expected. Merge it

PR #1840, opened 2016-10-28
Migrate on jshimko/meteor-launchpad base docker image

shaliko:
- Migrate on jshimko/meteor-launchpad base docker image (more secure, by default install right meteor version, less image layers).
- Changed port number on proxy (nginx with LetsEncrypt).
- Dropped extra port expose.


jykae:
Looks ok. Merging.

PR #1844, opened 2016-10-28
Hotfix/only admin permission

### LINKED ISSUE #1835: Enabled 'Only admin can add APIs' shows 'Add API' button in 'Dashboard view'
# Steps
1. Visit https://staging.apinf.io 
2. Login as admin  
3. Go to 'Settings' 
4. Enable 'Only platform administrators are allowed to add new APIs  
5. Save settings  
6. Log out  
7. Login as normal user  
8. Go to 'Dashboard'  
# Expected Result

Add API button shouldn't be there on navbar and dashboard view.  
# Actual Result

Add API button is visible in dashboard view. Submit filled 'Add API' form fails though.  
# Browser
1. Mozilla Firefox 48.0 
   (Mozilla Firefox for Ubuntu Canonical – 1.0 )  
2. Google Chrome 52.0.2743.116
# Operating System

Ubuntu 16.0 LTS  
# Screenshots

selection_072  
selection_073  
selection_074

marla-singer:
Closes #1835

### DIFF
--- a/core/client/navbar/navbar.html
+++ b/core/client/navbar/navbar.html
@@ -24,7 +24,7 @@
         <ul class="nav navbar-nav navbar-left">
           {{# if currentUser }}
             {{# if proxyIsDefined }}
-              {{# if userCanViewPage }}
+              {{# if userCanViewDashboard }}
                 <li class="{{ isActiveRoute 'dashboard' }}">
                   <a href="/dashboard">
                     <i class="fa fa-bar-chart" aria-hidden="true"></i>
--- a/core/client/navbar/navbar.html
+++ b/core/client/navbar/navbar.html
@@ -24,12 +24,14 @@
         <ul class="nav navbar-nav navbar-left">
           {{# if currentUser }}
             {{# if proxyIsDefined }}
-              <li class="{{ isActiveRoute 'dashboard' }}">
-                <a href="/dashboard">
-                  <i class="fa fa-bar-chart" aria-hidden="true"></i>
-                  {{_ "navbar_dashboard" }}
-                </a>
-              </li>
+              {{# if userCanViewPage }}
+                <li class="{{ isActiveRoute 'dashboard' }}">
+                  <a href="/dashboard">
+                    <i class="fa fa-bar-chart" aria-hidden="true"></i>
+                    {{_ "navbar_dashboard" }}
+                  </a>
+                </li>
+              {{/ if }}
             {{/ if }}
           {{/ if }}
           <li class="{{ isActiveRoute 'catalogue' }}">
--- a/core/client/navbar/navbar.js
+++ b/core/client/navbar/navbar.js
@@ -117,6 +117,36 @@ Template.navbar.helpers({
     // By default allowing all user to add an API
     return true;
   },
+  userCanViewPage () {
+    // Allow or not regular user to view Dashboard page
+    // It depends on onlyAdminsCanAddApis settings
+
+    // Get settigns document
+    const settings = Settings.findOne();
+
+    if (settings) {
+      // Get access setting value
+      // If access field doesn't exist, these is false. Allow users to view page
+      const onlyAdminsCanAddApis = settings.access ? settings.access.onlyAdminsCanAddApis : false;
+
+      // Allow user to view page because not only for admin
+      if (!onlyAdminsCanAddApis) {
+        return true;
+      }
+
+      // Otherwise check of user role
+      // Get current user Id
+      const userId = Meteor.userId();
+
+      // Check if current user is admin or manager
+      const userIsAdminOrManager = Roles.userIsInRole(userId, ['admin', 'manager']);
+
+      return userIsAdminOrManager;
+    }
+    // Return true because no settings are set
+    // By default allowing all user to add an API
+    return true;
+  },
 });
 
 Template.navbar.onRendered(() => {
--- a/core/client/navbar/navbar.js
+++ b/core/client/navbar/navbar.js
@@ -91,29 +91,31 @@ Template.navbar.helpers({
     return false;
   },
   userCanAddApi () {
-    try {
-      // Get settigns document
-      const settings = Settings.findOne();
+    // Get settigns document
+    const settings = Settings.findOne();
 
+    if (settings) {
       // Get access setting value
-      const onlyAdminsCanAddApis = settings.access.onlyAdminsCanAddApis;
+      // If access field doesn't exist, these is false. Allow users to add an API on default
+      const onlyAdminsCanAddApis = settings.access ? settings.access.onlyAdminsCanAddApis : false;
 
+      // Allow user to add an API because not only for admin
       if (!onlyAdminsCanAddApis) {
         return true;
       }
 
+      // Otherwise check of user role
       // Get current user Id
       const userId = Meteor.userId();
 
       // Check if current user is admin
       const userIsAdmin = Roles.userIsInRole(userId, ['admin']);
 
-      return onlyAdminsCanAddApis && userIsAdmin;
-    } catch (e) {
-      // If caught an error, then returning true because no access settings is set
-      // By default allowing all user to add an API
-      return true;
+      return userIsAdmin;
     }
+    // Return true because no settings are set
+    // By default allowing all user to add an API
+    return true;
   },
 });
 
--- a/core/client/navbar/navbar.js
+++ b/core/client/navbar/navbar.js
@@ -117,7 +117,7 @@ Template.navbar.helpers({
     // By default allowing all user to add an API
     return true;
   },
-  userCanViewPage () {
+  userCanViewDashboard () {
     // Allow or not regular user to view Dashboard page
     // It depends on onlyAdminsCanAddApis settings
 


core/client/helpers.js
```diff
@@ -15,3 +22,29 @@ Template.registerHelper('niceName', function (_id) {
...
+
+    // Get current user Id
+    const userId = Meteor.userId();
+
+    // Check if current user is admin
+    const userIsAdmin = Roles.userIsInRole(userId, ['admin']);
+
+    return onlyAdminsCanAddApis && userIsAdmin;
+  } catch (e) {
+    // If caught an error, then returning true because no access settings is set
```
brylie: I am not confident of the conditions here. It is probably not a safe assumption to say that 'any error means the user can add an API'. 

Rather, our code should be explicit, for safety and self documentation:

- if only admins can add API and 
  - this **user is not an admin**, then 
    -  **user cannot add** an api
  - this **user is an admin**, then 
    - **user can add** an api
- otherwise, if user is logged in
  - **user can add** an api
- otherwise
  - **user cannot add** an api

brylie: The above code is intentionally verbose, so that we can easily spot semantic errors. It also has a safe default, *not allowing* a user to add an API.

brylie: Ping @frenchbread

marla-singer: @brylie Should I change code in this PR? I just only removed it

frenchbread: Agreed that it looks a bit messy. Ideally we can create a controller class with methods for each check so that in the end we helper's return will look like:
```js
Template.registerHelper('userCanAddApi', () => {
    const permissionController = new PermissionController();
    return permissionController.userIsAdmin() || (permissionController.userIsAdmin() && permissionController.onlyAdminsCanAddApis()) || (permissionController.userIsLoggedIn() && !permissionController.onlyAdminsCanAddApis())
});
```

Simplifying what @brylie documented:
User can add API when:
- User is an admin
- User is admin **and** `onlyAdminsCanAddApis` setting set to `true`.
- User is logged in **and** `onlyAdminsCanAddApis` setting set to `false`.

brylie: @marla-singer in the short-term, please fix the code to be more verbose, as part of this PR. 

We can consider whether a controller class is useful at a later point.

brylie: @frenchbread keep in mind also, we are working to make sure our code is very readable. This means making it clear 'why' the code is doing what it is doing. We don't need to be clever, just clear and simple.

frenchbread: @brylie I think it's readable enough. Even thought we can start making it more readable by putting entire sentences into variable names. E.g.

```js
Template.registerHelper('helperThatChecksIfCurrentUserCanAddAPI', () => {
    const instanceOfPermissionController = new InitializePermissionControllerClass();
    return instanceOfPermissionControllerClass.checkIfCurrentUserIsAdmin() || (instanceOfPermissionControllerClass.checkIfCurrentUserIsAdmin() && instanceOfPermissionControllerClass.onlyAdminUsersCanAddApisWhenSuchSettingIsSet()) || (instanceOfPermissionControllerClass.currentUserIsLoggedInToWebApp() && !instanceOfPermissionControllerClass.onlyAdminUsersCanAddApisWhenSuchSettingIsSet());
});
```

marla-singer: @frenchbread Nope! It looks awful

brylie: Words like 'instance', 'controller', and 'class' are not going to make it clearer what is going on.

brylie: “Debugging is twice as hard as writing the code in the first place. Therefore, if you write the code as cleverly as possible, you are, by definition, not smart enough to debug it.”

– Brian Kernighan

brylie: Keep it simple.

marla-singer:
@brylie Hm. I learnt code detaily and don't actually understand, why this solution is bad? The main logic of this part is good. I added the few comments and show error if it is catched. As for me everithing is fine and obviously

brylie:
> I learnt code detaily and don't actually understand, why this solution is bad?

The code can be improved for the following reasons:

- it is difficult to reason about
  - we have to think about all of the possible cases that could cause an error, in order to understand when the 'catch' statement would be called
- it has unsafe defaults
  - if **any** error is thrown, the user is 'allowed' to add an API 
- it is overly complicated
  - a simple if, else if, else flow would suffice here, and be more readable

marla-singer:
@brylie Well... I can't understand yet what do you want from my side. 


> we have to think about all of the possible cases that could cause an error, in order to understand when the 'catch' statement would be called

I shouldn't create the checking for every "sneeze". In our case I can only add the checking for settings exists or not  and no more to replace try/catch bundle. What kind of errors can be else? I don't know.

Please point to me a line where how do you think the code isn't good enough. My eyes look at code too much and they get tunnel vision

frenchbread:
Please correct me if I am wrong but I would leave try/catch since it's more safe. If changing it to if/else we would need to check each case 1. `settings` 2. `settings.access` 3. `settings.access. onlyAdminsCanAddApis`.

frenchbread:
Also try/catch is much more readable than stacks of if/elses.

marla-singer:
@frenchbread I think @brylie try to say that if error was cathced we wouldn't understand exactly why it happened. Because settings don't exist or field "access" isn't available. But yes - the long chain of if/elses doesn't look pretty and readable too.

frenchbread:
@marla-singer The first that can happen there is that settings were not set for current user. To resolve this we can create an empty settings document when user is created.

Second error is `onlyAdminsCanAddApis` value not being available since it was not set. To bypass this we can give default value to it when Settings document is created. In [this file](https://github.com/apinf/platform/blob/develop/settings/collection/schema.js) changing this part

```js
'access.onlyAdminsCanAddApis': {
    type: Boolean,
    optional: true,
    label: TAPi18n.__('settings_schema_onlyAdminsCanAddApis'),
},
```

to this

```js
'access.onlyAdminsCanAddApis': {
    type: Boolean,
    label: TAPi18n.__('settings_schema_onlyAdminsCanAddApis'),
    autoValue: function () {
      // Allow everyone to add APIs by default
      return false;
    }
},
```

frenchbread:
I can help with that.

marla-singer:
@frenchbread Is it a good idea to allow user to add api on default?

brylie:
@marla-singer the original feedback request was to structure the code somewhat like this:


>
1. Check if user is logged in, 
  - store it in a variable: e.g. 'userIsloggedIn'
1. Check if user is an admin, 
  - store it in a variable: e.g. 'userIsAdmin'
1. Get the value of 'onlyAdminsCanAddApis', 
  - store it in a shorter variable name
> 
> Then, create a nested conditional, using if/else if/else or something similar (rather than try/catch)
> 
- if only admins can add API and 
  - this **user is not an admin**, then 
    -  **user cannot add** an api
  - this **user is an admin**, then 
    - **user can add** an api
- otherwise, if user is logged in
  - **user can add** an api
- otherwise
  - **user cannot add** an api

brylie:
Feel free to make the code simpler or more elegant, while keeping descriptive comments. If there is a more logical order for the conditionals, use it instead.

The main guiding principles to consider, from [PEP 20](https://www.python.org/dev/peps/pep-0020/), are:

- Beautiful is better than ugly.
  - use as few if/else if/else statements as possible
  - consider a different flow control (aside from try/catch)
- Explicit is better than implicit.
  - the code and comments should clearly say what they are doing and why
- Simple is better than complex
  - if/else if/else is not too comlicated
  - I am not sure how to get rid of the nested conditions, aside from abstraction
- Complex is better than complicated.
- Flat is better than nested.
  - although try/catch is not desirable here without explicit error handling
- Sparse is better than dense.
- Readability counts.
- Errors should never pass silently.

marla-singer:
Thanks @brylie. I'll try to improve code and make it simple and readable :ok_hand:

brylie:
Cool. Thanks for your patience @marla-singer and @frenchbread

brylie:
One last consideration. We probably don't even need to allow users to access the Dashboard when they are not admin and the `onlyAdminCanAddApi` is true. In other words, there is no reason for a regular user to see the dashboard if they cannot add an API.

Taking the above into consideration may mean a different outcome, such as only running this authorization method in the Navbar, and showing a 403 not authorized to normal users when they view the Dashboard with `onlyAdminCanAddApi` set to true.

marla-singer:
@brylie regular users is user, who sign up and don't have any api, right? If it's so, I agree, there is no reason to see the Dashboard for them.

brylie:
> regular users is user, who sign up and don't have any api, right?

Jep. We have basically four roles:

- anonymous - person not logged in
- regular user - registered for an account
- manager - has added an API
- admin - platform superuser

When the `admin` user enables `onlyAdminCanAddApi = true`, then regular users have no benefit from seeing the Dashboard.

marla-singer:
@brylie Done. I did over the solution. 
Now regular users can't navigate to Dashboard via navbar. The enhacement is to add "forbidden" template, if regular user navigates to dashboard or add/new via url.

core/client/navbar/navbar.js
```diff
@@ -91,29 +91,61 @@ Template.navbar.helpers({
...
-      // If caught an error, then returning true because no access settings is set
-      // By default allowing all user to add an API
-      return true;
+      return userIsAdmin;
+    }
+    // Return true because no settings are set
+    // By default allowing all user to add an API
+    return true;
+  },
+  userCanViewPage () {
```
brylie: This should be called `userCanViewDashboard`, since it is only used in that context.

core/client/navbar/navbar.html
```diff
@@ -24,12 +24,14 @@
         <ul class="nav navbar-nav navbar-left">
           {{# if currentUser }}
             {{# if proxyIsDefined }}
-              <li class="{{ isActiveRoute 'dashboard' }}">
-                <a href="/dashboard">
-                  <i class="fa fa-bar-chart" aria-hidden="true"></i>
-                  {{_ "navbar_dashboard" }}
-                </a>
-              </li>
+              {{# if userCanViewPage }}
```
brylie: This should be called `userCanViewDashboard`, since we are only checking for permission to view Dashboard here. I.e. 'page' is ambiguous, and is not what we are really checking.

marla-singer:
@brylie Done.

PR #1848, opened 2016-10-31
login-issues

### LINKED ISSUE #1855: Login issues
Found issues:
* Cleanup onLogin hook, left from boilerplate, https://github.com/apinf/platform/blob/develop/users/server/accounts_hooks.js#L38
* Init userLoginAllowed variable, might cause return undefined, https://github.com/apinf/platform/blob/develop/users/server/login_verify.js#L9
* allow login in loginattemptverifier when mailsettings not enabled or given, https://github.com/apinf/platform/blob/develop/users/server/login_verify.js

jykae:
Closes #1855 

- cleanup legacy onLogin hook
- init userLoginAllowed variable (just in case we don't enter if-statement, would result returning undefined)
- add mailsettings check inside loginattemptverifier, allow login when mailsettings not enabled or given.

### DIFF
--- a/users/server/accounts_hooks.js
+++ b/users/server/accounts_hooks.js
@@ -1,6 +1,7 @@
-import Github from 'github';
+import { Meteor } from 'meteor/meteor';
+import { Accounts } from 'meteor/accounts-base';
 
-Accounts.onCreateUser(function(options, user) {
+Accounts.onCreateUser((options, user) => {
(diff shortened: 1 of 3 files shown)


jykae:
I am not not totally sure if this closes related issue, I could not reproduce "Internal server error" locally. Anyway ready for review :)

jykae:
@brylie added issue, updated PR description.

PR #1857, opened 2016-11-02
Fix api owner in api profile view

### LINKED ISSUE #1854: API owner is always admin in detail view
On production. When in the catalog the user is someone "not admin", in the detailview it shows "admin".

jykae:
Closes #1854 

- subscribe to allUsers, returns only usernames


marla-singer:
Works as expected, merge it

PR #1859, opened 2016-11-02
Cases for local & github registration

jykae:
Closes #1856 

- add blocks for local & github registration, no need to send verification email on Github registration

PR #1865, opened 2016-11-02
Github username taken

### LINKED ISSUE #1864: If Github username is taken, routing to user profile fails
Solution for Github username taken has been broken at some point of development, when we have improved settings. Route is wrong. Results 404 page not found.

joxi_screenshot_1478161060047

Overall implementation of this might be also bit too complex.

jykae:
Closes #1864


jykae:
works, but I clean it tomorrow

users/client/lib/router.js
```diff
@@ -35,14 +41,25 @@ Router.route('/settings/profile', {
   template: 'profile',
 });
 
+// Redirect to profile page if user doesn't have username
+// Eg. logged in with Github & username already taken
+const redirectToProfile = function () {
+  if (Meteor.userId() && !Meteor.user().username) {
```
brylie: Would the following work here instead?

```js
if (Meteor.user() && Meteor.user().username !== undefined) {
  ...
};
```

or, for example:

```js
// Get user, if exists
const user = Meteor.user();

if (user && user.username != undefined) {
  ...
};
```

- The `!== undefined` is more explicit
- since we are already relying on `Meteor.user()`, why introduce `Meteor.userId()`? 
  - E.g. keep the code consistent by using only one `Meteor.method()`
- storing the `user` object, or `undefined`, in a temporary variable might also reduce the noise here
  - only use `Meteor..` once
  - only one function call
- `Meteor.user()` [should return `null` if user is not logged in](http://docs.meteor.com/api/accounts.html#Meteor-user)

brylie:
This is a nice improvement, and is generally very readable. :sunglasses:

marla-singer:
@jykae Form allows me update profile without username
joxi_screenshot_1478162282813

marla-singer:
@jykae And can't navigate to Catalog and Add api page

brylie:
@jykae we should also notify the user why they have been redirected to the profile page, if this is not already happening. Otherwise, it would be confusing:

> why does it keep showing me the profile page?

marla-singer:
@brylie These has sAlert 
joxi_screenshot_1478162498656

brylie:
> Ah, I need to set my username.
>
> -- santa :santa:

marla-singer:
@brylie In PR #1859 I suggest to add popup like we have on Reset password. Maybe create another issue to improve an user usability?

brylie:
Cool, it looks like this PR is headed in the right direction.

users/client/lib/router.js
jykae: @brylie everything else I agree here but that explicitly saying "!== undefined", it should be common knowledge for JavaScript developer that it checks case "undefined".

jykae:
@marla-singer this logic setting for username if it is already taken has been like this for about year, as it was designed to function like this. 
This PR was basically meant to fix just redirect as route for profile has been changed when we have improved settings, and some cleanup for router code. 

If we need rethink usability add new issue.

marla-singer:
@jykae Okay, put aside popup and user usability. But form allows user to update information with empty value

jykae:
@marla-singer yep, I'm checking that one, good find 👍

jykae:
Interesting we have separate template for username that does not seem to be included anywhere https://github.com/apinf/platform/tree/develop/users/client/account/username O_o

jykae:
Also one other find, username also has different requirements on registration vs. profile form

nayttokuva 2016-11-03 kello 12 57 59

nayttokuva 2016-11-03 kello 12 58 30

brylie:
Cool. Set `optional: false` in UserSchema.

jykae:
@brylie I really hope it would be that simple. That solution provides "Internal server error" already in Github registration/login phase, when trying to create user, as username is required.

jykae:
@marla-singer now it doesn't allow empty username, but shows "Please set username" 2 times on first redirect to profile. :) I can't see where this bug hides..

jykae:
@marla-singer ready for review

marla-singer:
@jykae I'll merge both branch on local and test

marla-singer:
@jykae Detected problem and can't understand why.
1. Close all tab with site
2. Go to site again

Found from my side when router to site in first time: 
- Empty main page, only navbar
- After refresh page, it's appeared

jykae:
@marla-singer sorry, moved this.next outside if-statement, what about now?

users/client/lib/router.js
```diff
@@ -35,16 +42,38 @@ Router.route('/settings/profile', {
   template: 'profile',
 });
 
+// Redirect to profile page if user doesn't have username
+// Eg. logged in with Github & username already taken
+const checkUsername = function () {
```
brylie: This is an improvement in the function name. However, lets be just slightly more explicit, with something like:

- `ensureUsernameExists`
- `ensureUserHasUsername`

Using 'ensure' means that we are 'making sure' the user has a username, and is consistent with the `ensureSignedIn` function name.

users/client/profile/profile.js
```diff
@@ -1,5 +1,21 @@
 AutoForm.hooks({
   updateProfile: {
+    before: {
+      update (user) {
```
brylie: Can't we do this easier by simply making `username: { optional: false }` in the user profile schema?

jykae: @brylie Go ahead & try, I already reported functionality on registration/login if we set field required (see comment https://github.com/apinf/platform/pull/1865#issuecomment-258118447)

jykae:
@brylie Different simplified approach, took away unnecessary complexity, cleaned up, ready for review.

PR #1891, opened 2016-11-07
Maximum endDate today

### LINKED ISSUE #1730: Dashboard: Future date is clickable in date picker and on click dashboard keeps on loading
# Possible solution
It may be possible to fix this bug by editing/removing the following line:

- [/dashboard/client/charts/filtering/datepicker/datepicker.js:10](https://github.com/apinf/platform/blob/develop/dashboard/client/charts/filtering/datepicker/datepicker.js#L10)

```js
this.$('.input-daterange').datepicker({
  ...
  endDate: "+1d"
});
```

# Steps to reproduce
- login to nightly.apinf.io
- make sure you have an API and is connected to a proxy.
- click on dashboard.
- click on date picker.
-  result: date after current day is enabled (e.g. if today is 10/10/2016, then 11/10/2016 is also enabled)
  image
- click on that date.
  >  result: dashboard keeps on loading and charts don't appear eventually
  image
# Expected Result
- Only dates upto current day should be available for selection in date picker. Dates after current day should be disabled.
# System
- OS: Windows 8
- Browser: Chrome

brylie:
Closes #1730

# Changes
- set dashboard datepicker endDate to today

### DIFF
--- a/dashboard/client/charts/filtering/datepicker/datepicker.js
+++ b/dashboard/client/charts/filtering/datepicker/datepicker.js
@@ -7,7 +7,7 @@ Template.timeFrameSelectPicker.onRendered(function () {
   this.$('.input-daterange').datepicker({
     format: "dd M yyyy",
     todayHighlight: true,
-    endDate: "+1d"
+    endDate: "today"
   });
 
 })


marla-singer:
@brylie Nice, work as ecpected

PR #1892, opened 2016-11-07
Add unique ids

### LINKED ISSUE #1819: Add an id for header API-URL to make it easier to locate in selenium IDE
Add id for API URL and tabs in API profile page. Adding the ID will make the template more readable and will make automated testing easier, because items can be reach directly instead of going through DOM hierarchy or X-path. See comment for ítems where ID is needed.

image

Reference for correct id usage: 
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id
https://developer.mozilla.org/en-US/docs/Web/API/Element/id

jykae:
Closes #1819


jykae:
@Nazarah could you review this one, if I have ids everywhere needed?

PR #1899, opened 2016-11-08
Feature/version from package json

### LINKED ISSUE #1898: Get package version from package.json
We are currently bumping the package version numbers in two files:

- package.json
- /about/client/about.html

This is leading to errors.

In about.html/js, fetch the platform from package.json, so that we can manage our versioning from a single, standard file.

brylie:
Closes #1898

# Changes
- Add server method to fetch version from package.json
- Call server method in about template
- Show return value in about template

### DIFF
--- a/about/client/about.html
+++ b/about/client/about.html
@@ -28,7 +28,7 @@
               Apinf
             </dt>
             <dd>
-              0.34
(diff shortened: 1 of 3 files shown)


jykae:
@brylie nice work!

PR #1905, opened 2016-11-09
Feature/dashboard select single api

### LINKED ISSUE #1871: Allow only selecting one API from API select list
This is a temporary solution to allow owner and admin to select only one API at a time for the dashboard.

# Definition of done
* [ ] User (either owner or admin) can select only one API from the menu.
* [ ] APIs connected to different proxies are shown in the same list. This applies to owners, if they have connected their APIs to different proxies and to admins, if there is more than one proxy.
  *Note: APIs do not need to be distinguished by proxy.*

brylie:
Closes #1871 

# Changes
- reorganize dashboard filter code
  - rename api_select files
- remove multiple option from API select


jykae:
Awesome 🍾

PR #1913, opened 2016-11-10
Feature/allow select proxy

### LINKED ISSUE #1869: Allow API owner to select one from several configured proxies
If there is more than one API Umbrella proxy defined for deployment, allow the API owner to select one of them in the Proxy backend settings.

# Definition of done
* [x] If there is more than one API Umbrella defined, owner can select any of the defined proxies, when adding or editing proxy backend settings.
* [x] If there is only one API Umbrella defined, a simple solution is available for defaulting to the only API Umbrella.
* [x] Proxy base path will come from selected proxy.
* [x] Show an alert to owner, if switching between proxies: "This change will potentially break API calls from applications, if API users are not warned about the change."

Open issues: If owner changes the proxy, do we need to delete the proxy backend from the original proxy? Users will have a broken link, is it our responsibility to notify or is it owner's responsibility to warn the users?

# Wireframes

* In API profile Proxy tab, a dropmenu would appear depending on if there are more than 1 proxy available for selection. 
Open question: in case of only one proxy avialable, should we show the name as hard coded? 
apiprofilemultiproxyselector

* On selecting a specific proxy, associated fields for that proxy would be populated. In the case for only one available proxy, the fields can appear instantly Proxy tab is selected for viewing:

apiprofilemultiproxyselector2

apiprofilemultiproxyselector3

* When API owner tries to switch from one proxy to another, the following alert window should appear:
apiprofilemultiproxyselector4

marla-singer:
Closes #1869 

WIP

### DIFF
--- a/core/client/userMenu.html
+++ b/core/client/userMenu.html
@@ -34,6 +34,12 @@
           {{_ "userMenu_settings" }}
         </a>
       </li>
+      <li>
+        <a href="{{pathFor 'proxies' }}">
+          Proxies
+          <!--{{_ "userMenu_proxies" }}-->
+        </a>
+      </li>
       {{/ if }}
       <li class="divider"></li>
       <li>
--- a/apis/client/view/details/details.js
+++ b/apis/client/view/details/details.js
@@ -28,20 +28,18 @@ Template.apiDetails.helpers({
     // Get reference to template instance
     const instance = Template.instance();
 
-    // Get the proxy settings
-    // TODO: refactor this to support multi-proxy
-    const proxy = Proxies.findOne();
-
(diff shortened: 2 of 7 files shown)


core/client/userMenu.html
```diff
@@ -34,6 +34,12 @@
           {{_ "userMenu_settings" }}
         </a>
       </li>
+      <li>
+        <a href="{{pathFor 'proxies' }}">
+          Proxies
```
brylie: Add i18n token here.

brylie:
This PR looks good so far. Please remember to keep the code as *simple* and *clear* as possible. Great work so far!

PR #1914, opened 2016-11-11
Feature/cover photo scope improvements

brylie:
This is an idea to manage the scope of cover photo feature.


branding/cover_photo/client/upload/uploadButton/uploadButton.js
```diff
@@ -1,9 +1,84 @@
...
+        return;
+      }
+
+      const acceptedExtensions = ['jpg', 'jpeg', 'png', 'gif'];
+
+      if (fileNameEndsWith(file.file.name, acceptedExtensions)) {
+        // Turn on spinner
+        instance.uploadingSpinner.set(true);
+        // Upload the cover photo
+        return CoverPhoto.resumable.upload();
```
brylie: We might not need to `return` here. Rather, just call `CoverPhoto.resumable.upload()`

PR #1908, opened 2016-11-09
Reintroduce cover photo

marla-singer:
Closes #1823

PR #1915, opened 2016-11-11
Revert "Reintroduce cover photo"

brylie:
Reverts apinf/platform#1908


brylie:
@marla-singer please review.

PR #1929, opened 2016-11-17
Fix/return link errorpage

### LINKED ISSUE #1776: for "Access Denied" or "403 Forbidden" page, show a link to return to Catalog page
image

nnn:
Closes #1776

@apinf/developers please, review


core/client/error_templates/forbidden.html
```diff
@@ -10,6 +10,9 @@ <h1 class="headline text-danger">
           &nbsp;
           {{_ "forbidden_Message" }}
         </p>
+        <a href='#' class="go-back">
```
brylie: This should be an ID, since it is the only 'go-back' element on the page.

brylie: Also, we have had trouble with Internet Explorer when using `href="#"`. Consider changing this to a Bootstrap button, since it is a 'call to action' for the error page:

```html
<button class="btn btn-success">
  {{_ "forbidden_GoBack_Link" }}
</button>
```

That way, no href is necessary.

core/client/layouts/errors_layout/errors_layout.html
```diff
@@ -0,0 +1,13 @@
+<template name="errorsLayout">
```
brylie: Why do we need a new layout here? If this is the same markup as master_layout, why not just use master_layout?

core/client/layouts/errors_layout/errors_layout.js
```diff
@@ -0,0 +1,9 @@
+import { Template } from 'meteor/templating';
+import $ from 'jquery';
+
+Template.errorsLayout.onRendered(() => {
```
brylie: How about this, rather than creating a new layout, create a new template `goBackButton`. Then, include the `goBackButton` template in each of the error templates. That way we don't need a new layout, which basically duplicates *master_layout*.

core/client/lib/router.js
```diff
@@ -1,5 +1,5 @@
 Router.configure({
-  layoutTemplate: 'masterLayout',
+  layoutTemplate: 'errorsLayout',
```
brylie: Revert this to use `masterLayout`.

core/client/lib/router.js
```diff
@@ -26,13 +26,13 @@ Router.plugin('ensureSignedIn', {
 Router.map(function() {
   this.route('notAuthorized', {
     path: '/not-authorized',
-    layoutTemplate: 'masterLayout',
+    layoutTemplate: 'errorsLayout',
```
brylie: Revert this to use `masterLayout`.

core/client/lib/router.js
```diff
@@ -26,13 +26,13 @@ Router.plugin('ensureSignedIn', {
...
     render: 'notAuthorized',
   });
 
   this.route('forbidden', {
     path: '/forbidden',
-    layoutTemplate: 'masterLayout',
+    layoutTemplate: 'errorsLayout',
```
brylie: Revert this to use `masterLayout`.

core/client/layouts/errors_layout/errors_layout.html
brylie: Remove this layout. Use instead a template include, so that you don't have to duplicate the JavaScript logic.

core/client/error_templates/go_back_button/goBackButton.js
```diff
@@ -0,0 +1,9 @@
+import { Template } from 'meteor/templating';
+import $ from 'jquery';
+
+Template.goBackButton.onRendered(() => {
+  $('#go-back').click(() => {
+    history.go(-1);
```
brylie: How about using `history.back();` here, since it is more semantic. E.g. the `go-back` button triggers the `history.back()` method.

Also, would this work as a template event rather than jQuery?

```js
Template.goBackButton.events({
  'click #go-back': function () {
    // Go back one page in the browser history
    history.back();
  }
});
```

nnn:
@brylie sounds good, replaced.

PR #1932, opened 2016-11-17
Warn admin about proxy removal

### LINKED ISSUE #1920: Warn admin user, if removing a proxy from platform settings
If admin user wants to remove a proxy from Settings, that can (or rather will) break all API calls made to the connected proxy backends. User should be asked for a confirmation for removing the proxy with a message like this:

 "You are about to remove a proxy that has NN proxy backends connected to it. Removing the proxy will mean that API calls made to any of the connected API backends will break. Are you sure you want to remove the configured proxy?"

jykae:
Closes #1920


marla-singer:
@jykae Reviewing

jykae:
@marla-singer wait a moment, I am adding improved publication & some linting :) that's why I took ready tag away.

marla-singer:
Okay, am waiting for

jykae:
@marla-singer ok, ready

marla-singer:
@jykae Nice work 👍

PR #1937, opened 2016-11-18
Remove URL parameter instruction

### LINKED ISSUE #1349: API Key should not be passed as a parameter
Adding keys in URL is not a safe practice. It stays in (browser) history and call logs. (It is secured though: https://groups.google.com/forum/#!forum/api-craft)
Better option is to pass the key in the header of the requests, e.g. authentication header.

edit: found that URLs are secure, but still best practice to not include sensitive data in URLs

Edited by bajiat:
> Usernames, passwords, session tokens, and API keys should not appear in the URL, as this can be captured in web server logs, which makes them intrinsically valuable.
> NOT OK:
>https://example.com/controller<id>/action?apiKey=a53f435643de32 (API Key in URL)
Source: https://www.owasp.org/index.php/REST_Security_Cheat_Sheet

# Definition of done:
* [x] Remove instruction for passing API key in URL

jykae:
Closes #1349

### DIFF
--- a/apis/client/view/details/details.html
+++ b/apis/client/view/details/details.html
@@ -59,14 +59,10 @@
           {{# if hasApiKey }}
             <!-- Proxy Backend is completed, "Disable api key" is false, user has got an api key -->
             <li>
-              {{_ "apiDetails_usageInstructions_urlParameters_text" }}
-              <i>{{ proxyUrl }}&lt;optional_path&gt;?api_key={{ apiKey }}</i>
-            </li>
-            <li>
               {{_ "apiDetails_usageInstructions_requestHeaders_text" }}
               <span>
-            <i>X-Api-Key:</i> {{ apiKey }}
-        </span>
+                <i>X-Api-Key:</i> {{ apiKey }}
+              </span>
             </li>
           {{ else }}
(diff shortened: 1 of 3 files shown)


jykae:
## For reviewer
- check out the proposal to implement some real security instead of only instruction: https://github.com/apinf/platform/issues/1349#issuecomment-261474187

apis/client/view/details/details.html
```diff
@@ -59,14 +59,10 @@ <h3 id="api-url-title">
           {{# if hasApiKey }}
             <!-- Proxy Backend is completed, "Disable api key" is false, user has got an api key -->
             <li>
-              {{_ "apiDetails_usageInstructions_urlParameters_text" }}
```
marla-singer: Don't forget to delete an useless i18n tag

marla-singer:
> check out the proposal to implement some real security
@jykae What do you mean? Can you please tell how can I check out these?

jykae:
@marla-singer that one was note to check idea, if we would really like to restrict API key usage to header only but I left it enhancement because the original task was just to instruct safer way. I could add separate enhancement issue of that, if we decide to go that way.

jykae:
@marla-singer removed texts

marla-singer:
@jykae Merge it

PR #1923, opened 2016-11-14
Refactor Dashboard code

brylie:
Closes #1876 
Closes #1904

# Changes
- refactor state handling for proxy api backends check
- edit .dropdown-toggle font color (API Select List)
- refactor filter method to work with single proxy frontent prefix
- remove imperative 'loading' state UI management
- move start and end date state to URL

# Known issues
- Loading spinners have been disabled and need to be re-added (enhancement?)
- API Select does not trigger analytics reload
- Dashboard 'no proxy added' text is behaving oddly

PR #1949, opened 2016-11-22
Feature/query to right umbrella

### LINKED ISSUE #1872: Allow analytics query to be sent to the right API Umbrella proxy
When selecting API from the Select API list, the analytics query needs to be sent to the correct proxy (= the one connected to the selected API).

# Definition of done
* [ ] Analytics query is made to the correct proxy path.

*Consider whether the query should be made to ElasticSearch or API Umbrella Admin API. The latter is more secure and more future proof.*

marla-singer:
Closes #1872  
Based on #1923 
- Set proxy id of selected api in query parameters
- Unset proxy id if proxy isn't found
- stdout ElasticSearch URL for verify of correct work
- Add in functions another one parameter for proxy id

### DIFF
--- a/dashboard/client/charts/charts.js
+++ b/dashboard/client/charts/charts.js
@@ -447,12 +447,16 @@ Template.dashboardCharts.events({
     // Get selected value
     const frontendPrefix = event.target.value;
 
+    // TODO: Case is when Proxy Admin API is selected, which proxy is?
     // Find proxy id of selected api
     const proxyBackend = ProxyBackends.findOne({ 'apiUmbrella.url_matches.0.frontend_prefix': frontendPrefix });
 
     if (proxyBackend && proxyBackend.proxyId) {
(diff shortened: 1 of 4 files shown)


dashboard/client/charts/charts.js
```diff
@@ -446,6 +447,18 @@ Template.dashboardCharts.events({
     // Get selected value
     const frontendPrefix = event.target.value;
 
+    // TODO: Case is when Proxy Admin API is selected, which proxy is?
+    // Find proxy id of selected api
+    const proxyBackend = ProxyBackends.findOne({ 'apiUmbrella.url_matches.0.frontend_prefix': frontendPrefix });
```
brylie: We will move this to the server. Basically, all we need is the API ID to find the Proxy Backend. In other words, we have a relationship between APIs and Proxy Backends that we can traverse, rather than using frontend prefix.

PR #1950, opened 2016-11-22
Fix README links

shaliko:
Fix README links:

- LICENSE.
- Docker image metadata.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,8 @@
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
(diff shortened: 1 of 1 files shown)


shaliko:
@brylie Any comment on that request?

brylie:
Can we open this PR in the Apika repository? We are trying to minimize commits to Apinf while we work on Apika. That way, it will hopefully be easy for us to pull changes back from Apika after January.

shaliko:
@brylie sure, no problem.

PR #1964, opened 2016-12-03
Replaced backend_prefix regex with apiBasePathLooseRegEx

### LINKED ISSUE #1965: Add 'Append Query String Parameters' field to Proxy Backend form
# User story
> As an **API Manager**
I want to **add query parameters to every API request** made from the proxy
so that end users do not need to see or add the query parameters themselves

# Context
 API Umbrella, or primary proxy, has a field called 'Append Query String Parameters' on the Backend configuration form. However, we have not 'mirrored' the 'query string parameters' field in our user interface.

## API Umbrella screenshot
API Umbrella showing query parameters and request headers fields

# Feature
Add the 'query string parameters' field to our schema and user interface.

# Original idea
To make calling of prefix simple even with queryParams 

https://apinf.io/myPrefix -> https://proxyproxyservicesite.com/query?queryParam1=http://www.finalproxytarget.site/catalog/subcatalog/show.php&queryParam2=find('%20.my-class%20div')&queryParam3=true

Instead of (mapping myPrefix to query)

https://apinf.io/myPrefix?queryParam1=http://www.finalproxytarget.site/catalog/subcatalog/show.php&queryParam2=find('%20.my-class%20div')&queryParam3=true

(Tried - perhaps it is older version of apinf - in APIKA -> https://umbrella.digipalvelutehdas.fi/urheilukellot/?api_key=YOUR_KEY&site=gigantti.fi/catalog/puhelimet-ja-gps/fi-urheilukellot/urheilukellot&selector=find(%27a.product-name%27)&forceText=true )

And leaving for the user's responsibility all the queryparams.

santeriv:
Closes #1965 

Problem:
To make calling of prefix simple even with queryParams https://apinf.io/myPrefix 
-> https://html-to-json.herokuapp.com/jquery?site=https://www.tsr.fi/uudet-hanketiedotteet&selector=find(%27.list-items%20li%27)&forceText=true

And not like
https://apinf.io/myPrefix/?site=https://www.tsr.fi/uudet-hanketiedotteet&selector=find(%27.list-items%20li%27)&forceText=true

To make this kind of proxy work (which is proxy in itself, see https://github.com/santeriv/pulljson)

URL:
https://html-to-json.herokuapp.com
apiBasePath:
/jquery?site=https://www.tsr.fi/uudet-hanketiedotteet&selector=find(%27.list-items%20li%27)&forceText=true

Other option would have been to make work with just URL and no prefix:
URL:
https://html-to-json.herokuapp.com/jquery?site=https://www.tsr.fi/uudet-hanketiedotteet&selector=find(%27.list-items%20li%27)&forceText=true

But it seems not to work either. Tried with nightly build 2.12.2016 (https://nightly.apinf.io/).

* Not tested, no docker capable environment at hand, used just browser, regex was tested with regexpal 
* Thought also making a loose/strict checkbox but I wanted to await your reaction also, because first idea was to just replace existing regex. Tried to learn something about meteor app development also.
* Motivation: I do this just for fun and not for commercial purposes
*  Found this tool from twitter post : https://twitter.com/kyyberi/status/804708650925363200?lang=fi

### DIFF
--- a/proxy_backends/collection/regex.js
+++ b/proxy_backends/collection/regex.js
@@ -1,3 +1,5 @@
 export const proxyBasePathRegEx = new RegExp(/^\/[a-z0-9A-Z_\-\/]+\/$/);
 
 export const apiBasePathRegEx = new RegExp(/^\/[a-z0-9A-Z_\-\/]*$/);
+
+export const apiBasePathLooseRegEx = new RegExp(/^\/[a-z0-9A-Z-?=.&%()_\-\/]*$/);
--- a/proxy_backends/client/form/autoform.js
+++ b/proxy_backends/client/form/autoform.js
@@ -29,7 +29,7 @@ AutoForm.hooks({
           const requiredUrlMatches = apiUmbrella.url_matches &&
             apiUmbrella.url_matches[0] &&
             apiUmbrella.url_matches[0].frontend_prefix &&
-            apiUmbrella.url_matches[0].backend_prefix;
+            apiUmbrella.url_matches[0].backend_loose_prefix;
 
           const requiredServers = apiUmbrella.servers && apiUmbrella.servers[0] &&
             apiUmbrella.servers[0].host &&
@@ -112,7 +112,7 @@ AutoForm.hooks({
           const requiredUrlMatches = currentProxyBackend['apiUmbrella.url_matches'] &&
             currentProxyBackend['apiUmbrella.url_matches'][0] &&
             currentProxyBackend['apiUmbrella.url_matches'][0].frontend_prefix &&
-            currentProxyBackend['apiUmbrella.url_matches'][0].backend_prefix;
+            currentProxyBackend['apiUmbrella.url_matches'][0].backend_loose_prefix;
 
           const requiredServers = currentProxyBackend['apiUmbrella.servers'] &&
             currentProxyBackend['apiUmbrella.servers'][0] &&
--- a/proxy_backends/client/form/form.html
+++ b/proxy_backends/client/form/form.html
@@ -62,7 +62,7 @@
     <div class="row">
       <div class="form-group col-md-9">
         <label for="api-base-path-field">
-          {{ afFieldLabelText name='apiUmbrella.url_matches.0.backend_prefix' }}
+          {{ afFieldLabelText name='apiUmbrella.url_matches.0.backend_loose_prefix' }}
         </label>
 
         <div class="input-group">
@@ -72,7 +72,7 @@
           </span>
 
           {{> afFieldInput
-            name="apiUmbrella.url_matches.0.backend_prefix"
+            name="apiUmbrella.url_matches.0.backend_loose_prefix"
             class="form-control"
             aria-describedby="api-url"
             id="api-base-path-field"
@@ -80,9 +80,9 @@
         </div>
 
         <!-- validation messages -->
-        {{# if afFieldIsInvalid name='apiUmbrella.url_matches.0.backend_prefix' }}
+        {{# if afFieldIsInvalid name='apiUmbrella.url_matches.0.backend_loose_prefix' }}
           <p class="text-danger">
-            {{{ afFieldMessage name='apiUmbrella.url_matches.0.backend_prefix' }}}
+            {{{ afFieldMessage name='apiUmbrella.url_matches.0.backend_loose_prefix' }}}
           </p>
         {{/ if }}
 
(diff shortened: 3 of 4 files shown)


proxy_backends/collection/regex.js
```diff
@@ -1,3 +1,5 @@
 export const proxyBasePathRegEx = new RegExp(/^\/[a-z0-9A-Z_\-\/]+\/$/);
 
 export const apiBasePathRegEx = new RegExp(/^\/[a-z0-9A-Z_\-\/]*$/);
+
+export const apiBasePathLooseRegEx = new RegExp(/^\/[a-z0-9A-Z-?=.&%()_\-\/]*$/);
```
brylie: I would prefer to add these to the `apiBasePathRegEx`, rather than creating a second regex. It would be fewer lines changed in the PR, fewer to maintain, and would work the way you intend.

brylie:
This idea is good, but lets make the improvement with less impact. The current pull request makes changes to our database schema and adds a new regular expression, and doesn't cleanup unused code.

All that really needs to be changed here, from my understanding, is the `apiBasePathRegEx`.

brylie:
@santeriv, we really appreciate your contribution and thoughtfulness.

After further considering this pull request, I think it will not work as intended:

- our proxy may reject 'base paths' with special characters (&, ?, etc.)
- our proxy expects to append user request strings to the base path
  - e.g. `http://proxy.example/base_path/additional_path_provided_by_user`
- if our `base_path` contains query strings, then we might get malformed urls
  - e.g. `http://proxy.example/base_path/?queryParam=something/additional_path_provided_by_user`

In addition to `base_path`, our proxy allows `query_string_params` and `request_headers` to be configured per endpoint:
screenshot_20161207_131933

So, it would probably be better to extend our **Proxy Backend** form to contain these fields, while aligning with the API Umbrella schema:

```js
 ...
  "append_query_string": "string",
  ...
```

@Nazarah, would you please provide a wireframe of the **Proxy Backend** form with the additional *append_query_string* field?

Nazarah:
@brylie @santeriv from the discussions above, I have created the following wireframes:

1. The proxy tab in API profile contains an expand/collapse pane that contains the global request setting fields. Since there are lot of fields to consider and it is sort of advanced setting so the pane rests under Advanced Settings label
I have only kept the rate limit option outside, since this is one of most crucial fields.
globalrequestsettings

2. On expanding the pane, you can see the fields along with info icons. We can incorporate the default info texts from Umbrella in their relative placeholders
globalrequestsettings2

3. On expanding different dropmenus, you can see available values for selection from dropmenu
globalrequestsettingsapikeyrequirements


globalrequestsettingsapikeyverifications


globalrequestsettingshttpsrequirements

santeriv:
@brylie @Nazarah thanks for quick reaction and improving this idea. And I also like my proposed change least of these two new alternatives
* changing regex of orginal to loose or
* adding new advanced field is probably doing least harm to existing code

I like the append_query_string approach most.

bajiat:
@Nazarah Can we concentrate on the two first fields only: 
- Append Query String Parameters
- Set Request Headers

The others should implemented as an enhancement later. (Also, note that API key setting is already implement in the Proxy settings above, we did the UI differently than API Umbrella. Let's not just copy their UI.)

Nazarah:
I am working on them.
Previously I could not catch the message that we only need the above mentioned fields.
I thought we would be including all the fields under an expandable form.
I'll upload the modified wireframes asap

brylie:
Thanks @Nazarah :rocket:

Nazarah:
@santeriv @brylie based on the feedback above, I have modified the wireframes.
It is now a single one.
The new fields should appear under Advanced Settings label.
They should be placed right below Disable API Key checkbox and Rate Limit dropdown
globalrequestsettings

brylie:
@santeriv, would you consider modifying the PR based on the discussion above? Specifically:

- modify the [ApiUmbrellaSchema](https://github.com/apinf/platform/blob/develop/proxy_backends/collection/apiUmbrellaSchema.js), adding the following fields:
  - `append_query_string`: String
  - `headers_string`: String
    - double check that this field corresponds with the **Set Request Headers** field in the UI mockups, I may be wrong here
- Add the new fields to the **Proxy Backend** form
  - Fields appear under **Advanced Settings** section

## References
The [API Umbrella Admin API documentation](http://api-umbrella.readthedocs.io/en/latest/admin/api.html) describes the schema structure under the API endpoint POST section (Model Schema).

## Note
We are working for the next couple of months directly on the Apika project. Please open a new pull request against the Digipalvelutehdas / Apika repository, to track your work against our current development efforts.

I will open an issue describing this enhancement, and mention your GH username in the issue.

santeriv:
@brylie this can be closed on my behalf. I don't  have enough time and energy to focus on this. I guess @nnn was interested to solve this.

PR #1990, opened 2017-01-11
Use ENV variable for flexible change docker tag name

shaliko:
The target: Fix issue with Travis config, after merge APIKA code to APINF. Now both projects generate "apika_develop" Docker image.

* Use ENV variable for flexible change docker tag name for APINF and APIKA with TravisCI builds.
* From now we can change $DOCKER_TAG ENV variable in TravisCI settings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,6 @@ services:
   - docker
 
 script:
-  - docker build -t apinf/platform:apika_develop .
+  - docker build -t apinf/platform:$DOCKER_TAG .
   - docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD"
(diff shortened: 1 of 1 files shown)


brylie:
@shaliko can you please also open this PR against Apika? I want to avoid losing these changes.

shaliko:
@brylie I planned did APIKA PR after you will accept this one. git cherry pick + pull request.

mauriciovieira:
It is also possible to use TRAVIS CI variables to have a different tag for each Pull Request, for instance. Is that desirable? For instance, we could `docker push apinf/platform:${TRAVIS_BUILD_NUMBER}` when building a Pull Request.

PR #2079, opened 2017-02-07
Feature/hide org icons

### LINKED ISSUE #2057: Don't show icons on organization profile for contacts
_From @marla-singer on January 10, 2017 12:39_

Now contacts of manager are no required fields. If there are empty then icons in organization header can confused users.

### Examples
**Good**
joxi_screenshot_1484051773112
**Not very good**
joxi_screenshot_1484051849606 2
joxi_screenshot_1484051849606 1
joxi_screenshot_1484051849606

### Suggestion
- Show icons if situation so requires
joxi_screenshot_1484051917775 1
joxi_screenshot_1484051917775

_Copied from original issue: Digipalvelutehdas/APIKA#215_

nnn:
Closes #2057

### DIFF
--- a/organizations/client/profile/header/header.html
+++ b/organizations/client/profile/header/header.html
@@ -32,20 +32,26 @@
           </div>
           <div class="row">
             <ul class="list-inline">
+              {{# if organization.contact.person }}
               <li>
                 <i class="fa fa-user" aria-hidden="true"></i>
                 {{ organization.contact.person }}
               </li>
+              {{/ if }}
+              {{# if organization.contact.phone }}
               <li>
                 <i class="fa fa-phone" aria-hidden="true"></i>
                 {{ organization.contact.phone }}
               </li>
+              {{/ if }}
+              {{# if organization.contact.email }}
               <li>
                 <i class="fa fa-envelope" aria-hidden="true"></i>
                 <a href="mailto:{{ organization.contact.email }}">
                   {{ organization.contact.email }}
                 </a>
               </li>
+              {{/ if }}
             </ul>
           </div>
         </div>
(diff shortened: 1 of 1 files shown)


organizations/client/profile/header/header.html
```diff
@@ -17,11 +17,13 @@ <h1 id="organization-name">
               {{/ if }}
             </h1>
           </div>
-          <div class="row">
-            <p>
-              {{ organization.description }}
-            </p>
-          </div>
+          {{# if organization.contact.person }}
```
marla-singer: @nnn Why do you use this condition for description area?

nnn: Redundant space if description is not available.

marla-singer: So what? If organization have the description but don't have the contact person, the area will be empty instead of has description text.

marla-singer: I think it's okay with redundant space. In future, it will have the cover photo or something else. Don't delete redundant space

nnn: It also hides empty `.row`.
From design perspective:
- without a condition: 
localhost 3000 organizations test 1

- with a condition: 
localhost 3000 organizations test 2

marla-singer: > If organization have the description but don't have the contact person, the area will be empty instead of has description text.

nnn: And the idea is if we hide something, we need to also to hide markup around this element.

marla-singer: So change the condition

nnn: I am so sorry, haven't notice a typo. Now fixed.

Issue #2093, opened 2017-02-09
User have to create totally unique proxy base path


marla-singer:
Now we have the rule that user must create the unique proxy base path. But we need to limit it more. Don't let user create the proxy base path if the first section of this path has is the part of other API

On local, I have three different APIs with the next settings. Bring to your notice that APIs have the different URL. 

![joxi_screenshot_1486625338473 2](https://cloud.githubusercontent.com/assets/20643403/22773384/f673dee6-eeb2-11e6-92e2-cd04b3e7a70b.png)
![joxi_screenshot_1486625338473 3](https://cloud.githubusercontent.com/assets/20643403/22773667/6cccfd42-eeb4-11e6-94cd-2740960b9715.png)
![joxi_screenshot_1486625338473 4](https://cloud.githubusercontent.com/assets/20643403/22773672/7012ed5e-eeb4-11e6-9b11-d6370a9bbcd4.png)

### Found &expected
1.  When I call **` https://nightly.apinf.io:3002/alternative/ `**
it redirects to **` https://music.yandex.ru/genre/alternative/`** that's **correct**
3. When I call **`https://nightly.apinf.io:3002/alternative/artists/ `**
it redirects to **`https://music.yandex.ru/genre/alternative/artists/`** that's **wrong**. 
**Expected:** `http://petstore.swagger.io/test-elastic/`
2. When I call **` https://nightly.apinf.io:3002/alternative/rock/ `**
it redirects to **` https://music.yandex.ru/genre/alternative/rock/`** that's **wrong**. 
**Expected:** `https://learn.javascript.ru/courses/nodejs`

brylie:
We are not committing to resolve this issue right away, but want to get a better understanding of the underlying parts. I have self-assigned and will look for upstream issues or discussions that might give us better understanding.

brylie:
This seems related to the API Umbrella internal **matching order** field:

![screenshot_20170210_122004](https://cloud.githubusercontent.com/assets/17307/22823321/bc960bf2-ef8b-11e6-8f98-6081cfafaad9.png)

The **matching order** field tells API Umbrella in which order to route requests. API Backends with lower matching order value will be routed before higher level routes, when their frontend prefix matches, even when a higher level route may be a more accurate route.

brylie:
In the spirit of this issue title, we might consider 'solving' this challenge by allowing multiple, explicit frontent prefixes per Proxy Backend. This aligns with the design of API Umbrella:

[...]

PR #2099, opened 2017-02-10
Bugfix/deleting api

### LINKED ISSUE #2032: Consider whether to remove analytics data from API Umbrella when deleting ProxyBackend in Apinf
When deleting a ProxyBackend from Apinf, the backend is also deleted from API Umbrella. However, API Umbrella does not delete analytics data related to the backend, perhaps by design. 

# Discussion
Consider the pros and cons of deleting related analytics data from API Umbrella. E.g. might the data be useful to a system administrator? Determine why API Umbrella does not delete this data. Is it by design, or perhaps an oversight?

Decide whether we align with the design of API Umbrella. If not,
- can/should the problem be solved upstream?
- is it necessary to produce a solution directly in our project?

# Original post
_From @Nazarah on January 30, 2017 13:37_

## Reproduction Steps:

1. Sign in to https://apika.digipalvelutehdas.fi/
2. Add an API with proper URL
3. Go to proxy tab in API profile and provide Proxy Base Path and API Base path.
4. Make a call to the API.
5. Go to Dashboard and select the added API from the dropmenu. Remember the no. of request.
6. Go to API profile and delete the API.
7. After confirming the deletion, add the same API. (includng the proxy base path and API base path)
8. go to Dashboard and select the API from dropmenu.

# Result:
Dashboard shows the number of requests made before the API was deleted and readded.

apiproxydeletion

## Expected Result
API details should also be deleted from proxy on deleting an API.
So that when the same API is added again, Dashboard should show fresh history (of 0 or requests made after added, etc.) than the previous history of requests.

## Environment
Chrome, Windows 8

_Copied from original issue: Digipalvelutehdas/APIKA#324_

marla-singer:
Closes #2032 

The main idea is to search related documents in ElasticSearch and delete found documents using the required fields as `_index, _id, _type`. There can be a lot of found documents then I use `bulk` function which can perform many operations at one time. 
Links:
[Delete document](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/api-reference-2-4.html#api-delete-2-4)
[Bulk](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/api-reference-2-4.html#api-bulk-2-4)

Now I find the documents via `frontend_prefix` that it isn't safe but only available method right now. And I faced with the really long API deleting.

### DIFF
--- a/elasticsearch/server/methods.js
+++ b/elasticsearch/server/methods.js
@@ -133,7 +133,6 @@ Meteor.methods({
                   {
                     wildcard: {
                       request_path: {
-                        // Add '*' to partially match the url
                         value: `${proxyData.frontendPrefix}`,
                       },
                     },
--- a/elasticsearch/server/methods.js
+++ b/elasticsearch/server/methods.js
@@ -112,4 +112,80 @@ Meteor.methods({
     // Now it returns some placeholder
     return { size: 1000 };
   },
+  deleteAnalyticsInformation (proxyBackendId) {
+    // Make sure proxyBackendId is s String
+    check(proxyBackendId, String);
+
+    // Get the frontend prefix of Proxy backend and ES URL
+    const proxyData = Meteor.call('getProxyData', proxyBackendId);
+
+    // Create parameters object to find API via frontend prefix
+    // TODO: Redo it using api_backend_id and do query for eMQTT protocol
+    const params = {
+      size: 50000,
+      body: {
+        query: {
+          filtered: {
+            query: {
+              bool: {
+                should: [
+                  {
+                    wildcard: {
+                      request_path: {
+                        // Add '*' to partially match the url
+                        value: `${proxyData.frontendPrefix}`,
+                      },
+                    },
+                  },
+                ],
+              },
+            },
+          },
+        },
+      },
+    };
+
+    // Init ES client
+    const esClient = new ElasticSearch.Client({ host: proxyData.elasticSearchUrl });
+
+    // Search related documents in ElasticSearch storage
+    const searchResult = esClient.search(params);
+
+    // Return the Promise result
+    searchResult
+      .then(response => {
+        // Get the hits data if searching was success
+        const hits = response.hits.hits;
+
+        // If ES doesn't have data for this frontend prefix then nothing to delete
+        if (hits.length < 1) {
+          // Exit from function
+          return hits;
+        }
+
+        // Create parameters object to perform many delete operations
+        const bulkBody = _.map(hits, (hitDocument) => {
+          return {
+            delete: {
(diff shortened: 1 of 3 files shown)


marla-singer:
@brylie Ready for review

elasticsearch/server/methods.js
```diff
@@ -112,4 +113,79 @@ Meteor.methods({
...
+      body: {
+        query: {
+          filtered: {
+            query: {
+              bool: {
+                should: [
+                  {
+                    wildcard: {
+                      request_path: {
+                        value: `${proxyData.frontendPrefix}`,
```
brylie: Basically, this PR and your recent observations are leading us towards 'namespacing'. In immediate effect, once someone creates a frontend prefix, say `/hsl/`, and adds Proxy backends that share a portion of the frontend prefix `/hsl/transit` `/hsl/maps`, this solution may delete too much data.

@bajiat and I would like to have a broader discussion with you, @shaliko, and other developers to figure out a way forward.

marla-singer: @brylie Actually, I don't use the `*` symbol in request so result must be exact match

marla-singer: I'm confused about other things. For instance, delete API operation is too long. And I don't know how to catch the case when deleting throw error and so on

brylie: This is *somewhat* related to issue #2093, in that API Umbrella uses an internal **matching order** field to handle path collisions when routing proxy requests:

screenshot_20170210_122004

marla-singer: if you have three frontend prefix as `/hsl/`,/hsl/maps/ and /hsl/parkandride/` and you want to delete `/hsl/` then It will delete only documents with `/hsl/` parameter. Check it

brylie:
Adding 'blocked' label to indicate there are higher level concerns here.

marla-singer:
@brylie Compare: 
1. My search request to ES:
https://github.com/apinf/platform/blob/792331d680d854403521591229a891cb7650a525/elasticsearch/server/methods.js#L136

2. Search request to ES for Dashboard:
https://github.com/apinf/platform/blob/792331d680d854403521591229a891cb7650a525/elasticsearch/server/methods.js#L64

In the second case, we add `*` in the and of frontend prefix

My request returns the document with `/hsl/`.
The second one returns the documents with `/hsl/`, `/hsl/maps/` and `/hsl/parkandride/`

brylie:
Right, we need to clarify why and how we use wildcard queries (with or withough wildcard characters such as '*').

brylie:
Closing this PR, since it is inactive. We can continue work in a new PR, if necessary.

Issue #2045, opened 2017-02-06
Add eslint check to travis build


bajiat:
_From @mauriciovieira on January 15, 2017 13:50_

Currently, there is no static code check in the [continuous integration](https://travis-ci.org/Digipalvelutehdas/APIKA). It is desirable to add a eslint check during pull requests because that will help avoiding the introduction of new bugs.

# Definition of done
* [x] Add eslint check to continuous integration (#237)
* [ ] Make eslint to check all directories
  * [x] about (#237)
  * [x] api_catalog (#237)
  * [x] api_keys (#237)
  * [x] apis
    * [x] apis/client (#282)
    * [x] apis/collection (#316)
    * [x] apis/flags (#317)
    * [x] apis/logo (#318)
    * [x] apis/server (#252)
  * [x] apis_authorization (#281)
  * [x] backlog (#341)
  * [x] bookmarks (#350)
  * [x] branding (#351)
  * [x] core (#352)
  * [x] dashboard (#353)
  * [x] docker
  * [x] documentation (#354)
  * [x] elasticsearch (#355)
  * [x] feedback (#2070)
  * [x] feedback_votes (#2070)
  * [x] home (#2082)
  * [x] metadata (#2083)
  * [x] monitoring (#2084)
  * [x] node_modules
  * [x] organization_apis (#260)
  * [x] organizations (#260)
  * [x] proxies (#2103)
  * [x] proxy_backends (#2104)
  * [x] public
  * [x] ratings (#2105)
  * [x] scripts
  * [x] search (#2106)
  * [x] settings (#2108)
  * [x] setup (#2109)
  * [x] status (#2110)
  * [x] users (#2111)

_Copied from original issue: Digipalvelutehdas/APIKA#236_

bajiat:
_From @mauriciovieira on January 15, 2017 13:50_

@bajiat I am working on this. I talked to @brylie Friday and we think it's a good idea.

bajiat:
_From @mauriciovieira on January 15, 2017 15:10_

I am doing this 'locally', and it seems we'll have a long work to do. According to https://travis-ci.org/mauriciovieira/platform/builds/192108795, there are _1656 problems (1358 errors, 298 warnings)_  right after the latest Jan 12th commit (https://github.com/mauriciovieira/platform/commit/0c6b28f7d00ee1eb9328e0341fa829cdd095e827) 

What I propose is to not add all directories in a single step, but one directory at a time, and fix the lints progressively.

[...]

PR #2104, opened 2017-02-11
Lint proxy_backends directory

mauriciovieira:
This PR is part of #2045 endeavor.


mauriciovieira:
@apinf/developers  please review

proxy_backends/client/form/autoform.js
```diff
@@ -77,16 +78,18 @@ AutoForm.hooks({
...
-                      (error) => {
-                        if (error) {
+                      (publishApiBackendOnApiUmbrellaError) => {
+                        if (publishApiBackendOnApiUmbrellaError) {
                           Meteor.throw(500, error);
                         } else {
                           // Insert the Proxy Backend document, asynchronous
                           form.result(proxyBackend);
                         }
+                        return true;
```
marla-singer: `AutoForm.hooks([])` can't return `true` value. [Docs](https://github.com/aldeed/meteor-autoform#callbackshooks)
Should be `return proxyBackend` but I'm not pretty sure that it will be work. I'll test it.
However is returning necessary?

proxy_backends/client/form/form.js
```diff
@@ -1,21 +1,16 @@
-// Meteor package imports
-import { ReactiveVar } from 'meteor/reactive-var';
-import { Template } from 'meteor/templating';
 import { Counts } from 'meteor/tmeasday:publish-counts';
 import { Modal } from 'meteor/peppelg:bootstrap-3-modal';
+import { ReactiveVar } from 'meteor/reactive-var';
+import { Template } from 'meteor/templating';
+import { URI } from 'meteor/olragon:uri-js';
```
marla-singer: @mauriciovieira Delete this import because there are browser error when I'm opening the API page

proxy_backends/client/form/form.js
```diff
@@ -1,21 +1,16 @@
...
 
-// Apinf import
-import { ProxyBackends } from '/proxy_backends/collection';
-import { Proxies } from '/proxies/collection';
+import ProxyBackends from '/proxy_backends/collection';
 import deleteProxyBackend from '/proxy_backends/client/methods/delete_proxy_backend';
+import { Proxies } from '/proxies/collection';
 
-// NPM import
 import 'urijs';
```
marla-singer: Use `import URI from 'urijs';`

proxy_backends/server/methods.js
```diff
@@ -1,9 +1,13 @@
+import { check } from 'meteor/check';
 import { Meteor } from 'meteor/meteor';
 
-import { ProxyBackends } from '/proxy_backends/collection';
+import ProxyBackends from '/proxy_backends/collection';
 
 Meteor.methods({
   deleteProxyBackend (proxyBackend, deleteFromMongoDB = true) {
+    // Make sure proxyBackend is a String
+    check(proxyBackend, String);
```
marla-singer: @mauriciovieira Change `String` type to `Object`

proxy_backends/server/methods.js
```diff
@@ -34,8 +38,12 @@ Meteor.methods({
     );
   },
   uniqueFrontendPrefix (proxyBackend) {
+    // Make sure proxyBackend is a String
+    check(proxyBackend, String);
```
marla-singer: Change String type to Object

proxy_backends/client/form/autoform.js
mauriciovieira: It requires returning a value.

proxy_backends/client/form/form.js
mauriciovieira: OK!

proxy_backends/client/form/form.js
mauriciovieira: :1st_place_medal:

proxy_backends/server/methods.js
mauriciovieira: OK!

proxy_backends/server/methods.js
mauriciovieira: OK!

proxy_backends/client/form/autoform.js
```diff
@@ -77,16 +78,18 @@ AutoForm.hooks({
...
-                      (error) => {
-                        if (error) {
+                      (publishApiBackendOnApiUmbrellaError) => {
+                        if (publishApiBackendOnApiUmbrellaError) {
                           Meteor.throw(500, error);
                         } else {
                           // Insert the Proxy Backend document, asynchronous
                           form.result(proxyBackend);
                         }
+                        return proxyBackend;
```
marla-singer: Let's delete all `return proxyBackend` and `return true` from AutoForm. The form  doesn't work as expected and I can understand quickly why it is so

mauriciovieira: OK. The problem is that this code is too complex and difficult to understand.

mauriciovieira: Previously it was returning nothing...

proxy_backends/client/form/autoform.js
```diff
@@ -204,6 +215,7 @@ AutoForm.hooks({
               });
           }
         }
+        return nothing;
```
marla-singer: @mauriciovieira Works well. 
@brylie Is it okay with the semantic name for `nothing` variable?

brylie: Well, we sometimes just use `return undefined;` which means the same thing pretty much.

brylie: Also, a quick comment about **why** we are returning 'nothing' would help other developers understand this code better.

mauriciovieira: @marla-singer @brylie OK, I think it's better to return undefined. And I will add a comment.

PR #2121, opened 2017-02-14
Feature/swagger url

### LINKED ISSUE #1993: Allow swagger URL and not just swagger file
# Context
Currently, Apinf only allows API Managers to upload Swagger file to be rendered in the documentation browser.

# User story
> As an API Manager
I would like for Apinf to render a Swagger file hosted on a third party server
so I can develop and host my swagger file elsewhere and not have to manually upload changes

# Feature request
Allow API Managers to specify a URL pointing to a Swagger file somewhere on the Internet. When the URL is provided, the Swagger file should be rendered in the Swagger UI.

# Definition of done
* [x] User can choose whether they want to upload a file or add a URL for showing a Swagger document in the Swagger UI
* [x] If the user adds a URL, it needs to be stored for the purpose of showing the file in the Swagger UI
* [x] When the user has made a selection, the other option should be disabled

Note: this solution should not affect the possibility of adding external documentation links

# Original request
This is so that when we update our swagger file on our site, we don't have to remember to upload it to apinf.io

# Wireframe
On opening the Manage dialog, user will now see Documentation link option and Documentation upload function appearing as radio buttons.
Selecting one option would disable the other option and vice versa. This is mentioned with instruction texts.
A feedback using S-alert can be shown mentioning which option has been enabled and which one disabled.

apidocumentation

apidocumentation2

### LINKED ISSUE #2013: API profile: popover text doesn't have translation
Verified on nightly.apinf.io (0.39.0 version)
API profile -> Documantation tab -> Manage button

joxi_screenshot_1486123942225

nnn:
Closes #1993 
Closes #2013


documentation/client/manage/manage.html
```diff
@@ -21,6 +21,15 @@ <h1 class="modal-title">
               {{> showHelp 'uploadApiDocumentation' }}
               {{_ "import_ApiDocumentation_Title" }}
             </legend>
+            {{> afQuickField
+              name="documentationMethod"
+              options="allowed"
```
brylie: Keep in mind, we will need to internationalize these options values. For internal purposes, the keys can be (lowercase) strings, but the values will be localized into the end-user language.

jawidahmadi:
Still not completed. 
Some functionalities have to be enhanced, for example, choose file button.

matleppa:
When Manage button in API documention template is pressed, the API documentation modal opens.
As default in Documentation field there is value File. However, the Documentation url is visible instead of button Choose file.
 When method is changed to URL and then back to file, the fields are synchronised.
In case the initial value of field Method were URL, the problem is avoided.

nnn:
Having reactivity problem here. When switching between URL and file viewer doesn't update it simultaneously. Works only if I refresh the page.
I think that we need to change logic over here: https://github.com/apinf/platform/blob/develop/documentation/client/swaggerUI/swaggerUi.js
to validate swagger onRender. @apinf/developers will appreciate your help.

apis/collection/schema.js
```diff
@@ -27,6 +27,25 @@ Apis.schema = new SimpleSchema({
     type: String,
     optional: true,
   },
+  documentationUrl: {
+    type: String,
+    optional: true,
+    regEx: SimpleSchema.RegEx.Url,
+  },
+  documentationMethod: { // any better name?
```
marla-singer: @nnn What about `documentationType`?

apis/collection/schema.js
```diff
@@ -27,6 +27,25 @@ Apis.schema = new SimpleSchema({
     type: String,
     optional: true,
   },
+  documentationUrl: {
```
marla-singer: Suggest to rename it as well. For example, `swaggerDocumentationUrl`. But it's just suggestion

marla-singer:
@nnn  First at all you didn't realise point as "When the user has made a selection, the other option should be disabled" from Definition of done.

nnn:
@marla-singer that was decided during discussion with @brylie and @bajiat as well as dropdown-select instead of two different fields.

marla-singer:
@nnn Yes, but don't show drop-down menu when user already uploaded document or save the URL.

marla-singer:
https://github.com/apinf/platform/pull/2127
I prepared some solution. It allows resolving problems when user:
1. Changes File to URL
2. Changes URL1 to URL2

But when you delete URL, click "Save" and don't change `documentaMethod` when the swagger UI will be empty. Because type File wasn't be selected

nnn:
@marla-singer thanks for help! Could you please resolve conflicts over here #2127 and I will merge this code here?

apis/collection/schema.js
brylie: Although, we have tried to remove the word `swagger` from our code/UI for at least a couple of reasons:
- swagger is now called the Open API Specification
- swagger is not really a word

marla-singer: @brylie I see your point. But it looks explicit why we have `documentationUrl` and `otherDocumentationUrl`

marla-singer:
@nnn Done

nnn:
thanks @marla-singer for pair-programming!

@jawidahmadi @matleppa please review.

Try to upload documentation: 
e.g. https://api.apis.guru/v2/specs/apitransformer.com/1.0/swagger.json
Try to add as URL:
e.g. https://api.apis.guru/v2/specs/amadeus.com/1.2/swagger.json 
Try to switch from one to another.

bajiat:
@nnn Suggestion for help text:

You can add a Swagger document by uploading a file or by giving a link to the remote file. 
Note: Try out methods can only be controlled, if the file is uploaded.

documentation/client/manage/autoform.js
```diff
@@ -1,10 +1,19 @@
...
 import { TAPi18n } from 'meteor/tap:i18n';
 import { sAlert } from 'meteor/juliancwirko:s-alert';
+import Apis from '/apis/collection';
 
 AutoForm.hooks({
   apiDocumentationForm: {
     onSuccess () {
+      const api = Apis.findOne();
+
+      // Checks if user selected option but submitted without content
```
brylie: @nnn and @marla-singer, what is a different way to solve this?

documentation/client/manage/autoform.js
```diff
@@ -1,10 +1,19 @@
...
+      if (api.documentationType === 'url' && !api.documentationUrl) {
+        Apis.update(api._id, { $set: { documentationType: 'file' } });
```
brylie: I.e. this is clever code, but it is a bit 'sneaky'. How else might this problem be prevented, with explicit user feedback?

documentation/client/view/documentation.js
```diff
@@ -29,34 +26,15 @@ Template.apiDocumentation.onRendered(() => {
...
-    return documentationFileUrl;
+  apiDoc () {
+    return this.api.documentation();
   },
-  documentationLink () {
-    // get documentation link
-    return this.api.documentation_link;
+  otherDocumentationUrl () {
+    // Get other documentation link
+    return this.api.otherDocumentationUrl;
```
brylie: This helper is redundant, since the `api` document should already be bound to the template data context.

apis/collection/schema.js
brylie: Right, it is ambiguous. Lets see what else we can come up with.

apis/collection/helpers.js
```diff
@@ -175,4 +177,37 @@ Apis.helpers({
       Apis.update(this._id, { $unset: { bookmarkCount: '' } });
     }
   },
+  documentation () {
+    // Get documentation method (URL of File)
+    const documentationType = this.documentationType;
+
+    // Get uploaded documentation file ID
+    const documentationFileId = this.documentationFileId;
```
brylie: What happens if `documentationFileId` is `undefined`?

documentation/client/swaggerUI/swaggerUiContent/swaggerUiContent.js
```diff
@@ -104,14 +104,22 @@ Template.swaggerUiContent.onCreated(function () {
 
   // Subscribe to api collection
   instance.autorun(() => {
+    const apiId = instance.data.api._id;
     // Get relevant api collection
-    instance.subscribe('apiBackend', instance.data.api._id);
+    // instance.subscribe('apiBackend', apiId);
```
brylie: Remember to clean up commented code.

documentation/client/view/documentation.js
```diff
@@ -29,34 +26,15 @@ Template.apiDocumentation.onRendered(() => {
...
-      // Build documentation files base url
-      const meteorAbsoluteUrl = Meteor.absoluteUrl().slice(0, -1);
-      const documentationFilesBaseURL = meteorAbsoluteUrl + DocumentationFiles.baseURL;
-
-      // Get documentation file URL
-      documentationFileUrl = `${documentationFilesBaseURL}/id/${documentationFileId}`;
-    }
-    return documentationFileUrl;
+  apiDoc () {
+    return this.api.documentation();
```
brylie: This helper might be made redundant if we bind the `api` document to the `apiDocumentation` template. You can change the swagger UI template include `apiDoc` attribute to `api.documentation`, I am fairly certain.

documentation/client/manage/autoform.js
bajiat: Could we solve this by making the documentation URL field required, when documentation type is URL? I would like to get this fixed and to be able to make a release today.

marla-singer: We discussed it with @nnn. It can be a case when user deletes URL and saves the result. If documentation URL field is required then user will not be available to delete URL

documentation/client/view/documentation.js
marla-singer: Don't understand this comment about `otherDocumentationUrl`

documentation/client/manage/autoform.js
brylie: I think we can find a middle ground solution here, such as:

1. use conditional validation to ensure `documentationUrl` or `documentationFileId` fields are provided, based on the field selection
  - show user explicit validation message, indicating which field is required
  - allow user to back out of submission with cancel
2. provide a 'remove documentation' button that actually deletes the ApiDocumentation collection document

In effect, we want to 
- prevent invalid data from being saved and 
- give the end-user an explicit understanding of what is happening.

brylie: The current solution works, but is concerning because it
- does not validate required form fields
- performs an action 'behind the scenes' - changing the value of the `documentationType` field without notifying the user
- allows an invalid document to be saved into the ApiDocumentation collection
  - e.g. `documentationType` is url while `documentationUrl` is undefined
  - e.g. `documentationType` is file while `documentationFileId` is undefined

marla-singer: 1. We didn't have required fields in the form before. 
2. Then I need clarification about behaviours what is expected

### Scenario
1. I'm a user. I downloaded the file, saved. Then I opened the form, **didn't** delete the file, changed `documentationType` to URL, added the URL, saved. Then I opened the form again, deleted URL, **didn't change type** and try to save.

What behaviour is expected?

2. **The reverse situation**
I'm a user. I added the URL, saved. Then I opened the form, **didn't** delete the URL, changed `documentationType` to file, added the file, saved. Then I opened the form again, deleted file, **didn't change type** and try to save.

What behaviour is expected?

documentation/client/view/documentation.js
marla-singer: Done

documentation/client/swaggerUI/swaggerUiContent/swaggerUiContent.js
marla-singer: Done

apis/collection/helpers.js
marla-singer: `DocumentationFile` collection returns undefined. But I added the condition

marla-singer:
@brylie Did requested changes

documentation/client/view/documentation.js
brylie: Basically, it is about template data binding. If we bind the `api` document to the template, we can access properties directly from the markup. E.g.

```html
{{# if api.otherDocumentationUrl }}
  // show api other documentation url
  {{ api.otherDocumentationUrl
{{/ if }}
```

brylie: In the above example, we don't need a template helper, since the data is already bound to the template.

documentation/client/manage/autoform.js
brylie: In scenario 1, inform the user that the URL field is required when documentation type is set to URL.

brylie: In scenario 2, inform the user that the File is required when documentation type is set to File.

brylie: The proposed solution involves conditional validation, making the field required based on the value of another field.

marla-singer: OK, see your point.

apis/collection/schema.js
```diff
@@ -40,6 +40,23 @@ Apis.schema = new SimpleSchema({
     type: String,
     optional: true,
     regEx: SimpleSchema.RegEx.Url,
+    custom () {
+      let validation;
+
+      // Make sure type is URL and value is empty
+      if (this.field('documentationType').value === 'url' && !this.value) {
```
marla-singer: @brylie Nope, this solution doesn't work. Either it will be a really complex solution for validation or it needs to change all logic for inserting information about URLand file.

### Case
1. Add the URL
2. Save
3. Open again
4. Delete URL and click on Save
5. It shows "field is required" (as you suggested)
6. User's confusing - how can I delete URL? Ok, try to change type to file. Oh, no. This field is required too! 

Do you see my idea?

brylie: They just need to change the `documentationType` field to 'File'.

brylie: They can alternatively hit 'Cancel' or 'Remove documentation', if they need.

brylie: > It shows "field is required" (as you suggested)

Note also, we can change the message to be more user-friendly:

"URL field is required when documentation type is set to 'URL'"

brylie: In any case, we need conditional validation, so users cannot submit invalid data.

apis/collection/schema.js
```diff
@@ -41,6 +32,32 @@ Apis.schema = new SimpleSchema({
...
+    type: String,
+    optional: true,
+    regEx: SimpleSchema.RegEx.Url,
+    custom () {
+      let validation;
+
+      // Make sure type is URL and value is empty
+      if (this.field('documentationType').value === 'url' && !this.value) {
+        // Set this field as required
+        validation = 'required';
```
brylie: We can customize the validation message by using a different string than `'required'`.

marla-singer: The problem is not in validation message

brylie: In the case you describe, we simply need a **Remove Documentation** button that deletes the ApiDocumentation document.

apis/collection/schema.js
```diff
@@ -23,10 +23,46 @@ Apis.schema = new SimpleSchema({
...
   },
+  documentationUrl: {
+    type: String,
+    optional: true,
+    regEx: SimpleSchema.RegEx.Url,
+    custom () {
+      let validation;
+
+      // Make sure type is URL and value is empty
+      if (this.field('documentationType').value === 'url' && !this.value) {
```
brylie: Lets move `this.field('documentationType').value` into a separate step:

```js
const documentationType = this.field('documentationType').value;
```

apis/collection/schema.js
```diff
@@ -41,6 +32,32 @@ Apis.schema = new SimpleSchema({
...
+        // Set this field as required
+        validation = 'required';
+      }
+
+      // Make sure type is URL and value is empty
+      if (this.field('documentationType').value === 'file' && !this.value) {
```
brylie: We probably don't need the inverse case, where we 'unset' validation. Just return the validation 'required' string when it is required. Otherwise, don't return anything.

apis/collection/schema.js
brylie: Also, we can just `return 'validationType';` here, rather than checking the inverse condition below.

brylie:
I just realized, with so many documentation-related fields in the Apis schema, we probably need to migrate those fields to an `ApiDocs` collection. To some extent, my underlying assumption has been that we were working within an `ApiDocs` collection, rather than packing more fields into the `Apis` collection schema.

bajiat:
@nnn Will you be able to finalize this? This involves creating a new collection, adding the fields and removing from Apis collection. Also, you will need to write a migration.

brylie:
@nnn, be careful how you do conflict resolution. I think you have reverted some significant changes here, specifically the LICENSE text.

nnn:
@brylie @jawidahmadi @matleppa please review.

api_docs/client/codegenerator/codegenerator.js
```diff
@@ -7,7 +7,7 @@ import { Template } from 'meteor/templating';
 
 import _ from 'lodash';
 
-import DocumentationFiles from '/documentation/collection';
+import DocumentationFiles from '/doc_files/collection';
```
brylie: Since the `doc_files` is related to the 'Api Documentation' module, we can make an exception and store `doc_files` inside of the `api_documentation` directory. E.g.

- `/api_documentation/files/`
- `/api_docs/files/`

api_docs/client/manage/manage.js
```diff
@@ -101,12 +122,25 @@ Template.manageApiDocumentationModal.helpers({
...
+    return ApiDocs;
+  },
+  formType () {
+    const instance = Template.instance();
+    if (instance.data.apiDoc) {
+      return 'update';
+    }
+    return 'insert';
+  },
+  apiId () {
```
brylie: Where is this `apiId` helper used?

api_docs/collection/permissions.js
```diff
@@ -0,0 +1,28 @@
+import ApiDocs from './';
+
+ApiDocs.allow({
+  insert (userId) {
```
brylie: We might need to check for permissions within these allow rules. E.g. 

```js
if ( api.currentUserCanEdit() ) {
  return true;
}
```

brylie: The above snippet implies the `api` document is available, which means you need to fetch it within the allow rule function. For reference, check the allow rules for the `api_media` module.

api_docs/collection/schema.js
```diff
@@ -0,0 +1,44 @@
...
+    ],
+    defaultValue: 'url',
+  },
+  fileId: {
+    type: String,
+    optional: true,
+  },
+  remoteFileUrl: {
+    type: String,
+    optional: true,
```
brylie: Make sure to open an 'enhancement' issue for conditional validation of this field.

api_docs/collection/schema.js
```diff
@@ -0,0 +1,44 @@
...
+  type: {
+    type: String,
+    optional: false,
+    allowedValues: [
+      'url',
+      'file',
+    ],
+    defaultValue: 'url',
+  },
+  fileId: {
```
brylie: Make sure to open an 'enhancement' issue for conditional validation of this field.

apis/collection/helpers.js
```diff
@@ -175,4 +178,49 @@ Apis.helpers({
       Apis.update(this._id, { $unset: { bookmarkCount: '' } });
     }
   },
+  documentation () {
+    // Get API ID
+    const apiId = this._id;
+
+    const apiDocs = ApiDocs.findOne({ apiId });
```
brylie: It is possible that `ApiDocs.findOne()` might return `undefined`. In that case, trying to access properties of undefined, such as `type` or `fileId`, will cause errors.

core/migrations/server/3-move-api-docs.js
```diff
@@ -0,0 +1,24 @@
...
+      // New apiDoc object
+      const apiDoc = {
+        apiId: apis._id,
+        fileId: apis.documentationFileId,
+        otherUrl: apis.documentation_link,
+        submit_methods: apis.submit_methods,
+      };
+
+      // Insert migrated API docs
+      ApiDocs.insert(apiDoc);
```
brylie: This `ApiDocs` document should only be created if the Api

- has a `documentationFileId` **and/or**
- has a `documentation_link`

brylie: Basically, we don't want to create `ApiDocs` documents for Apis with no documentation.

PR #2164, opened 2017-02-21
Bugfix/proxy backend form

### LINKED ISSUE #2011: API profile: Proxy backend tab doesn't have completed translation
joxi_screenshot_1486123683725

joxi_screenshot_1486123554801

marla-singer:
Closes #2011 
The main idea is to delete the first option `(Select proxy)` from the proxy list on Proxy tab.
joxi_screenshot_1486123683725


marla-singer:
Have a problem with translate label in proxy backend form.
**Before add label for Rate limits**
joxi_screenshot_1487766712641

**After add label for Rate limits**
joxi_screenshot_1487766577122

Don't have any idea how to fix the label duplication. Need help/advice @apinf/developers

brylie:
@marla-singer can you add a label to the `rate_limit` object as well?

screenshot_20170223_102050

core/lib/i18n/en.i18n.json
```diff
@@ -722,6 +723,9 @@
                           "unlimited": "Unlimited requests"
                       }
                   },
+                  "rate_limits": {
+                      "label": "Rate limits"
+                  },
                   "rate_limit": {
```
brylie: Can you add a label here as well? Would that fix the 'label duplication' problem?

marla-singer: @brylie Nope, it doesn't help

brylie: OK, well this is a somewhat minor issue, so probably isn't worth fighting with right now. Lets move on and complete this task, opening an enhancement task to come back to the issue later.

marla-singer: Ok, thanks :+1:

brylie: Jeppers. Just remember to open the enhancement task, so we remember the problem.

marla-singer: Done #2199

marla-singer:
@brylie Ready for review

proxy_backends/client/form/autoform.js
```diff
@@ -148,6 +125,11 @@ AutoForm.hooks({
             return false;
           }
 
+          // Get API id
+          const apiId = currentProxyBackend.apiId;
+          // Get proxy backend id
+          const proxyBackendFromMongo = ProxyBackends.findOne({ apiId });
```
brylie: How about calling this `previousProxyBackend`, since it refers to a ProxyBackend document.

- [Use the same vocabulary for the same type of variable](https://github.com/ryanmcdermott/clean-code-javascript#use-the-same-vocabulary-for-the-same-type-of-variable)

Also, the word `mongo` is redundant, since all of our documents come from Mongo.
- [Don't add unneeded context](https://github.com/ryanmcdermott/clean-code-javascript#dont-add-unneeded-context)

proxy_backends/client/form/form.js
```diff
@@ -5,61 +5,46 @@ import { Template } from 'meteor/templating';
...
 import URI from 'urijs';
 
 Template.proxyBackend.onCreated(() => {
   const instance = Template.instance();
-  // Set the proxy id empty on default
-  // instance.data.proxyId = new ReactiveVar('');
 
   instance.getProxyId = () => {
+    // Set the proxy id empty on default
+    instance.data.proxyId = new ReactiveVar('');
```
brylie: Why are we binding this reactive variable to the `instance.data` context? What would be different if we just bound it directly to the `instance`?

proxy_backends/client/form/form.js
```diff
@@ -5,61 +5,46 @@ import { Template } from 'meteor/templating';
...
     if (proxyCount === 1) {
       // Set these id as current
       currentProxyId = Proxies.findOne()._id;
-    } else if (proxyBackendExists) {
+    } else {
       // Second case: multi proxy
 
-      // If ProxyBackend configuration exists then it has a proxy id
+      // Get Proxy Backend config
+      const proxyBackendExists = instance.data.proxyBackend;
```
brylie: Consider just calling this `proxyBackend`. E.g. a noun `proxyBackend` is semantic, since this variable references an object.

If you prefer to use the *verb phrase* "proxy backend exists", then the variable should reference a Boolean.

proxy_backends/client/form/form.js
```diff
@@ -5,61 +5,46 @@ import { Template } from 'meteor/templating';
...
 
-      // Update & save the current proxy id
-      currentProxyId = proxyBackendExists.proxyId;
+      // If ProxyBackend configuration exists then it has a proxy id
+      if (proxyBackendExists) {
```
brylie: Using `if (proxyBackend)` here is alright. I.e. we don't need the verb 'exists'.

proxy_backends/client/form/form.js
```diff
@@ -5,61 +5,46 @@ import { Template } from 'meteor/templating';
...
+        const proxies = Proxies.find({}, { sort: { name: 1 } }).fetch();
 
-      // Otherwise empty string is current id (after deleting configuration)
+        currentProxyId = proxies[0]._id;
+      }
     }
 
-    // Set the current proxy id
-    return instance.data.proxyId.set(currentProxyId);
+    instance.data.proxyId.set(currentProxyId);
```
brylie: Lets attach the `proxyId` reactive variable directly to the instance.

proxy_backends/client/form/form.js
```diff
@@ -146,12 +131,15 @@ Template.proxyBackend.helpers({
   proxyHost () {
     // Get a reference of Template instance
     const instance = Template.instance();
-    // Get & save current proxy id
-    instance.getProxyId();
-    const currentProxyId = instance.data.proxyId.get();
+
+    // Get proxy ID from template instance
+    const savedProxyId = Template.instance().data.proxyId;
```
brylie: Does this code work as expected? `instance.data.proxyId` refers to a reactive variable above.

proxy_backends/client/form/form.js
```diff
@@ -228,24 +207,15 @@ Template.proxyBackend.events({
...
-      } else {
-        Modal.show('changeSelectedProxy', {
-          proxyBackendEvent: event,
-          proxyBackend: templateInstance,
-          selectedItem,
-        });
-      }
+      Modal.show('changeSelectedProxy', {
+        proxyBackendEvent: event,
+        proxyBackend: templateInstance,
```
brylie: Why are we calling the `templateInstance` "proxy backend"? That does not seem very semantic.

proxy_backends/client/methods/convert_to_apiUmbrella_object.js
```diff
@@ -2,23 +2,27 @@
 // object doesn't have standard fields to creating & publishing in api Umbrella
 // This function is conversion
 
-/* eslint max-len: ["error", 120] */
 export default function convertToApiUmbrellaObject (noStandardObject) {
```
brylie: What would be a simpler, more meaningful name for `noStandardObject`?

brylie: Moreover, why are we changing the structure of this object so drastically? Our schema is supposed to align with the API Umbrella schema, so we can avoid mapping functions.

proxy_backends/client/methods/convert_to_apiUmbrella_object.js
```diff
@@ -2,23 +2,27 @@
...
-  apiUmbrellaObject.apiUmbrella.servers = noStandardObject['apiUmbrella.servers'];
-  apiUmbrellaObject.apiUmbrella.settings = {};
-  apiUmbrellaObject.apiUmbrella.settings.disable_api_key = noStandardObject['apiUmbrella.settings.disable_api_key'];
-  apiUmbrellaObject.apiUmbrella.settings.rate_limit_mode = noStandardObject['apiUmbrella.settings.rate_limit_mode'];
-  apiUmbrellaObject.apiUmbrella.settings.rate_limits = noStandardObject['apiUmbrella.settings.rate_limits'];
-  apiUmbrellaObject.apiUmbrella.url_matches = noStandardObject['apiUmbrella.url_matches'];
+  const apiUmbrellaObject = {
+    apiId: noStandardObject.apiId,
+    proxyId: noStandardObject.proxyId,
+    apiUmbrella: {
```
brylie: Would it be shorter to say `apiUmbrella: noStandardObject['apiUmbrella']`?

brylie:
Mostly the suggestions/questions are related to code semantics. Some of the issues make it hard to understand the code and might allow for possible bugs.

proxy_backends/client/methods/convert_to_apiUmbrella_object.js
marla-singer: Unfortunately, noStandardObject doesn't have `apiUmbrella` field

marla-singer:
Thanks for fresh view. Try to improve code to the extent that I can

proxy_backends/client/form/form.js
marla-singer: Done

proxy_backends/client/methods/convert_to_apiUmbrella_object.js
marla-singer: The structure of `noStandardObject` is
```
{
apiId: "c2amhNQpRXgHLMny9"
apiUmbrella.backend_host: "github.com"
apiUmbrella.backend_protocol: "https"
apiUmbrella.balance_algorithm: "least_conn"
apiUmbrella.frontend_host: "nightly.apinf.io:3002"
apiUmbrella.name: "00 SimpleSchema2"
apiUmbrella.servers: Array[1]
apiUmbrella.settings.disable_api_key: false
apiUmbrella.settings.rate_limit_mode: "custom"
apiUmbrella.settings.rate_limits: Array[1]
apiUmbrella.url_matches: Array[1]
proxyId: "3ZkC35aGDbBFvTnAa"
}
```
After `apiUmbrella` object is used here https://github.com/apinf/platform/blob/366578ce07e1b9743cfe18e92a12510b57e675d9/proxy_backends/client/form/autoform.js#L170-L174

And `createApiBackendOnApiUmbrella` method expects object as one of parameters
https://github.com/apinf/platform/blob/develop/proxies/server/methods/api_umbrella.js#L85

proxy_backends/client/methods/convert_to_apiUmbrella_object.js
marla-singer: Why did you think that I changed the structure of this object "so drastically"? Contrariwise, I create the expected object by parts.

proxy_backends/client/form/form.js
marla-singer: You didn't look below.
```javascript
    // Get proxy ID from template instance
    const savedProxyId = Template.instance().data.proxyId;

    // Make sure proxy ID exists or calculate it
    const proxyId = savedProxyId ? savedProxyId.get() : instance.getProxyId();
```

proxy_backends/client/form/form.js
marla-singer: Good question. Don't have any strong arguments.

proxy_backends/client/form/form.js
marla-singer: Done

proxy_backends/client/form/form.js
marla-singer: Done

proxy_backends/client/form/form.js
marla-singer: Agree that it doesn't look pretty semantic. Just me it needed a couple minutes to remember why it's so. 

When a user clicks on the proxies list and tries to change the current proxy to another, there are showing the modal form.
- If a user clicks on "Cancel" button, the proxy value must save the current state. For this, it needs `event` handler of `<select>` object and number of the previously selected option.
- If a user clicks on "Change" button, the proxy value must save the current state. For this, it needs the template instance of proxy backend form and value of the selected option.

marla-singer: I rename `proxyBackend` to `proxyBackendForm`

proxy_backends/client/form/autoform.js
marla-singer: Good suggestion

marla-singer: It is always trouble to create a variable name :confused:

marla-singer: Done

marla-singer:
@brylie Ready for repeated review

proxy_backends/client/methods/convert_to_apiUmbrella_object.js
brylie: How can we convert the `apiUmbrella` keys to be a properly nested object on creation? E.g. why do we store the object with keys like:

- "apiUmbrella.backend_host"
- "apiUmbrella.backend_protocol"
- ...?

It seems like the original design might have an oversight, which we can hopefully fix.

brylie: It would be cleaner to just store the API Umbrella fields as properties of an `apiUmbrella` object, rather than having to map and clean up the object later.

marla-singer: @brylie This object is parameter of AutoForm
https://github.com/apinf/platform/blob/develop/proxy_backends/client/form/autoform.js#L109-L115

brylie: Ah, that I see. That is not very convenient.

PR #2202, opened 2017-02-28
Feature/filter api via labels

### LINKED ISSUE #2133: In catalog, make life-cycle phase icon a link to filtered list
As a developer I would like to click from catalog "production" label and get filtered list of APIs containing just APIs in production. This is faster than selecting from filtering options.

# Wireframes

Currently, APIs with lifecycle status appears like below in catalog (card view)
apicatalognew

On clicking any lifecycle status (including unknown), the API cards get sorted and only cards with that specific lifecycle appears.
Catalog view also indicates how the result is filtered with and a clear all link text.
Clicking the "clear all" should refresh the page and display the default catalog.
apicatalognew4

Same effect should also be possible for tabular view:
apicatalogtablefiltered

marla-singer:
Closes #2133 
- Delete `Unknown` label if lifecycle isn't set
- Return the feature with text "filtered by" again

### DIFF
--- a/api_filter/client/filtered_by/filtered_by.html
+++ b/api_filter/client/filtered_by/filtered_by.html
@@ -4,7 +4,9 @@
       <strong>
         {{_ 'apiFilteredBy_text.apiCount' count=apisCount }}&nbsp;
       </strong>
-      <i>{{_ 'apiFilteredBy_text_filteredBy' tag=selectedOption }} </i>
+      <i>
+        {{_ 'apiFilteredBy_text_filteredBy' tag=selectedOption }}
+      </i>
       <a class="clear-all">
         {{_ 'apiFilteredBy_buttonText_clear' }}
       </a>
--- a/api_filter/client/filtered_by/filtered_by.html
+++ b/api_filter/client/filtered_by/filtered_by.html
@@ -1,0 +1,13 @@
+<template name="apiFilteredBy">
+  {{# if selectedOption }}
+    <p>
+      <strong>
+        {{_ 'apiFilteredBy_text.apiCount' count=apisCount }}&nbsp;
+      </strong>
+      <i>{{_ 'apiFilteredBy_text_filteredBy' tag=selectedOption }} </i>
+      <button class="btn btn-default btn-xs clear-all">
+        {{_ 'apiFilteredBy_buttonText_clear' }}
+      </button>
+    </p>
+  {{/ if }}
+</template>
--- a/api_filter/client/filtered_by/filtered_by.html
+++ b/api_filter/client/filtered_by/filtered_by.html
@@ -5,9 +5,9 @@
         {{_ 'apiFilteredBy_text.apiCount' count=apisCount }}&nbsp;
       </strong>
       <i>{{_ 'apiFilteredBy_text_filteredBy' tag=selectedOption }} </i>
-      <button class="btn btn-default btn-xs clear-all">
+      <a class="clear-all">
         {{_ 'apiFilteredBy_buttonText_clear' }}
-      </button>
+      </a>
     </p>
   {{/ if }}
 </template>
--- a/api_filter/client/filtered_by/filtered_by.html
+++ b/api_filter/client/filtered_by/filtered_by.html
@@ -7,7 +7,7 @@
       <i>
         {{_ 'apiFilteredBy_text_filteredBy' tag=selectedOption }}
       </i>
-      <a class="clear-all">
+      <a id="clear-all">
         {{_ 'apiFilteredBy_buttonText_clear' }}
       </a>
     </p>
--- a/api_filter/client/filtered_by/filtered_by.js
+++ b/api_filter/client/filtered_by/filtered_by.js
@@ -1,0 +1,39 @@
+import { Template } from 'meteor/templating';
+import { ReactiveVar } from 'meteor/reactive-var';
+import { FlowRouter } from 'meteor/kadira:flow-router';
+import { TAPi18n } from 'meteor/tap:i18n';
+
+Template.apiFilteredBy.onCreated(function () {
+  const instance = this;
+
+  instance.selectedOption = new ReactiveVar();
+
+  instance.autorun(() => {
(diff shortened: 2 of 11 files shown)


brylie:
> Delete Unknown label if lifecycle isn't set

@Nazarah @bajiat do we want to show the "Unknown" label when the API lifecycle is not set?

api_filter/client/filtered_by/filtered_by.html
```diff
@@ -0,0 +1,13 @@
+<template name="apiFilteredBy">
+  {{# if selectedOption }}
+    <p>
+      <strong>
+        {{_ 'apiFilteredBy_text.apiCount' count=apisCount }}&nbsp;
+      </strong>
+      <i>{{_ 'apiFilteredBy_text_filteredBy' tag=selectedOption }} </i>
```
brylie: Put nested child element on own line. Add explicit `&nbsp;` if space is desired:

```html
<i>
    {{_ 'apiFilteredBy_text_filteredBy' tag=selectedOption }}&nbsp;
</i>
```

api_filter/client/filtered_by/filtered_by.js
```diff
@@ -0,0 +1,39 @@
...
+Template.apiFilteredBy.helpers({
+  selectedOption () {
+    const instance = Template.instance();
+
+    return instance.selectedOption.get();
+  },
+});
+
+Template.apiFilteredBy.events({
+  'click .clear-all': () => {
```
brylie: `.clear-all` can be an ID, since there will be only one "clear all" button on the page.

apis/client/profile/lifecycle_status/lifecycle_status.less
```diff
@@ -0,0 +1,3 @@
+.generic-card span[data-lifecycle]:hover {
+  cursor:pointer;
```
brylie: Lint.

core/lib/i18n/en.i18n.json
```diff
@@ -88,6 +88,12 @@
   "apiExport_proxy_title": "Export proxy configuration",
   "apiExport_proxy_description": "Export proxy configuration in JSON or YAML format",
   "apiExport_title": "Export",
+  "apiFilteredBy_text": {
+    "apiCount": " __count__ API",
```
brylie: Cool! Parameterized i18n strings?

brylie:
This looks good. Really minor code changes/lint. The bigger question is whether or not to show the 'Unknown' label, since this was an explicit decision before.

brylie:
It looks like an outdated version of node bcrypt is breaking the Travis CI build. This is related to an [issue the restivus swagger project](https://github.com/apinf/restivus-swagger/issues/11).

api_filter/client/filtered_by/filtered_by.js
marla-singer: Exactly. I though a unique id must be on the entire site

brylie: > a unique id must be on the entire site

What do you mean?

brylie: > The id global attribute defines a unique identifier (ID) which must be unique in the whole document.
>
> MDN: [id - HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id)

core/lib/i18n/en.i18n.json
marla-singer: Yep. This changes text dependent on the number of items.

api_filter/client/filtered_by/filtered_by.js
marla-singer: Nevermind, I had the wrong data in my brain :sweat_smile:

brylie: It is an easy mistake, since we are sort-of building a 'single-page' app :smile:

brylie: In any case, it is minor. Also, after reading the MDN article, I am wondering if we are using IDs correctly anyway:

> This attribute's value is an opaque string: this means that web author must not use it to convey any information. Particular meaning, for example semantic meaning, must not be derived from the string.
>
> MDN [id - HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id)

What are your thoughts?

brylie:
OK, cool. Since you are allowing users to filter API Lifecycle Status by clicking a label, I can see how the 'unknown' status would add complexity. It seems like a good idea to hide the label when the lifecycle status is unknown, but I just want to loop @bajiat and @Nazarah into the decision.

api_filter/client/filtered_by/filtered_by.html
marla-singer: Done

apis/client/profile/lifecycle_status/lifecycle_status.less
marla-singer: Done

api_filter/client/filtered_by/filtered_by.js
marla-singer: Hm, yes, why not? We've used IDs like some anchor and for querySelector search and don't storage the semantic value

marla-singer:
@brylie You can read @Nazarah opinion in the issue
> In my opinion, "Unknown" label should not exist. If an API is not labeled with a lifecycle, it is clear to users from catalog/profile visualization that it is not yet within any lifecycle. keeping a status "Unknown" is itself vague and might create confusion among users.
However, for a quick fix, I suggest to have the similar action for unknown as any other labels. That is clicking unknown label should filter the results with all APIs that are not yet under any specific lifecycle label.
However, I'd suggest everyone to reconsider whether or not we should have an "unknown" status as filter category.

api_filter/client/filtered_by/filtered_by.js
marla-singer: Done

Nazarah:
In favor of removing label "unknown"

brylie:
Alright, well we basically have consensus then. We can explain the decision to @bajiat when she returns.

PR #2217, opened 2017-03-03
Filter APIs via label on organization page

### LINKED ISSUE #2173: On organization profile, allow filtering from lifecycle phase label
Related to #2133 

# User story

> As a developer
I would like to click from lifecycle label and get filtered list of APIs containing just APIs in production for the current organization
so that I can easily see for instance only production APIs for the organization I am interested in.

# Definition of done
* [ ] User can click on an API lifecycle label. As a result, I would see only the APIs that have the lifecycle label in question.
* [ ] User can see the criteria the APIs are filtered by and can press Clear all to clear the filter.

The implementation for #2133 should be reused as much as possible.

# Wireframes
In organization page, all Added APIs appear as below:
altorganizationprofile3

On clicking any lifecycle label (e.g. "Design" here), all APIs having this lifecycle phase appears in the search result.
The number of APIs in the page gets edited based on the filtered result.
A text shows how the result is filtered.
A "Clear All" link text appears beside the filtered criteria text.
on clicking the above text, the filtered gets cleared and the default result comes in view.
altorganizationprofileapifilter3

marla-singer:
Closes #2173


marla-singer:

api_filter/client/filtered_by/filtered_by.html
```diff
@@ -1,15 +1,17 @@
 <template name="apiFilteredBy">
   {{# if selectedOption }}
-    <p>
-      <strong>
-        {{_ 'apiFilteredBy_text.apiCount' count=apisCount }}&nbsp;
-      </strong>
+    <span>
+      {{# if showApisCount }}
```
brylie: Why not just do `{{# if apisCount }}` here? That way the `showApisCount` helper can be removed.

api_filter/client/filtered_by/filtered_by.js
```diff
@@ -29,6 +29,11 @@ Template.apiFilteredBy.helpers({
 
     return instance.selectedOption.get();
   },
+  showApisCount () {
+    const apisCount = Template.instance().data.apisCount;
+    // Return true if apisCount was set
+    return apisCount > -1;
```
brylie: Remember, zero is falsy. That means this helper might be unnecessary, since we are basically looking for zero or more APIs. We might not need to display anything if there are zero APIs.

api_filter/client/filtered_by/filtered_by.html
brylie: E.g. we don't need to display anything if there are zero APIs.

marla-singer: What is more explicitly and understanding for users:
- this one "**0 APIs** _filtered by "Design"_" + empty selection
- or this one "_filtered by "Design"_" + empty selection ?

And as you can see I limited only show or not apis count. It was done because Organization profile has own feature to showing apis count

brylie: The third option: don't show anything.

brylie: >  we don't need to display anything if there are zero APIs

brylie: A different idea would be to use `else` to display a different message when zero APIs exist:

```html
{{# if apisCount }}
  <strong>
    {{_ 'apiFilteredBy_text.apiCount' count=apisCount }}&nbsp;
  </strong>
{{ else }}
  <strong>
    {{_ 'apiFilteredBy_text.apiCount' count=0 }}&nbsp;
  </strong>
```

brylie: Declarative HTML is generally easier to understand than JavaScript tricks like checking whether a number is greater than negative one.

brylie: On a higher level, this is a chance to improve our coding style. Right now, the code is cleverly written, but is not self evident why it is written how it is. There are several ways to solve this, and I am trying to encourage the habit of writing clear, self-documenting code.

marla-singer: I did it for Organization profile. Organization profile has own counter
joxi_screenshot_1488543925911
And we need hide counter in `filtered_by` component if it's shown on organization page. On other hands, APIs catalog still needs apis counter.
joxi_screenshot_1488544036691

marla-singer: You can suggest to deprecate APIs counter from `filtered_by` component. In this case, I will create the helper for show/hide the apis counter on APIs catalog page. Anyway, there is additional helper

brylie: > In this case, I will create the helper for show/hide the apis counter on APIs catalog page. Anyway, there is additional helper

The point is that **we can do this without an additional helper**.

brylie: We are probably just conflating the APIs count and filter text. Decoupling the text/count would make this simpler.

brylie: I am creating a branch to show what I mean.

brylie: The first thing I notice, is that we have the same information in two templates (shown next to each other on the Organization Profile):

screenshot_20170303_151104

brylie: Lets just remove the second APIs count.

brylie: In effect, the Filtered By template only needs to show information about what filter is active.

brylie: screenshot_20170303_151353

brylie: We just need to figure out how to make it consistent on the API Catalogue page.
screenshot_20170303_152155

brylie: It is also worth pointing out that we should use the singlular word 'API' when showing only one API.

marla-singer: > It is also worth pointing out that we should use the singlular word 'API' when showing only one API.
It's already done

marla-singer: It's really awful to show count of API so high in title! 
screenshot_20170303_152427
**We do it for User Usability, don't for pretty code**

brylie: Here is an updated version of the Organization Profile:

screenshot_20170303_153930


screenshot_20170303_153848

marla-singer: Okay. What about  APIs catalog?

brylie: I opened a PR against this branch with the suggested changes.

https://github.com/apinf/platform/pull/2219

brylie: screenshot_20170303_154732

marla-singer:
@brylie Clean tag by myself

marla-singer:
@brylie Can it be merged?

jawidahmadi:
It is functioning well according to requirements. Only when you filter and perform a clear action, then the state of dropdown box doesn't change. Is it ok? Forexample dropdown value should become (select one) after clear button is used.

marla-singer:
@jawidahmadi Done. Please test it again

PR #2219, opened 2017-03-03
Org filter via label less helpers

brylie:
- Remove unneeded helper
- add conditional checks for `apisCount`
- Decouple API count and 'Filtered by' text.
- Pluralize i18n strings
- cleanup
- consistency between Organization Profile and APIs Catalog

# Organization Profile
screenshot_20170303_160454

# APIs Catalog
screenshot_20170303_161545

## APIs Catalog alternative
screenshot_20170303_160831


marla-singer:
@brylie Wait feedback from @Nazarah and @bajiat about replace counter
joxi_screenshot_1488548886188

brylie:
We can still work with the text positioning and wording. The main point here is to demonstrate the requested changes, suggesting a cleaner way to organize this feature, not to stalemate or argue. I am simply showing that the suggestions will work, since the conversation had become a bit of a struggle.

marla-singer:
@brylie I don't mind about your suggestion. It allows saving a simple way. But is it okay that we change the wireframe?

brylie:
@marla-singer a wireframe/mockup is generally a guide to get started. When we start implementing the code, things will naturally change slightly. It is not always possible, or desirable, to implement code exactly as initially sketched - although we try to stick as close to the original **intention**.

marla-singer:
@brylie Just recommendation, no obligation?

brylie:
It is not a *strict* oblication. We try, as developers, to honor the spirit of the mockup (the intention). However, we sometimes alter the design when we encounter limitations or constraints the design did not conceive.

marla-singer:
Your suggestion is good, anyway. One feature must do one thing. Either show count or filter criterion

brylie:
Cool. I have posted two alternatives for the API Catalog changes (in the PR description). Which do you like better?

api_filter/client/filtered_by/filtered_by.html
```diff
@@ -1,11 +1,6 @@
 <template name="apiFilteredBy">
   {{# if selectedOption }}
     <span>
-      {{# if showApisCount }}
-        <strong>
-          {{_ 'apiFilteredBy_text.apiCount' count=apisCount }}&nbsp;
```
marla-singer: Don't forget to delete useless i18n tags

marla-singer:
The second one. But don't place it in `<small>` tag, just like in Organization profile

api_filter/client/filtered_by/filtered_by.html
brylie: I can't find that tag in the en.i18n.js file.

marla-singer: https://github.com/apinf/platform/blob/develop/core/lib/i18n/en.i18n.json#L91-L94

brylie:
Here is what happens when I pull the text outside of the `<small></small>` tags:

screenshot_20170303_161952

nnn:
In my opinion, this [approach](https://github.com/apinf/platform/pull/2219#issuecomment-283959844) is better.

brylie:
Cool, @Nazarah also just reviewed it and will post her preference.

marla-singer:
@nnn Sure?
joxi_screenshot_1488551277687
joxi_screenshot_1488551309980

brylie:
@Nazarah likes the design that is currently active in this PR:

screenshot_20170303_161545

Nazarah:
In catalog view, the number of APIs can appear beside the "APIs" headline.
But "filtered by" text and clear all should come below the sorting bar.
Showing those texts in bigger letters beside headline doesn't make it look good.
Beside two views are different: organization profile and api catalog.
So in catalog, change text about filtered text should appear below the sort/action bar.

brylie:
Thanks @Nazarah. That corresponds with the current code in this branch:
screenshot_20170303_161545

marla-singer:
@brylie clean up tag

PR #2247, opened 2017-03-07
Added dot as allowed char in apiBasePathRegEx

### LINKED ISSUE #2152: Proxy path validation blocks adding huuto.net API
As a real world example, I tried to add huuto.net API to catalog and add root of the API to proxy. According to huuto.net API documentation, their root is "/1.1/".  http://dev.huuto.net/methods.html#root

kuvakaappaus 2017-02-20 18-53-46

kuvakaappaus 2017-02-20 18-57-18

matleppa:
Closes #2152 
Added dot (.) as an allowed character in apiBasePathRegEx.

### DIFF
--- a/proxy_backends/collection/regex.js
+++ b/proxy_backends/collection/regex.js
@@ -1,3 +1,3 @@
 export const proxyBasePathRegEx = new RegExp(/^\/[\w\-/]+\/$/);
 
-export const apiBasePathRegEx = new RegExp(/^\/[\w\-/]*$/);
+export const apiBasePathRegEx = new RegExp(/^\/[\w\-/.]*$/);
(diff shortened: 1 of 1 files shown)


brylie:
While we are working here, take time to review the documents linked in issue #2152. Get an understanding of various characters allowed in URLs, and make sure our RegEx is permissive enough to allow most or all of them. I.e. lets go a bit beyond the original error report.

brylie:
Ah, sorry. I merged this but realized it was a mistake. Some of the characters in your regular expression have special meaning in the regular expression. For example parenthesis are used as capture groups.

We need to escape any of the regular expression special characters with a backslash `\` character. Please open another pull request, and make sure to **escape any regular expression special characters**.

Issue #2185, opened 2017-02-24
UI changes based on normalization task


kyyberi:
Related to #2184

# Goal
API metadata contains information about organization and organization contact person, if API is connected to an organization
Migration is handled, if needed

# Definition of done (TBD by assignee)
- [x] Display all available information about connected Organization in API Metadata tab: Name, description, url
- [x] Organization name should be as a link to organization profile (optional)
- [x] When a user connects API to an organization, API Metadata contains information about these organization
- [x] When a user disconnects API from an organization, API Metadata doesn't contain information about these organization

bajiat:
@marla-singer This is related to #2184, so could you take this one as well?

marla-singer:
@bajiat Done

marla-singer:
@bajiat I suggest using all profit from linking ApiMetadata and Organization collections. For example, add URL and Logotip in ApiMetadata section which will be able from organization document.

bajiat:
@marla-singer We are experimenting with a process that each assignee creates the definition of done based on the goal of the task after the task has been assigned. Please create a definition of done and include also your ideas from the comment above. We can review and comment.

marla-singer:
@bajiat If API metadata doesn't exist and API is already connected to an organization. What to do? Create the API Metadata document by hand in code?

bajiat:
@marla-singer Definition of done seems ok.

> If API metadata doesn't exist and API is already connected to an organization. What to do?

In which situation this would occur?

marla-singer:
@bajiat 
-  For example, you have an API without API Metadata. 
![joxi_screenshot_1489390203919](https://cloud.githubusercontent.com/assets/20643403/23845245/0b283f74-07d8-11e7-9147-9434782ed0df.png)

- You decide to connected your API to an organization. Navigate to `Settings` tab, select the organization, the message about success operation is displayed.
- You navigate to `Metadata` tab. What are expected to see?

1. Empty page because Metadata document was empty before
![joxi_screenshot_1489390203919](https://cloud.githubusercontent.com/assets/20643403/23845266/21bc1490-07d8-11e7-9fce-2a916462f3f5.png)

2. Created page with organization data. In this case, the document will be created by the programmer behind the scene.
![joxi_screenshot_1489390301812](https://cloud.githubusercontent.com/assets/20643403/23845290/46a245a4-07d8-11e7-90dc-5d332ae1f539.png)

marla-singer:
[...]

PR #2254, opened 2017-03-07
Normalize apimetadata

### LINKED ISSUE #2184: Normalize ApiMetadata and Organizations collections
ApiMetadata collection contains organization (and organization contact) data, while there already is an organization collection. 

screenshot_20170303_131020

# Goal
All data is stored in database once or in one collection only.

screenshot_20170303_132750

# Definition of done
* [x] Organization fields exist only in Organizations collection
* [x] API Metadata object contains `organizationId` field
  * [x] `organizationId` field is a string, containing the ID of an Organization document
* [x] API Metadata UI displays fields from the related organization document, if connected
* [x] Schema migration has been defined from old database structure to new changes

marla-singer:
Closes #2184 
- Add migration to use `apiId` instead of `apiBackendId`
- Add migration to use `organizationId` instead of `organization` object
- Archive the old data
- Change schema of apiMetadata collection

### DIFF
--- a/apis/server/methods/delete.js
+++ b/apis/server/methods/delete.js
@@ -35,13 +35,13 @@ Meteor.methods({
     }
 
     // Remove backlog items
-    ApiBacklogItems.remove({ apiId });
+    ApiBacklogItems.remove({ apiBackendId: apiId });
 
     // Remove feedbacks
-    Feedback.remove({ apiId });
+    Feedback.remove({ apiBackendId: apiId });
 
     // Remove metadata
-    ApiMetadata.remove({ apiId });
+    ApiMetadata.remove({ apiBackendId: apiId });
 
     // Get proxyBackend
     const proxyBackend = ProxyBackends.findOne({ apiId });
--- a/apis/server/methods/delete.js
+++ b/apis/server/methods/delete.js
@@ -34,14 +34,16 @@ Meteor.methods({
       Meteor.call('removeMonitoring', apiId);
     }
 
+    // TODO: migrate to use 'apiId' instead of 'apiBackendId'
     // Remove backlog items
     ApiBacklogItems.remove({ apiBackendId: apiId });
 
+    // TODO: migrate to use 'apiId' instead of 'apiBackendId'
     // Remove feedbacks
     Feedback.remove({ apiBackendId: apiId });
 
     // Remove metadata
-    ApiMetadata.remove({ apiBackendId: apiId });
+    ApiMetadata.remove({ apiId });
 
     // Get proxyBackend
     const proxyBackend = ProxyBackends.findOne({ apiId });
--- a/core/migrations/server/4-change-api-metadata.js
+++ b/core/migrations/server/4-change-api-metadata.js
@@ -17,7 +17,7 @@ Migrations.add({
 
     // Linking between apiMetadata and Organization collections
     // Get all apiMetadata without organization data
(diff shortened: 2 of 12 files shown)


marla-singer:
@brylie 
I've tested migration and feature on local. Described the different cases and results.

### Case №1
- ApiMetadata contained information about Contact and Service
- API was connected
### Result after migration
- Organization block has information about connected organization
- Contact & Service information are still here

### Case №2
- ApiMetadata contained information about Contact and Service
- API was not connected
### Result after migration
- Contact & Service information are still here
- Organization block is empty as before

### Case №3
- ApiMetadata contained information about Organization & Contact
- API was connected
### Result after migration
- Contact information is still here
- Organization block has information about connected organization
- Old information from organization block is not displayed and is stored in database

### Case №4
- ApiMetadata contained information about Organization & Contact
- API was not connected
### Result after migration
- Contact information is still here
- Organization block isn't displayed and information is stored in database

### Case №5
- ApiMetadata didn't contain information
- API was connected
### Result after migration
- Contact & service blocks are empty as before
- Organization block has information about connected organization

### Case №6
- ApiMetadata contained information about Contact and Service
- API was not connected
### Result after migration
- Nothing happened

marla-singer:
### Cases after migration
Description of feature work when a user connects/disconnects API to organization
- When API is connected to an organization:
    - API Metadata will be created, if it doesn't have information before, and will have information about connected organization
    - API Metadata will be updated, if it has information about Contact or Service before, and will have information about connected organization as well

- When API is disconnected from an organization:
    - API Metadata will be removed totally, if it doesn't have information about Contact or Service
    - API Metadata will be updated, if it has information about Contact or Service.

brylie:
@marla-singer I would hope we can decouple the Organization from API Metadata. In other words, we probably don't need to trigger actions in the API Metadata collection when editing an Organization/API relationship.

marla-singer:
@brylie What? Then I don't understand why we do it. Maybe I misunderstood task at all. Why do we need `organizationId`? How will it be appeared after migration? :confused:

marla-singer:
@brylie Did you mean we don't need to trigger actions in this PR (it should be done in related #2185) or we don't need it at all?

brylie:
@marla-singer I am hoping we don't need trigger actions at all. Hopefully, we can just use collection helpers to retrieve related media. 

In effect, we don't want to duplicate data at all, and we want our features to be as independent as possible.

marla-singer:
@brylie OK, I see your point. You gave me another way to resolve. Let's save in this PR only migration

marla-singer:
@brylie Wait. What do you mean when you're talking about "trigger actions"?

marla-singer:
Remove trigger actions during connect/disconnect API ro/from an organization

marla-singer:
@brylie Ready for review

marla-singer:
> In effect, we don't want to duplicate data at all, and we want our features to be as independent as possible.
It's impossible to make apiMetadata independent because **it depends on Organization collection**
image

brylie:
> We have organizationApis collection and we can fetch data via collection helper and apiBackendId.

Feel free to try that out.

The key phrase in the quote above is "we want our features to be as independent **as possible.**"

brylie:
Also, the feedback was pertaining to the event triggers, which you removed.

marla-singer:
@brylie Triggers were removed from these PR. Can you please merge it? I'd like to go ahead and think about linking in 2185 related issue

PR #2283, opened 2017-03-14
Pages via URL

### LINKED ISSUE #2222: Empty page and error during navigating pages directly via URLs
In the process of working on #2056 issue, there were found the bug with accessibility on the pages "Branding", "Settings", "Proxies" directly via URL by an administrator.

However, that behavior repeats with an ordinary user who is log in. 

The result is an empty page and the error in browser console:
```
Uncaught RangeError: Maximum call stack size exceeded(…)
```
I repeated it on local. The error links to kadira-flow library.

### LINKED ISSUE #2281: The incorrect template is displayed after the page refresh
### Steps
1. Sign in as admin in nightly.apinf.io
2. Navigate to Users page (https://nightly.apinf.io/users)
3. Refresh the page

### Expected
The same page as before refresh

### Found
No authorize template

### Reason
The `Roles` package actually depends on a subscription, which means that if the subscription is not ready the `Roles.userIsInRole` method [will always return false](https://github.com/apinf/platform/blob/develop/users/client/lib/router.js#L28-L31).
 And then your route fails, because FlowRouter always runs no matter what.

marla-singer:
Closes #2222 
Closes #2281 

- Replace `Meteor.user()._id;` to `Meteor.userId()`. It did the stack overflow
- Add FlowRouter.wait(). It's waiting for the subscription ready. I found this way in [this article](https://medium.com/@satyavh/using-flow-router-for-authentication-ba7bb2644f42#.6i9w52ka9). Look at sentence **UPDATE May 26th 2016**

### DIFF
--- a/core/client/lib/router.js
+++ b/core/client/lib/router.js
@@ -10,7 +10,7 @@ import { Roles } from 'meteor/alanning:roles';
 Tracker.autorun(() => {
   // The Roles package actually depends on a subscription.
   // If the subscription is not ready the Roles.userIsInRole method will always return false.
-  // That is used for checking of is user is admin
+  // That is used for checking of does user have the admin role
 
   // Make sure the roles subscription is ready & FlowRouter hasn't initialized already
   if (Roles.subscription.ready() && !FlowRouter._initialized) {
--- a/core/client/lib/router.js
+++ b/core/client/lib/router.js
@@ -16,29 +16,6 @@ const signedIn = FlowRouter.group({
   }],
 });
 
-const requireAdminRole = function () {
-  if (Meteor.user()) {
-    // Get user ID
-    const userId = Meteor.user()._id;
-
-    const userIsAdmin = Roles.userIsInRole(userId, 'admin');
-
-    if (!userIsAdmin) {
-      // User is not authorized to access route
-      FlowRouter.go('notAuthorized');
-    }
-  } else {
-    FlowRouter.go('signIn');
-  }
-};
-
-// Define 404 route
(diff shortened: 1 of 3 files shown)


marla-singer:

core/client/lib/router.js
```diff
@@ -1,11 +1,24 @@
...
 // Meteor contributed packages imports
 import { BlazeLayout } from 'meteor/kadira:blaze-layout';
 import { FlowRouter } from 'meteor/kadira:flow-router';
 import { Roles } from 'meteor/alanning:roles';
 
+Tracker.autorun(() => {
+  // There are specific cases that this reruns
+
+  // Make sure the roles subscription is ready & FlowRouter hasn't initialized already
+  if (Roles.subscription.ready() && !FlowRouter._initialized) {
```
brylie: Is there a more FlowRouter specific way to wait for the Roles subscription? E.g. an `onInitialize` or `beforeInitialize` hook?

brylie: > FlowRouter limits reactive data sources to a single run; when it is first called.

It might not be the case that we can check for the Roles subscription in a different way. I just want to double check.

brylie: After scanning the FlowRouter docs, I can't come up with a different proposal for this code.

One thing that would be helpful, since this is a very unique situation, is to add a multi-line comment describing the need for the Tracker.autorun function. What problem is it solving?

marla-singer:
@brylie Add the comment. Is it full enough?

Issue #2220, opened 2017-03-03
User accessibility: user can add API when settings don't allow that


marla-singer:
In the process of working on #2056 issue, there were found the bug with user accessibility.
- if **`Only platform administrators are allowed to add new APIs`** option is checked nevertheless a user can add API via URL `https://nightly.apinf.io/apis/new`.
- Case: the user doesn't have any API and  **`Only platform administrators are allowed to add new APIs`** option is checked. The User has access to Dashboard page via URL and can click on `Add API` button.

### Part 1
1. Sign in as administrator
2. Go to Settings page
3. Set enabled for "Only platform administrators are allowed to add new APIs" in Adding APIs block
![joxi_screenshot_1489157813522](https://cloud.githubusercontent.com/assets/20643403/23799825/fdd66d2e-05ba-11e7-894d-5bf07ae9e327.png)
4. Save settings

### Part 2 
1. Sign in as ordinary user without administrator role
2. Make sure you don't have button "Add API" in navbar (it's expected)
![joxi_screenshot_1489158028926](https://cloud.githubusercontent.com/assets/20643403/23799948/7a28fe14-05bb-11e7-9b64-d781cb4f0ddd.png)
3. Navigate to `http://localhost:3000/apis/new` (or insert your localhost domain)

### Found
- A user can see form for filling data and can add new API

### Expected
- Show to users message with 403 message as we have for private api when user try to navigate to Private API via directly URL
![joxi_screenshot_1489158328865](https://cloud.githubusercontent.com/assets/20643403/23800130/2bf4f580-05bc-11e7-84cd-16edda6b1854.png)

bajiat:
@marla-singer Could you please add steps to reproduce? I'd like to assign the issue to @snittoor

marla-singer:
@bajiat Done
@snittoor If you have the questions about step, you can write me

bajiat:
@marla-singer Thanks!

bajiat:
Copying comments from #2361.

Comment from @snittoor
Issue:
  The code in the function userAddApi () in apis/client/lib/router.js , the Settings.find() is returning undefined. Can someone help why is it happenning?

Test Case failing:

1) If user is NonAdmin and  singleSetting.access.onlyAdminsCanAddApis = true .. then not authorised. This is failling because settings.findOne is returning undefined.

Comment from @brylie:

[...]

PR #2318, opened 2017-03-17
Added admin control for adding new APIs.

snittoor:
Added admin control for adding APIs.  

@marla-singer Could you please review it.

Issue #2220


.meteor/release
```diff
@@ -1 +1 @@
-METEOR@1.4.3.1
+METEOR@1.4.3.2
```
marla-singer: Please avoid running meteor update when working on tasks. We create specific tasks for updates.

.meteor/packages
```diff
@@ -3,8 +3,8 @@
 # 'meteor add' and 'meteor remove' will edit this file for you,
 # but you can also edit it by hand.
 
-accounts-base@1.2.14
-accounts-github@1.2.0
+accounts-base@1.2.15
```
marla-singer: Please avoid updating the packages when working on tasks. We create specific tasks for updates.

core/client/lib/router.js
```diff
@@ -77,6 +77,5 @@ const requireAdminRole = function () {
 FlowRouter.triggers.enter([redirectToCatalogue], { only: ['forgotPwd'] });
 
 // Routes that require admin role
-FlowRouter.triggers.enter([requireAdminRole], { only: ['settings', 'branding', 'proxies'] });
-
+FlowRouter.triggers.enter([requireAdminRole], { only: ['settings', 'branding', 'proxies','addApi'] });
```
marla-singer: Not correct way to the solution. Registered users can always add an API. But they can not do it if admin set the specific rule for that.  For example, look at a realization of `Add API` button on the navbar. You have to do the same.

https://github.com/apinf/platform/blob/develop/core/client/navbar/navbar.html#L50-L57

The logic is 
```js
if (userCanAddApi) {
  // Show the page Add API
} else {
  // Show the page Forbidden
 }
```

marla-singer:
## The available scenarios
### The first scenario
When `Only platform administrators are allowed to add new APIs` is not Enabled
joxi_screenshot_1490013940137

### Expected
1. Anonym user doesn't have access to page Add API. When this user goes to `/apis/new`, there should redirect to Sing in page.
2. A registered user has the access to page Add API. When this user goes to `/apis/new`, there displays the form "Add API".
3. Admin user has the access to page Add API. When this user goes to `/apis/new`, there displays the form "Add API".

### The second scenario
When `Only platform administrators are allowed to add new APIs` is Enabled.
joxi_screenshot_1490013940137 1

### Expected
1. Anonym user doesn't have access to page Add API. When user goes to `/apis/new`, there should redirect to Sing in page.
2. A registered user doesn't have the access to page Add API. When this user goes to `/apis/new`, there displays "Forbidden".
3. Admin user has the access to page Add API. When this user goes to `/apis/new`, there displays the form "Add API".

But now you allow adding a new API **only for admin**. It's incorrect because you resolved only the second scenario. You have to remember about the first scenario as well.

PR #2347, opened 2017-03-21
Bugfix/dashboard is blink

### LINKED ISSUE #2268: Dashboard console error when rendering proxy backend select widget
The dashboard on nightly.apinf.io is currently rendering blank. The following error is in the console:

```js
TypeError: Cannot read property 'name' of undefined
    at e.apiName (c984729….js?meteor_js_resource=true:447)
```

# Steps to reproduce
1. Log in as **admin** to nightly.apinf.io
2. Press Dashboard link

# Expected result
You are able to select an API for viewing

# Actual result
You see a blank page with only the title Dashboard

# Screenshot
empty dashboard

This may be due to a missing subscription in the Dashboard code:

screenshot_20170309_150406

marla-singer:
Closes #2268

### DIFF
--- a/apis/collection/server/publications.js
+++ b/apis/collection/server/publications.js
@@ -33,13 +33,6 @@ Meteor.publish('userManagedApisName', function () {
   return Apis.find(filter, { name: 1 });
 });
 
-Meteor.publish('apiBackend', (apiBackendId) => {
-  // Make sure apiBackendId is a String
-  check(apiBackendId, String);
-
-  return Apis.find({ _id: apiBackendId });
-});
-
 Meteor.publish('apisByIds', (apiIds) => {
   // Make sure apiIds is an Array
   check(apiIds, Array);
--- a/apis/collection/server/publications.js
+++ b/apis/collection/server/publications.js
@@ -6,6 +6,7 @@ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence
 // Meteor packages imports
 import { Meteor } from 'meteor/meteor';
 import { check } from 'meteor/check';
+import { Roles } from 'meteor/alanning:roles';
 
 // Collection imports
 import Organizations from '/organizations/collection';
@@ -13,12 +14,23 @@ import OrganizationApis from '/organization_apis/collection';
 import Apis from '../';
 
 // eslint-disable-next-line prefer-arrow-callback
-Meteor.publish('userManagedApis', function () {
+Meteor.publish('userManagedApisName', function () {
   // Get current user id
   const userId = this.userId;
 
-  // Get API Backends that user manages
-  return Apis.find({ managerIds: userId });
+  const userIsAdmin = Roles.userIsInRole(userId, ['admin']);
+  let filter;
+
+  // if user is not admin
+  if (userIsAdmin) {
+    // Set empty filter (return all APIs)
+    filter = {};
+  } else {
+    // Otherwise only managed
+    filter = { managerIds: userId };
+  }
+  // Get API names
+  return Apis.find(filter, { name: 1 });
 });
 
 Meteor.publish('apiBackend', (apiBackendId) => {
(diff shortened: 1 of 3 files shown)


marla-singer:

proxy_backends/collection/helpers.js
```diff
@@ -1,19 +0,0 @@
...
-This file is covered by the EUPL license.
-You may obtain a copy of the licence at
-https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence-eupl-v11 */
-
-import Apis from '/apis/collection';
-import ProxyBackends from '/proxy_backends/collection';
-
-ProxyBackends.helpers({
-// eslint-disable-next-line object-shorthand
-  apiName: function () {
```
brylie: We want to be able to get the API Name dynamically at runtime. In general, we want to **avoid writing triggers that cause changes in multiple collections**.

There is probably a different solution in this space.

apis/client/profile/settings/details/autoform.js
```diff
@@ -8,8 +8,27 @@ import { AutoForm } from 'meteor/aldeed:autoform';
...
+        const apiId = this.docId;
+        // Get related proxy backend configuration
+        const proxyBackend = ProxyBackends.findOne({ apiId });
+
+        // Make sure proxy backend configuration exists
+        if (proxyBackend) {
+          // Get the updated name
+          const apiName = this.updateDoc.$set.name;
+          // Update API name in proxy backend configuration
+          ProxyBackends.update(proxyBackend._id, { $set: { 'apiUmbrella.name': apiName } });
```
brylie: Lets see if we can find a run-time solution here, rather than using an autoform **update** hook on the **API Details** form to mutate data in the **ProxyBackends** collection.

In general, we want to **avoid writing triggers that cause changes in multiple collections**.

proxy_backends/collection/helpers.js
marla-singer: Ok. How do you assume to provide sorting in API selector?

apis/client/profile/settings/details/autoform.js
marla-singer: Don't see any reason to avoid trigger in this case.

brylie:
I see that we have difficulties sorting data based on a collection helper. How does this relate to the internationalized sorting work done by @manzapanza? I.e. the select list should probably follow the same sorting rules as the API catalog (in terms of API name).

apis/client/profile/settings/details/autoform.js
brylie: Well, the collection helper was intended as a way to avoid the trigger. What are some options for sorting?

proxy_backends/collection/helpers.js
brylie: I have asked @manzapanza to help us work through this challenge. Basically though, we are working with an array of objects, to create the select list. So, we might just map all of the proxy backends to retrieve the related API name and sort that based on localization rules.

dashboard/client/filtering/api_select/api_select.html
```diff
@@ -12,7 +12,7 @@
     <optgroup label="{{_ 'apiSelectPicker_myApis' }}" >
       {{# each backend in proxyBackends }}
         <option value="{{ backend._id }}" >
-              {{ backend.apiName }}
+          {{ backend.apiUmbrella.name }}
```
brylie: One additional thing to consider with this solution, is that we are actively working on multi-proxy support for our dashboard. So, the data structure `backend.apiUmbrella.name` is temporary. We can solve it this way for now, but this is another reason to consider alternative solutions.

proxy_backends/collection/helpers.js
manzapanza: You could create a template helper to sort the collection on client side. Try something like this:

```
import 'locale-compare-polyfill';
...
proxyBackends.sort((a, b) => {
    return a.apiName.localeCompare(b.apiName, language);
});
```

marla-singer:
@brylie Ready for review

Issue #2320, opened 2017-03-17
Organization view: Click on any API shows user info. text momentarily


saralavanip:
# Steps 
1.Visit https://apinf.io  v 0.41.0 
2.Login as admin/owner  
3.Click on Organizations
4.Go to any organization
5.Select and click on any API 

# Findings  
- shows info message "This Organization doesn't have any connected APIs " momentarily 
- Also reproducible on nightly V 0.41.0

marla-singer:
The miracle doesn't happen. Reopen it

bajiat:
Related to #2486 - could this be resolved by the suggested fix?

bajiat:
@55 See if this is fixed by  or related to #2486

PR #2353, opened 2017-03-23
Fix Api View route

manzapanza:
Closes: #2320


marla-singer:
It works too fast on local, couldn't observe any  improvement or deterioration/ Let's try it on nightly

PR #2360, opened 2017-03-24
Feature/friendly url

### LINKED ISSUE #2279: Understandable URLs for APIs and organizations
It might not sound a big issue but we need to have human understandable URLs for APIs.
This hardly is one: https://apinf.io/api/TYf33t5B6D6fwamKt (APInf Catalog REST API). Suggestion is to make it something like https://apinf.io/apis/APInf-Catalog-REST-API 

Note that in path I used plural "apis" which is the practice in REST APIs (in our catalogue API as well). This would make REST API and web behave with the same logic here. This would be just additional fine tuning and part of productisement.

marla-singer:
Closes #2279


apis/collection/server/publications.js
```diff
@@ -8,8 +8,12 @@ import { Meteor } from 'meteor/meteor';
 import { check } from 'meteor/check';
 
 // Collection imports
```
jawidahmadi: line 9 empty space may be removed.

feedback/client/form/autoform.js
```diff
@@ -10,11 +10,18 @@ import { Modal } from 'meteor/peppelg:bootstrap-3-modal';
 import { TAPi18n } from 'meteor/tap:i18n';
 import { sAlert } from 'meteor/juliancwirko:s-alert';
```
jawidahmadi: line 12 no need empty space.

brylie:
@jawidahmadi did you test this or read the code? There are changes in almost 20 files, and it would be surprising that no feedback was necessary.

backlog/client/form/autoform.js
```diff
@@ -8,12 +8,18 @@ import { AutoForm } from 'meteor/aldeed:autoform';
...
+import Apis from '/apis/collection';
+
 AutoForm.hooks({
   apiBacklogItemForm: {
     before: {
       insert (backlogItem) {
-        // Attach API Backend Id to backlog item
-        backlogItem.apiBackendId = FlowRouter.getParam('_id');
+        // Get related API document
+        const api = Apis.findOne({ slug: FlowRouter.getParam('slug') });
```
brylie: A more conventional way to do this in our code would be to pass the `apiId` in to the autoform template as an argument.

apis/collection/server/publications.js
```diff
@@ -70,15 +74,54 @@ Meteor.publish('latestPublicApis', (limit) => {
...
+          return Meteor.users.find(
+            { _id: { $in: api.authorizedUserIds } },
+            { fields: { username: 1, emails: 1, _id: 1 } }
+          );
+        },
+      },
+      {
+        find (api) {
+          // Make sure a user can edit this API
+          if (this.userId) {
```
brylie: This only seems to check that a user is logged in, while the comment indicates we want to 'Make sure a user can edit this API'. Does this code do what the comment indicates?

marla-singer: @brylie I can make the separate PR with improvements

PR #2361, opened 2017-03-27
Non admin add api

snittoor:
Refer to issue #2220 User can add Api when setting don't allow.

Issue:
  The code in the function userAddApi () in apis/client/lib/router.js , the Settings.find() is returning undefined. Can someone help why is it happenning?

Test Case failing:

1) If user is NonAdmin and  singleSetting.access.onlyAdminsCanAddApis = true .. then not authorised. This is failling because settings.findOne is returning undefined.

Thank you for the help.

### DIFF
--- a/apis/client/lib/router.js
+++ b/apis/client/lib/router.js
@@ -5,16 +5,53 @@ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence
 
 // Meteor packages imports
 import { Meteor } from 'meteor/meteor';
-
+import { Tracker } from 'meteor/tracker';
 // Meteor contributed packages imports
 import { BlazeLayout } from 'meteor/kadira:blaze-layout';
 import { FlowRouter } from 'meteor/kadira:flow-router';
+// APINF imports
+import signedIn from '/core/client/lib/router';
+import Settings from '/settings/collection';
 
-FlowRouter.route('/apis/new', {
+FlowRouter.wait();
+
+Tracker.autorun(() => {
+  // The Roles package actually depends on a subscription.
+  // If the subscription is not ready the Roles.userIsInRole method will always return false.
+  // That is used for checking of does user have the admin role
+
+  // Make sure the roles subscription is ready & FlowRouter hasn't initialized already
+  if (Roles.subscription.ready() && !FlowRouter._initialized) {
+    // Start routing
+    return FlowRouter.initialize();
+  }
+  // Otherwise nothing
+  return undefined;
+});
+
+signedIn.route('/apis/new', {
   name: 'addApi',
-  action () {
-    BlazeLayout.render('masterLayout', { main: 'addApi' });
-  },
+   action () {
+    const userId = Meteor.userId();
+      // Check that user is logged in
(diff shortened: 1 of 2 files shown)


apis/client/lib/router.js
```diff
@@ -5,18 +5,67 @@ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence
...
+});
+
+function userAddApi () {
+         console.log('in setValue function');
+           let singleSetting = Settings.find();
 
-FlowRouter.route('/apis/new', {
+           if (singleSetting) {
+
+             const onlyAdminsCanAddApis = singleSetting.access ? singleSetting.access.onlyAdminsCanAddApis : false;
```
brylie: @snittoor did you try using a Meteor method to retrieve this value from the server? If you use a Meteor method, you don't need to worry about the subscription. 

A Meteor method will return a value, which can be used to tell FlowRouter to proceed or redirect the user to a different page.

brylie: Ping @marla-singer. Can you help us to try how a Meteor method could be used instead of `Tracker.autorun()`? The `Tracker.autorun()` approach is somewhat unique in our project, and a Meteor method might be a simpler solution.

apis/client/lib/router.js
```diff
@@ -5,18 +5,67 @@ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence
...
   name: 'addApi',
-  action () {
-    BlazeLayout.render('masterLayout', { main: 'addApi' });
-  },
+   action () {
+    const userId = Meteor.userId();
+      // Check that user is logged in
+      if (userId) {
+       // Check if user is administrator
+        const userIsAdmin = Roles.userIsInRole(userId, ['admin']);
```
brylie: Try a Meteor method that does this same thing, but just on the server. That way, you don't need to worry about the Roles subscription.

brylie:
@snittoor did you try the Meteor method approach? If so, what difficulties did you encounter?

brylie:
I am fairly confident that [we can use Meteor methods inside of FlowRouter actions](https://github.com/apinf/platform/blob/develop/apis/client/lib/router.js#L38), meaning we can delay the rendering of the template while we check whether the user is an Admin.

apis/client/lib/router.js
marla-singer: @brylie Yeah, I think in this case we can hide all manipulation with `Roles` and `Settings` collections in Meteor method.

marla-singer: And I sure we shouldn't use `Tracker.autorun()` one more time in this file. It's enough now

brylie: @marla-singer would you consider pair-programming with @snittoor on this task?

marla-singer: @brylie Sure

bajiat:
Since there has be no activities on this PR, I'm closing it without merging, unassigning the issue and moving it to Planning column. Will move the comments to the issue, because @snittoor did a good job with finding the underlying problem.

PR #2392, opened 2017-03-30
Add Social Meta Tags

### LINKED ISSUE #2277: Metadata settings for social media
## Result now
At the moment it is not very appealing for API owner to share link to our API management. Below is how "HSL Vehicle positions" API is shown in FB.....
screenshot at maaliskuuta 10 21-05-04
 
## Expected
Given logo in the post. Title aka API name visible. API description visible. Possibly nice reference to apinf.io

manzapanza:
Closes #2277

### DIFF
--- a/apis/client/lib/router.js
+++ b/apis/client/lib/router.js
@@ -32,20 +32,20 @@ FlowRouter.route('/apis/:slug/', {
     const slug = params.slug;
 
     // Check if API exists
-    Meteor.call('apiExists', slug, (error, apiExists) => {
+    Meteor.call('getApi', slug, (error, api) => {
       // Check if API exists
-      if (apiExists) {
+      if (api) {
         // Set Social Meta Tags
         // Facebook & LinkedIn
-        DocHead.addMeta({ property: 'og:image', content: apiExists.logoUrl });
-        DocHead.addMeta({ property: 'og:title', content: apiExists.name });
-        DocHead.addMeta({ property: 'og:description', content: apiExists.description });
+        DocHead.addMeta({ property: 'og:image', content: api.logoUrl });
+        DocHead.addMeta({ property: 'og:title', content: api.name });
+        DocHead.addMeta({ property: 'og:description', content: api.description });
         DocHead.addMeta({ property: 'og:url', content: window.location.href });
         // Twitter
         DocHead.addMeta({ property: 'twitter:card', content: 'summary' });
-        DocHead.addMeta({ property: 'twitter:title', content: apiExists.name });
-        DocHead.addMeta({ property: 'twitter:description', content: apiExists.description });
-        DocHead.addMeta({ property: 'twitter:image', content: apiExists.logoUrl });
+        DocHead.addMeta({ property: 'twitter:title', content: api.name });
+        DocHead.addMeta({ property: 'twitter:description', content: api.description });
+        DocHead.addMeta({ property: 'twitter:image', content: api.logoUrl });
 
         // Ensure current user has permissions to view backend
         Meteor.call('currentUserCanViewApi', slug, (canViewError, userCanViewApi) => {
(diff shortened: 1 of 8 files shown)


manzapanza:
To do a truly test, we need to publish it in nightly.apinf.io

bajiat:
@marla-singer @mauriciovieira @matleppa @jawidahmadi @frenchbread Who of you would like to review?

organizations/collection/helpers.js
```diff
@@ -46,4 +48,37 @@ Organizations.helpers({
...
+      const objectId = new Mongo.Collection.ObjectID(this.organizationLogoFileId);
+
+      // Get Organization logo file Object
+      organizationLogo = OrganizationLogo.findOne(objectId);
+    }
+    return organizationLogo;
+  },
+  logoUrl () {
+    // Placeholder logo url with default image
+    let organizationLogoUrl = '/img/placeholder-logo.jpg';
```
brylie: This placeholder code should be moved to the template helper. In effect, this `organization.logoUrl()` function should only return the URL of the organization logo. Keep it simple and single purpose.

organizations/client/lib/router.js
```diff
@@ -37,9 +38,31 @@ FlowRouter.route('/organizations', {
...
+    // Get organization slug
+    const slug = params.slug;
+
+    // Get Organization
+    Meteor.call('organizationProfile', slug, (error, organizationProfile) => {
+      // Check if Organization exists
+      if (organizationProfile) {
+        // Set Social Meta Tags
+        // Facebook & LinkedIn
+        DocHead.addMeta({ property: 'og:image', content: organizationProfile.logoUrl });
```
brylie: Can you change this line to the following:

```
DocHead.addMeta({ property: 'og:image', content: organizationProfile.logoUrl() });
```

That way we don't need to override the `organization.logoUrl` helper.

apis/client/lib/router.js
```diff
@@ -34,6 +35,18 @@ FlowRouter.route('/apis/:slug/', {
     Meteor.call('apiExists', slug, (error, apiExists) => {
```
brylie: This return value is misleading. It is called `apiExists`, which implies a boolean. Instead, it seems to be an actual API object, so would be more semantic if it were just called `api`.

Likewise, the name of the server method is misleading, since it seems to imply returning a Boolean. Rather, it would be more semantic as `getExistingAPI`.

organizations/collection/helpers.js
brylie: In the Handlebars template, you can do something like:

```html
{{ if organization.logoUrl }}
  <img src="organization.logoUrl" />
{{ else }}
  <img src="/path/to/placeholder.png" />
{{/ if }}
```

organizations/server/methods.js
```diff
@@ -116,4 +116,17 @@ Meteor.methods({
...
   },
+  organizationProfile (slug) {
+    // Make sure slug is a string
+    check(slug, String);
+
+    // Look for Organization
+    const organization = Organizations.findOne({ slug });
+
+    // Attach logo url
+    organization.logoUrl = organization.logoUrl();
```
brylie: Lets not override this method name unless it is absolutely necessary. See above comments.

brylie:
The DocHead package looks fine. There are a few issues regarding semantics, overriding methods, and consistency between the `Api.logoUrl()` and `Organization.logoUrl()` collection helpers.

organizations/server/methods.js
manzapanza: But without this dosen't works. Sincerely I don't know why but the methods created with Meteor.methods dosen't returns collection helpers:  ```TypeError: apiExists.logoUrl is not a function```

organizations/collection/helpers.js
manzapanza: Is definitely better as you suggested, I will fix it

manzapanza:
I didn't add any package. The DocHead was already in the project.

organizations/collection/helpers.js
manzapanza: I changed my mind! To avoid duplicate the html code I will place this on uploadedOrganizationLogoLink template helper:

```
    // Return Organization logo or default file URL
    return organization.logoUrl() || '/img/placeholder-logo.jpg';
```

Issue #2039, opened 2017-02-06
Allow owner to connect API to eMQTT


bajiat:
_From @bajiat on December 1, 2016 8:16_

# User story
> As an API owner
I want to connect my MQTT API to eMQTT proxy
so that I can benefit from API management features, like rate limiting and analytics

# Definition of done
* [x] eMQTT can be selected as a proxy in Proxy tab in API profile
* [x] Owner can enter eMQ topic settings required for eMQTT keeping the same settings / structure as in Api Umbrella settings wherever possible

Open issue:
* Is there a need to expose ACL rules for owner?

# Wireframe
1. API owner can now select EMQTT proxy from API profile proxy tab
![emqtt](https://cloud.githubusercontent.com/assets/17976243/21003089/125d5342-bd32-11e6-9762-332f8128c8b1.png)

2. On selecting the EMQTT proxy, fields gets updated accordinly
(help texts under fields as well)
![emqtt2](https://cloud.githubusercontent.com/assets/17976243/21003124/49621f4e-bd32-11e6-8cde-68994d8b95af.png)

_Copied from original issue: Digipalvelutehdas/APIKA#72_

bajiat:
_From @frenchbread on December 4, 2016 21:22_

Interested

bajiat:
@frenchbread Please estimate the task

bajiat:
_From @frenchbread on December 5, 2016 11:25_

@bajiat Done 👍

bajiat:
Blocked by #161 and subsequent implementation task

bajiat:
@frenchbread Can you add an estimate to this issue?

frenchbread:
@bajiat Added estimation

PR #2411, opened 2017-04-05
Feature/add emq proxy

### LINKED ISSUE #2398: Allow admin to add eMQ as a proxy
# User story
> As an admin user
I want to add eMQ proxy to my deployment
so that real-time apis can be managed through the UI.

# Goal
API admin can add eMQ proxy to the deployment through Proxies view.
Admin is able to add the following configuration information for the eMQ broker:
- eMQ proxy URL
- eMQ Elasticsearch (analytics) endpoint
- eMQ API URL
- eMQ API credentials

frenchbread:
Closes #2398 

### Changes
 - Modified EMQ proxy schema. Added following fields: `proxyUrl`, `apiUrl`, `adminUsername`, `adminPassword`
- Modified UI (`proxyItem` template) accordingly to display added configs
- Modified validation function to validate newly added fields
- Added i18n tags

### TODO:
- [x] Fix `proxyItem` template to show emq proxy fields


frenchbread:
@apinf/developers Please review 🔍

core/lib/i18n/en.i18n.json
```diff
@@ -487,6 +487,10 @@
   "proxyItem_apiKey": "API Key",
   "proxyItem_apiToken": "API Token",
   "proxyItem_elasticsearch": "Elasticsearch",
+  "proxyItem_proxyUrl": "Proxy URL",
+  "proxyItem_apiUrl": "API URL",
+  "proxyItem_adminPassword": "Admin Username",
+  "proxyItem_adminUsername": "Admin Password",
```
phanimahesh: Shouldn't this be the other way around?

core/lib/i18n/en.i18n.json
```diff
@@ -487,6 +487,10 @@
   "proxyItem_apiKey": "API Key",
   "proxyItem_apiToken": "API Token",
   "proxyItem_elasticsearch": "Elasticsearch",
+  "proxyItem_proxyUrl": "Proxy URL",
+  "proxyItem_apiUrl": "API URL",
```
phanimahesh: Can we call this something else? Users may misunderstand this. This is the HTTP URL for an emq-rest-api instance, which apinf will use to manage MQTT users and their ACLs. All users are expected to connect directly to the MQTT broker specified in Proxy URL.

Additionally, the broker may allow connections over plain TCP, or over TLS, or over websockets, or over secure websockets. This is assuming additional protocol plugins are not enabled. A single "Proxy URL" can not capture all of these, we need multiple fields.

marla-singer:
@frenchbread Do you have the separate issue about "Implement emq proxy to Proxy tab in API profile" or something like that?

bajiat:
@marla-singer Please see #2039

core/lib/i18n/en.i18n.json
frenchbread: Yep, typo

core/lib/i18n/en.i18n.json
frenchbread: Agreed. Those are different port numbers, not URLs as long as default setup is used. It's possible to add either one field for broker and multiple fields for ports (TCP, TLS, websockets etc.) allowing to specify port mappings for each type of connection or add multiple URL fields for each conn type separately. Second approach is more flexible in case when broker is reverseproxies for to custom domain levels e.g. `tcp.domain.com`, `socket.domain.com` (counting on a fact that someone will do this)

phanimahesh: We can prefill the fields as soon as one of them is filled with defaults ports and same host for better UX. Potentially, they can all be different hostnames, on different machines even.

What if we let the user specify a list of hostname and port pairs, and for each pair we have check/radiobuttons for TLS/not, protocol:MQTT/MQTT over ws/CoAP/STOMP, etc?

marla-singer:
@phanimahesh I will re-assign you to this PR

core/lib/i18n/en.i18n.json
frenchbread: Sounds good, but this can add unneeded complexity to the UX

frenchbread: On the other hand, why not. I can't really get the picture of how this should look like.

Did you mean something like that?

| URL                            | PORT | TLS | Protocol |
| --------------------  | ----- | ---- | ----- |
| http://somehost.com | 1234 | yes | MQTT |
| http://somehost.com | 4321 | no | ... |

phanimahesh:
Yes, that's similar to what I had in mind. The host names should not have
protocol in them ( no http:// )

Protocol should be a dropdown. MQTT or MQTT over websockets should be the
options for now.

core/lib/i18n/en.i18n.json
phanimahesh: (copying from PR comment - it should have been made here - email replies always add to PR )

Yes, that's similar to what I had in mind. The host names should not have protocol in them ( no http:// ) Protocol should be a dropdown. MQTT or MQTT over websockets should be the options for now.

proxies/collection/schemas/emqtt.js
```diff
@@ -11,13 +11,46 @@ export default {
...
+    allowedValues: [
+      'MQTT',
+      'MQTT over websockets',
+    ],
   },
-  'emqtt.apiUrl': {
+  'emqtt.brokerEndpoints.$.host': {
+    type: String,
+  },
+  'emqtt.brokerEndpoints.$.port': {
```
brylie: Should this type be `Number`?

proxies/collection/schemas/index.js
```diff
@@ -34,7 +32,14 @@ let schema = {
...
 
-schema = _.assign(schema, apiUmbrellaSchema, emqttSchema);
+// Iterate through registered proxies and merge them all into one
+_.forEach(registeredProxies, (proxyName) => {
+  // Import proxy schema
+  const proxySchema = require(`./${proxyName}`).default
+
+  // Merge proxy schema with initial one
+  schema = _.assign(schema, proxySchema);
+});
```
phanimahesh: An `_.reduce` would have been nicer here IMO. Modifications in foreach blocks don't look good to me, but that is mostly a personal preference.

core/lib/i18n/en.i18n.json
```diff
@@ -780,12 +784,47 @@
...
+                  }
+              },
+              "httpApi": {
+                  "label": "HTTP REST API",
+                  "protocol": {
+                    "label": "Protocol"
+                  },
+                  "protocol": {
+                    "label": "Protocol"
+                  },
```
phanimahesh: duplicated protocol

proxies/client/form/autoform.js
```diff
@@ -17,27 +17,27 @@ import validateSchema from './validation';
...
+    //       return formData;
+    //     }
+    //   }
+    //
+    //   // Otherwise create & show message
+    //   const message = TAPi18n.__('proxyForm_errorText');
+    //   sAlert.error(message);
+    //
+    //   return false;
+    // },
```
phanimahesh: If this is unused it shouldn't be here. Please delete.

core/lib/i18n/en.i18n.json
```diff
@@ -780,12 +784,47 @@
...
+              },
+              "adminUsername": {
+                  "label": "Admin Username"
+              },
+              "adminPassword": {
+                  "label": "Admin Password"
               },
               "elasticsearch": {
-                  "label": "ElasticSearch"
+                  "label": "Analytics URL"
```
phanimahesh: I'm not sure about this label, shouldn't we mention ES? Thoughts?

frenchbread: Well, this field is made specifically for ES but potentially can be used for different data sources (can't imagine one at the moment, but generally mongodb or some other RESTful service for analytics)

frenchbread:
> Commits can be cleaned up by rebasing and squashing

I am on it.

In the mean time, @apinf/developers, could someone please test this?

brylie:
*Broker endpoints* and *HTTP REST API* panel styles are inconsistent:

screenshot_20170411_124142

brylie:
*HTTP REST API* admin credentials could be grouped better, to indicate that the admin username and password are related to the HTTP REST API:

screenshot_20170411_124409

brylie:
With such a complex configuration for EMQ Broker/HTTP REST API, we should really be adding inline help texts describing each field:

screenshot_20170411_124647

frenchbread:
> Broker endpoints and HTTP REST API panel styles are inconsistent

@brylie this is interesting, those are just default HTML5 form fields. I'm not sure we can do much at this point. We could change simpleschema `number` field type to string & add a regex, that will remove the arrows in `port` field

brylie:
URL fields for HTTP API and Elasticsearch are inconsistent, probably for simplicity:

- HTTP REST API
  - protocol
  - host
  - port
- Elasticsearch
  - URL

screenshot_20170411_125218

brylie:
Similar to the comment above, how can we better group the REST API settings with the admin username and password, so that it is **visually apparent** they are related?

screenshot_20170411_125845

frenchbread:
> how can we better group the REST API settings with the admin username and password

@brylie How about we move them to be under `httpApi` object?

frenchbread:
So it'll look like this:
screen shot 2017-04-11 at 13 06 41
screen shot 2017-04-11 at 13 06 36

frenchbread:
Pushed changes.

phanimahesh:
In response to @brylie 
>  Broker endpoints and HTTP REST API panel styles are inconsistent

@frenchbread said
> those are just default HTML5 form fields. 

I believe @brylie was referring to the panel labels "Broker Endpoints" and "HTTP REST API". The latter is in bold while the former isn't.

> What is the specific reason we handle URLs differently between HTTP REST API and Elasticsearch?

No reason. I missed this. Both HTTP REST user and access management API and Elasticsearch URL can have only one value, and it should be a single URL field.

frenchbread:
@phanimahesh Thanks for clarification. I misunderstood this. Agreed about moving HTTP API URL to a single field.

frenchbread:
Should we then hide the `httpApi` field from proxies list? (So that username/passwd are not exposed.

phanimahesh:
> Should we then hide the httpApi field from proxies list? (So that username/passwd are not exposed.

Can you hide just the credentials?

frenchbread:
Done. 

Looks like:

screen shot 2017-04-13 at 13 41 23

phanimahesh:
LGTM.
@brylie, any further comments?

PR #2416, opened 2017-04-05
Feature/media numbers

marla-singer:


organizations/client/content_management/form.js
```diff
@@ -11,12 +11,12 @@ Template.numberOfMediasPerPageForm.helpers({
   },
 });
 Template.numberOfMediasPerPageForm.events({
-  'click .edit-num-page': function () {
-    const selectedvalue = Template.instance().data.organization.numberOfMediasPerPage;
+  'click .edit-num-page': function (event) {
```
marla-singer: Delete this handler. We don't ned to update Organisation document by hand, autoForm do it automatically

PR #2420, opened 2017-04-06
Warning about using http protocol in swagger file

### LINKED ISSUE #2372: Issue with Try Out button in Swagger UI
# Issue
Try Out button in swagger editor in documentation tab seems not working.
The API endpoints, URL, proxy and proxy base path are defined correctly.
Results can be obtained using Postman, but swagger UI seems not working.

# Reproduction
* Create an API with the following information
  - API URL: http://makeup-api.herokuapp.com
  - Proxy: Nightly Umbrella
  - Proxy basepath suffix: /makeup-api/
  - API Basepath suffix: /api/v1/products.json
  - Example Screenshot:
image
  - get API key and then disable API key requirement
  - upload the json file given as (2nd) Comment
  - enable GET method
  - Try out the GET method with the this field value: **product_type** = _lipstick_
image

# Result
* the curl request doesn't return any value
image

* tried the same request using postman and the request returned search result for "lipstick"
image

# Consideration
* check out the json file to see if it has something to do with the way Definition is written

# API to test with:
https://nightly.apinf.io/apis/makeup-api

marla-singer:
Closes #2372

### User flow
1. Upload swagger file when `schemes` array containsonly `http` value
For example, using standard [swagger file](http://petstore.swagger.io/v2/swagger.json) 
**Expected:** displaying alert block
joxi_screenshot_1491463827906
 2. Upload swagger file when `schemes` array contains either `https` or `http, https` values
**Expected:** no alert block

@apinf/developers Ready


bajiat:
@brylie or/and @mauriciovieira Could you review this PR?

api_docs/client/swaggerUI/swaggerUi.html
```diff
@@ -4,31 +4,44 @@
...
-    </div>
-    <div id="message-bar" class="swagger-ui-wrap message-success" data-sw-translate=""></div>
-    {{# if dataFetched }}
+    {{/ if }}
+    <div class="swagger-section">
+      <div id="header">
+        <div class="swagger-ui-wrap">
+          <a id="logo" href="http://swagger.io" target="_blank" rel="nofollow">
+            <img src="/img/logo_small.png" class="logo__img" alt="swagger" height="30" width="30">
+            <span class="logo__title">swagger</span>
```
brylie: Perhaps remove this line?

brylie: That, or it should be something relevant, such as the API name.

marla-singer: It is header element 
joxi_screenshot_1491902510535

brylie:
In the back of my mind, I am thinking that **Apinf Platform does not enforce HTTPS protocol for deployments** (it is handled via the Docker packages and system administrator). If this is correct, there could be deployments of Apinf Platform that are accessed via HTTP. That would mean a different solution for this task.

However, we can probably assume that most/all deployments will use HTTPS. I just want to raise the issue.

cc: @bajiat @shaliko

core/lib/i18n/en.i18n.json
```diff
@@ -389,6 +389,7 @@
   "manageApiDocumentationModal_hints_uploadApiDocumentation": "You can add a Swagger document by uploading a file or by giving a link to the remote file. Note: Try out methods can only be controlled, if the file is uploaded.",
   "manageApiDocumentationModal_openDocumentationEditor": "Open API documentation editor",
   "manageApiDocumentationModal_addDocumentation_title": "Add documentation",
+  "manageApiDocumentationModal_helpText_useHttpsProtocol": "Application platform supports only HTTPS protocol. Keep in mind that requesting via Swagger might not work correctly if API endpoints use http protocol.",
```
brylie: This text can be shortened:

> This application platform uses HTTPS.
> API requests from documentation viewer will not work, if API uses HTTP.

core/lib/i18n/en.i18n.json
```diff
@@ -968,6 +969,8 @@
   "statusCodeCountsChart_title": "HTTP Status Codes",
   "statusCodeCountsChart_help_text": "You can use this chart to see the HTTP responses generated by your API for the requests it receives for a specific time period. You can filter the chart with response type for a single API or a combination of them. The overview information and other charts will be updated when you interact with HTTP Status Codes.",
   "swaggerUi_swaggerDocumentInvalidText": "Swagger document is not valid.",
+  "swaggerUi_alertText_Warning": "Warning!",
+  "swaggerUi_alertText_mainText": "Application platform supports only HTTPS protocol. Requesting via Swagger might not work correctly because API endpoints use http protocol.",
```
brylie: This text can be shortened:

> This application platform uses HTTPS.
> API requests from documentation viewer will not work, if API uses HTTP.

phanimahesh:
Why not make the alert conditional on current scheme from `document.location.protocol`? This warning is necessary because browsers enforce mixed content blocking, so whether requests succeed or not depend entirely on the scheme used when accessing this page.

api_docs/client/swaggerUI/swaggerUi.html
brylie: Cool. Does the text have to be 'swagger'?

brylie: I am just confused about what parts of this template we have written. Some of the code seems to come directly from Swagger UI, such as this line and the 'Explore' section below.

If we are providing this markup, we need to ensure that it can be translated.

marla-singer: Yes, it is

marla-singer: @brylie Also remember that Swagger green header has another button as "Authorization" which one we can't controll. So I don't see the profit from translate button "Explore" if "Authorization" dosn't have translation. 
joxi_screenshot_1491911850555

PR #2423, opened 2017-04-06
Apis after cover photo

### LINKED ISSUE #2375: APIs as first section below cover photo
# User story
As an APInf admin
I want users to see interesting (newest) APIs at the top of the page
so that they want to see more and visit the catalog.

# Goal
* APIs can be found as the first section after the cover photo.
* If possible, the size of the API card is adjusted for the mobile view (one card per row) to make it more easily clickable
  * see if you can use the same card, but use 12 columns in extra small screen size
* Number of APIs to be shown: 8
* There is no underlining or line below the header "Latest APIs"
* There is button to see "More APIs". If the user presses the button, they are redirected to the Catalog
  * Button color comes from Branding settings (primary color). Note that button text color needs to have sufficient contrast

### Definition of done
- [x] APIs displayed as first section below cover photo
- [x] Title "Latest APIs" has no underlining
- [x] Eight APIs are shown
- [x] One API card on a row in small device screens
- [x] A button with text "More APIs"
- [x] Pressing button redirects to the Catalog
- [x] Button color is the Primary color on Branding settings

# wireframe
1. Laptop view
frontpage1-1

2. Mobile View

mobile1
<br></br>
<br></br>
mobile3

matleppa:
Closes #2375

### DIFF
--- a/apis/client/latest/latest.html
+++ b/apis/client/latest/latest.html
@@ -7,6 +7,7 @@ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence
   <h2 class="feature-section-heading text-center">
     {{_ "latestApiBackends_Heading" }}
   </h2>
+  <br>
   {{# if latestPublicApis }}
     {{> apiCatalogGrid apis=latestPublicApis }}
   {{ else }}
--- a/apis/client/latest/latest.html
+++ b/apis/client/latest/latest.html
@@ -7,7 +7,6 @@ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence
   <h2 class="feature-section-heading text-center">
     {{_ "latestApiBackends_Heading" }}
   </h2>
-  <br>
   {{# if latestPublicApis }}
     {{> apiCatalogGrid apis=latestPublicApis }}
   {{ else }}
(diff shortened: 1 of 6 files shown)


apis/client/latest/latest.html
```diff
@@ -7,7 +7,7 @@
   <h2 class="feature-section-heading text-center">
     {{_ "latestApiBackends_Heading" }}
   </h2>
-  <hr class="primary">
+  <br>
```
nnn: Use CSS rule instead of ` <br>`

home/client/body/homeBody.html
```diff
@@ -25,9 +25,19 @@ <h1 id="site-title" class="text-center">
     </div>
   </div>
 
-
   <div class="container" id="features">
 
+    {{> latestPublicApis limit = 8 }}
+
+    <div class="row text-center">
```
nnn: No need for .row class <div> here, so just remove it.

home/client/body/homeBody.html
```diff
@@ -25,9 +25,19 @@ <h1 id="site-title" class="text-center">
...
+      <a href={{ pathFor 'apiCatalog' }}
+        class="btn btn-primary btn-lg"
+        id="get-more-apis">
+        {{_ "homeBody_getMoreApis_text" }}
+      </a>
+
```
nnn: Remove empty line.

PR #2431, opened 2017-04-07
Bugfix/save metadata

### LINKED ISSUE #2395: Add and save 'Metadata' not responding
# Steps  
1.Visit https://staging.apinf.io  v 0.42.0  
2.Login   
3.Add API with proxy (eg. API Umbrella New/second nightly)   
4.Go to 'Details' view  of above added API   
5.Go to  'Metadata'  view and click on 'Edit metadata'  
6.Enter Contact details (eg. Name/phone)  
7.Click on 'Save Metadata' 

# Expected Result  
Data saved successfully and displayed in 'Metadata' view    

# Actual Result 
Click on 'Save Metadata' not responding

marla-singer:
Closes #2395

### DIFF
--- a/core/migrations/server/4-change-api-metadata.js
+++ b/core/migrations/server/4-change-api-metadata.js
@@ -11,10 +11,8 @@ Migrations.add({
   version: 4,
   name: 'Uses apiId field instead of apiBackendId',
   up () {
-    // Code to migrate up to version 4
-
     // Migrate to use 'apiId' instead of 'apiBackendId' for documents with apiBackendId
-    ApiMetadata.find({ apiBackendId:{ $exists: true }}).forEach((apiMetadata) => {
+    ApiMetadata.find({ apiBackendId: { $exists: true } }).forEach((apiMetadata) => {
       // Create apiId field and get the value
       ApiMetadata.update(apiMetadata._id, { $set: { apiId: apiMetadata.apiBackendId } });
     });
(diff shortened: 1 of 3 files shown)


core/migrations/server/4-change-api-metadata.js
```diff
@@ -6,38 +6,17 @@ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence
 import { Migrations } from 'meteor/percolate:migrations';
 
 import ApiMetadata from '/metadata/collection';
-import OrganizationApis from '/organization_apis/collection';
 
 Migrations.add({
   version: 4,
```
mauriciovieira: I think the approach here would be to create another migration, but I don't know how the migration process work on currently deployed environments. So I ask @brylie and you @marla-singer if isn't changing this migration a problem.

brylie: In this case, the migration does the same thing. Basically, @marla-singer is removing unnecessary complexity from the migration.

@marla-singer can you please verify the expected result of the migration?

phanimahesh: If a migration has been applied on any instance, editing it is dangerous, and a new migration should be created if anything needs to be undone.

Since this migration is already in master we should expect that it may have been run somewhere, so I'm not sure if editing it is okay.

marla-singer: @phanimahesh I see your warning. Yes, migration has applied on staging and nightly sites but not on a production site. The second part of migration was implemented in process of working on this task. After a long discussion, we decided to not storage information about an organization in Metadata document at all. Now we don't need to create a new field in Metadata document and we don't use this field.

The expected result of the migration is to create the new field `apiId` and insert the value from `apiBackendId`.

brylie: We agree with @phanimahesh.

In this case, however, I believe we can accept @marla-singer's change. 

We will keep in mind that migrations should not change substantially going forward.

PR #2439, opened 2017-04-11
Pinning featured apis

### LINKED ISSUE #2311: Pinning featured APIs to the top of the API catalog
# Goal
* Organization manager can select certain APIs to be "Featured APIs" for this organization.
* The Featured APIs are displayed at the top of API area on the organization page, either as their own row or as first APIs.
  * If the APIs are sorted by name Featured APIs always precede any other APIs.  sorting by name means Featured APIs sorted by name followed by all other APIs sorted by name. Likewise for sorting by time added.
* Have a small visual indication of featured API on the API card.

# Definition of done
* [x] all users can see the featured API cards shown in own area above the area for all APIs
* [x] maximum amount of featured APIs per organization at a time is 4
* [x] organization manager can see a pin button in the upper right corner of API card
* [x] by pressing pin button the API card is marked as featured/not featured, which updates both screen view and the list of featured APIs
* [x] list of featured API card is attached to Organization
* [x] if featured API is deleted or disconnected from Organization, it is removed from Organization's featured API list

# Wireframe
Here is the new card
cardapidesign prototypepin
A Thumbtack icon appears at the corner of the API card.
we can use the following font [Awsome icon](http://fontawesome.io/icon/thumb-tack/)
Color: #555555
on mouse hover, there can be a tooltip telling organization manager about the featured API option.
When the icon is clicked, the API becomes featured and the icon gets opaque
cardapidesign prototypepinned

appearance of featured API in the API section in organization profile.
This should maintain the sorting rules explained above
featuredapi

**Open question**: would normal user viewing organization profile would see the opaque pin version of the API card that indicates the API to be a featured one?

matleppa:
Closes #2311


marla-singer:
@matleppa Why did you add other commits that are not related for your task?

matleppa:
This branch was containing unnecessary commits. It is replaces with PR 2445.

PR #2457, opened 2017-04-14
Catalog performance issues

### LINKED ISSUE #2351: Catalog performance issues
I've been using the catalog a lot lately and I'm somewhat worried about the performance. We have now like 120 APIs and getting listing of them takes several seconds. The same with organization views. Do we have means to collect facts about load times from end-user pov? There are free external tools for this, but do we have that info somewhere in our logs/system already?

marla-singer:
Closes #2351 
- Update package `kurounin:pagination` to 1.0.21 version
- Update subscribing for API logo
- Update subscribing for user names

@brylie Please look at updating of subscribing


api_catalog/client/grid/grid.js
```diff
@@ -0,0 +1,25 @@
+/* Copyright 2017 Apinf Oy
+ This file is covered by the EUPL license.
+ You may obtain a copy of the licence at
+ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence-eupl-v11 */
+
+// Meteor packages imports
+import { Template } from 'meteor/templating';
+
+Template.apiCatalogGrid.onCreated(function () {
+  const apis = Template.currentData().apis;
```
brylie: Since `Template.currentDAta()` is reactive, you are probably intending to use this inside of a reactive data context. However, template lifecycle callbacks (`onCreated` for instance) are not reactive, as they only run once at the lifecycle stage.

api_catalog/client/grid/grid.js
```diff
@@ -0,0 +1,25 @@
...
+  const apiLogoIds = [];
+
+  apis.forEach((api) => {
+    if (api.apiLogoFileId) {
+      apiLogoIds.push(api.apiLogoFileId);
+    }
+  });
+
+  // Subscribe to API logo collection
+  this.subscribe('apiLogoByIds', apiLogoIds);
```
brylie: Does this need to re-run at any point, or will this `onCreated` lifecycle callback be re-run somehow?

api_catalog/client/table/table.js
```diff
@@ -0,0 +1,28 @@
...
+ You may obtain a copy of the licence at
+ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence-eupl-v11
+ */
+
+// Meteor packages imports
+import { Template } from 'meteor/templating';
+
+Template.apiCatalogTable.onCreated(function () {
+  // Get APIs document
+  const apis = Template.currentData().apis;
```
brylie: Similar notes about reactivity as above.

Is this `onCreated` lifecycle callback running when we need it, or should we be using `templateInstance.autorun`?

users/collection/server/publications.js
```diff
@@ -5,7 +5,21 @@ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence
...
 // Meteor packages imports
 import { Meteor } from 'meteor/meteor';
+import { check } from 'meteor/check';
 
-Meteor.publish('allUsersUsernamesOnly', () => {
-  return Meteor.users.find({}, { fields: { username: 1 } });
+// NPM packages imports
+import _ from 'lodash';
+
+Meteor.publish('managersUsername', (apis) => {
```
brylie: Rename to `managersUsernames`  - managers have usernames (pluralization).

api_catalog/client/grid/grid.js
marla-singer: @brylie You're right. Maybe it was a simple copy-paste from another part. Reactive is irrelevant, will remove it

api_catalog/client/grid/grid.js
marla-singer: Nope, subscribing should be run once.

api_catalog/client/table/table.js
marla-singer: The same as above: reactive context is irrelevant.

users/collection/server/publications.js
marla-singer: Okay

api_catalog/client/grid/grid.js
brylie: Does the `apiLogoIds` array ever change when viewing the grid?

api_catalog/client/grid/grid.js
marla-singer: @brylie I could remember why I did so. Using a reactive data is needed but I totally forgot to use it in `autorun` function.

brylie: That's what I was thinking. It seems like the subscription for the API Logos needs to re-run (reactively) when the page changes (i.e. triggering a change to `apiLogoIds`).

brylie:
Removed the 'ready for review' label for safety. Add again, if/when this is ready.

api_catalog/client/grid/grid.js
marla-singer: @brylie Fixed it. It's ready

api_catalog/client/table/table.js
```diff
@@ -0,0 +1,30 @@
...
+      if (api.apiLogoFileId) {
+        apiLogoIds.push(api.apiLogoFileId);
+      }
+    });
+
+    // Subscribe to API logo collection by logo IDs
+    this.subscribe('apiLogoByIds', apiLogoIds);
+
+    // Subscribe to usernames of API managers
+    this.subscribe('managersUsernames', apis);
```
brylie: This subscription may be duplicating the [API composite publication](https://github.com/apinf/platform/blob/77097e978bda788667c7ec49ef74ffba744efe5c/apis/collection/server/publications.js#L68). In effect, the API composite publication returns users who are in `authorizedUserIds` or `managerIds` arrays:


```js
return Meteor.users.find({
            $or: [
              { _id: { $in: api.managerIds } },
              { _id: { $in: api.authorizedUserIds } },
            ],
          },
{ fields: { username: 1, emails: 1 } 
});
```

If I am not mistaken, we can clean-up/remove the `managersUsernames` subscription/publication.

marla-singer: @brylie Don't forget that `apiCompose` function we use **only in API profile template.** 
`apiCatalogGrid` and `apiCatalogTable` templates are not intersected with `apiCompose` function.

brylie: Good point. What would you think about making an `apiCatalogItemComposite` publication, similar to the API Profile, but subscribing only to information needed for an API Catalog item?

brylie: If it is too difficult or complex, we can use the current two subscription solution.

marla-singer: Yes, while we have two subscribing functions, it is better to leave it right now as for me. And I don't think that composite function can me applied in this case.

PR #2458, opened 2017-04-16
Bugfix/no admin settings

### LINKED ISSUE #2220: User accessibility: user can add API when settings don't allow that

marla-singer:
Closes #2220 
- User does not have permissions to add an API if user is anonymous
- - User does not have permissions to add an API if settings "Only administrator can add an API" and user is not admin

### DIFF
--- a/apis/server/methods/authorization.js
+++ b/apis/server/methods/authorization.js
@@ -60,29 +60,28 @@ Meteor.methods({
     return api && api.currentUserCanManage();
   },
   currentUserCanAddApi () {
-    // Get Settigns
-    const settings = Settings.findOne();
-
     // Get current user Id
     const userId = Meteor.userId();
 
-    if (settings) {
-      // If the access permission 'only admins can add APIs' is defined, use it
-      // Otherwise, allow registered users to add APIs by default
-      const onlyAdminsCanAddApis = settings.access ? settings.access.onlyAdminsCanAddApis : false;
+    // Make sure user is logged in
+    if (userId === null) {
+      // Anonymous user not allowed to add API
+      return false;
+    }
+    // Get Settings (used to check for access permission)
+    const settings = Settings.findOne();
 
-      if (!onlyAdminsCanAddApis) {
-        // Allow user to add an API
-        return true;
-      }
+    // If the access permission 'only admins can add APIs' is defined, use it
+    // otherwise, allow users to add APIs by default
+    const onlyAdminsCanAddApis = settings.access ? settings.access.onlyAdminsCanAddApis : false;
 
-      // Otherwise only admin can add an API
+    // Check if only admins are allowed to add APIs
+    if (onlyAdminsCanAddApis) {
       // Make sure current user is admin
       return Roles.userIsInRole(userId, ['admin']);
     }
 
-    // Registered user can add an API then returns true
-    // Anonymous user can not add an API then returns false
-    return !!userId;
+    // Allow users to add APIs by default
+    return true;
   },
 });
--- a/apis/server/methods/authorization.js
+++ b/apis/server/methods/authorization.js
@@ -9,10 +9,12 @@ import { check } from 'meteor/check';
 
 // Meteor contributed packages imports
 import { Accounts } from 'meteor/accounts-base';
+import { Roles } from 'meteor/alanning:roles';
 import { ValidEmail } from 'meteor/froatsnook:valid-email';
 
 // Collection imports
 import Apis from '/apis/collection';
+import Settings from '/settings/collection';
 
 Meteor.methods({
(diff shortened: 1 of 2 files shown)


apis/server/methods/authorization.js
```diff
@@ -57,4 +59,29 @@ Meteor.methods({
...
   },
+  currentUserCanAddApi () {
+    // Get Settigns
+    const settings = Settings.findOne();
+
+    // Get current user Id
+    const userId = Meteor.userId();
+
+    if (settings) {
+      // If access field doesn't exist, these is false. Allow users to add an API on default
```
brylie: Code and comment are a bit confusing. Consider a descriptive comment such as

```
// If the access permission 'only admins can add APIs' is defined, use it
// otherwise, allow users to add APIs by default
```

apis/server/methods/authorization.js
```diff
@@ -57,4 +59,29 @@ Meteor.methods({
...
+      const onlyAdminsCanAddApis = settings.access ? settings.access.onlyAdminsCanAddApis : false;
+
+      if (!onlyAdminsCanAddApis) {
```
brylie: Can we invert this logic for readability? E.g.

```js
// If the access permission 'only admins can add APIs' is defined, use it
// otherwise, allow users to add APIs by default
const onlyAdminsCanAddApis = settings.access ? settings.access.onlyAdminsCanAddApis : false;

// Check if only admins are allowed to add APIs
if (onlyAdminsCanAddApis) {
  // Make sure current user is admin
  return Roles.userIsInRole(userId, ['admin']);
}

// allow users to add APIs by default
return true;
```

brylie: Basically, avoiding negation in conditional statements makes it easier to understand the code. Less 'logical inversion'.

apis/server/methods/authorization.js
marla-singer: @brylie Done

apis/server/methods/authorization.js
```diff
@@ -57,4 +59,30 @@ Meteor.methods({
     // Check if user can edit
     return api && api.currentUserCanManage();
   },
+  currentUserCanAddApi () {
```
brylie: Consider rewriting this function for simple logic flow:

Can we invert this logic for readability? E.g.

```js
// Get current user Id
const userId = Meteor.userId();

// Make sure user is logged in
if (userId === null) {
  // Anonymous user not allowed to add API
  return false;
} else {
  // Get Settigns (used to check for access permission)
  const settings = Settings.findOne();

  // If the access permission 'only admins can add APIs' is defined, use it
  // otherwise, allow users to add APIs by default
  const onlyAdminsCanAddApis = settings.access ? settings.access.onlyAdminsCanAddApis : false;

  // Check if only admins are allowed to add APIs
  if (onlyAdminsCanAddApis) {
    // Make sure current user is admin
    return Roles.userIsInRole(userId, ['admin']);
  }

  // allow users to add APIs by default
  return true;
}
```

brylie: I.e. `Meteor.userId()` will return `null`, if no user is logged in:
https://docs.meteor.com/api/accounts-multi.html#AccountsCommon-userId

Checking for `null` is explicit and somewhat declarative, improving reader comprehension. The second half of the method is wrapped in `else` and checks for permissions in an explicit and declarative manner (without logical inversion).

brylie:
Overall great solution. Very few lines of code and clear reasoning. I suggested some improvements to the server method, mainly to make it a bit easier for humans to understand the logic.

apis/server/methods/authorization.js
marla-singer: Not bad. Let's use your first suggestion

marla-singer:
@brylie Ready

Issue #2322, opened 2017-03-17
Organization name visible as "who added" the API


bajiat:
# Context
Both the API catalog and API profile show information on who has added the API. 

# Current status
* API catalog shows the user who has added the API regardless of whether the API has been connected to an organization.
* API profile
  * always shows which user has added the API
  * if the API is connected to an organization, the organization name is displayed in addition to the user

# Proposal
* If API is connected to an organization, **only** the organization is shown both in Catalog and on API profile page.
* User is shown only for APIs that are not connected to an API.

PR #2472, opened 2017-04-19
Show organization name on api card

manzapanza:
Closes: #2322


marla-singer:
1. Don't forget that API catalog has two view mode: Grid and Table. It means we have 2 different templates for that `apiCard` and `apiCatalogTable`

Now you should add the same changes for `apiCatalogTable` template
**Grid view (correctly):**
joxi_screenshot_1492761317291 1


**Table view (incorrectly):**
joxi_screenshot_1492761317291

2. > If API is connected to an organization, only the organization is shown both in Catalog **and on API profile page.**

API profile page still has both user name and organization name. It is not correctly.

apis/client/card/api_card.js
```diff
@@ -0,0 +1,16 @@
...
+// Meteor packages imports
+import { Template } from 'meteor/templating';
+
+Template.apiCard.onCreated(function () {
+  // Get reference to template instance
+  const instance = this;
+  // Get id of API
+  const apiId = instance.data.api._id;
+  // Subscribe to basic details of related api metadata
+  instance.subscribe('apiOrganizationBasicDetails', apiId);
```
marla-singer: The ideal result is to add subscribing organizationApis and allOrgnaizations into apiCatalog template and home template because we have a few APIs in home page as well.

marla-singer:
@manzapanza A few changes:
**API catalog Table view (It's okay)**
joxi_screenshot_1493220796638 2

**Front page, Latest APIs block (It's wrong. There must be "added by Company")**
joxi_screenshot_1493220796638

**The page of current API (It's wrong. There must be only line with organization name)**
joxi_screenshot_1493220796638 1

marla-singer:
@manzapanza The last step is to add the one condition for API profile page.
Go to APIs page -> pick any API -> you will be navigated to API profile page like this one:
joxi_screenshot_1493392583348

As I understand from task:
- If API is connected to organization
joxi_screenshot_1493392629332

- If API is not connected to organization it should be like that
joxi_screenshot_1493392629332 1

marla-singer:
@manzapanza Looks good for me. Merge it

PR #2477, opened 2017-04-20
Footer code for platform admin

### LINKED ISSUE #2288: Footer code for platform admin
# Scenario
As owner of an installation of APInf, I want to analyze how users are going through the platform. What do they do, where do they drop off, where do they come from etc.

# Solution
There are a lot of 3rd party tools that are typically integrated with a line of Javascript code added in the bottom of the master template. By adding a field for admins to fill out, we can support any of these.

jawidahmadi:
close #2288 

### Defenition of done
- [x] Only admin can add the scripts
- [x] Add a field for the script in the schema
- [x] Add input field in brand settings
- [x] Print script(s) in the footer

### DIFF
--- a/home/client/footer/homeFooter.html
+++ b/home/client/footer/homeFooter.html
@@ -21,6 +21,7 @@ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence
         <div class="col-lg-5">
           <p class="footer-text">
             {{ branding.siteFooter }}
+            {{ branding.footerCode }}
           </p>
         </div>
         <div class="col-lg-3">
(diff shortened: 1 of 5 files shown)


home/client/footer/homeFooter.html
```diff
@@ -65,5 +65,6 @@
         </div>
       </div>
     </div>
+    {{ branding.footerCode }}
```
brylie: This needs to use triple curly-braces, so that our template system does not sanitize the code:

```js
{{{ branding.footerCode }}}
```

marla-singer:
@brylie I'll approve this PR and make the change in my related PR

PR #2479, opened 2017-04-20
Publish only public fields

### LINKED ISSUE #2221: Publish only necessary fields to catalog page
In the process of working on #2056 issue, there were found no effective data using
- Apis and Organisations collections have all information about items. But not all are used. Is it really necessary? Can we limit some fields? We can start, for example, from Organisations catalogue.

manzapanza:
Closes #2221


manzapanza:
@marla-singer I'm following the documentation [Security - Always restrict fields](https://guide.meteor.com/security.html#fields) but seems that all organization object is being published. What am I doing wrong?

marla-singer:
@manzapanza When I created the issue I meant that we publish all fields for Organizations and APIs catalogs. Yes, all fields are object is being published but we don't use **all data from this fields**.
I looked at the catalogs again and found that Organization catalog uses fields _name, description, createdAt, updatedAt, organizationLogoFileId, slug_ and APIs catalog uses all fields except for _latestMonitoringStatusCode_ and _url_. Looks like we don't upgrade publish function for APIs catalog.

Also, we don't need to create a new publish function to publish only necessary fields. Just add some changes for [Organizations Pagination function](https://github.com/apinf/platform/blob/develop/organizations/collection/server/publications.js#L32).

More details about Pagination package [here.](https://atmospherejs.com/kurounin/pagination)

apis/collection/server/publications.js
```diff
@@ -44,20 +47,24 @@ Meteor.publish('latestPublicApis', (limit) => {
...
+  },
+});
 
 // eslint-disable-next-line prefer-arrow-callback
 Meteor.publishComposite('apiComposite', function (slug) {
   check(slug, String);
   return {
     find () {
-      return Apis.find({ slug });
+      return Apis.find({ slug }, { fields: Apis.publicFields });
```
marla-singer: @manzapanza Function `Meteor.publishComposite('apiComposite')` have to returns all fields. Please remove `fields` from here

organizations/collection/index.js
```diff
@@ -9,4 +9,17 @@ import { Mongo } from 'meteor/mongo';
 
 const Organizations = new Mongo.Collection('Organizations');
 
+// Default public fields
+Organizations.publicFields = {
+  _id: 1,
+  apisPerPage: 1,
+  contact: 1,
+  description: 1,
+  logoUrl: 1,
```
marla-singer: `Organization` document doesn't have field `logoUrl`. Please remove this field

organizations/collection/server/publications.js
```diff
@@ -10,26 +10,26 @@ import { check } from 'meteor/check';
...
-    {
-      fields: {
-        _id: 1,
-        name: 1,
-        description: 1,
-        contact: 1,
-      },
-    });
+  // return Organizations.find();
+  return Organizations.find({}, { fields: Organizations.publicFields });
```
marla-singer: Let's leave this function as we had before.
We often use `allOrganizationBasicDetails` when we really need only couple fields or Organization document doesn't exist yet, for example, "Add new Organization" form

organizations/collection/server/publications.js
```diff
@@ -50,7 +50,7 @@ Meteor.publishComposite('organizationComposite', (slug) => {
   check(slug, String);
   return {
     find () {
-      return Organizations.find({ slug });
+      return Organizations.find({ slug }, { fields: Organizations.publicFields });
```
marla-singer: `organizationComposite` function needs to have access to all fields because it is used on Organization profile page which contains all information about the organization.

Please remove list of fields from here

organizations/collection/server/publications.js
```diff
@@ -10,26 +10,25 @@ import { check } from 'meteor/check';
...
-  return Organizations.find({},
-    {
-      fields: {
-        _id: 1,
-        name: 1,
-        description: 1,
-        contact: 1,
-      },
-    });
+  return Organizations.find({});
```
marla-singer: Please leave the old set of filed
```js
{
      fields: {
        _id: 1,
        name: 1,
        description: 1,
        contact: 1,
      },
    }
```

brylie:
@manzapanza what is the status of this PR?

PR #2483, opened 2017-04-21
Allow owner to connect api to emq

### LINKED ISSUE #2039: Allow owner to connect API to eMQTT

frenchbread:
Closes #2039 

### Changes
- i18n tags
- Added check of proxy backend type selected to autoform hooks
- Added check of proxy backend type selected to a template
- Moved proxies template from folder `schemas` to `proxies`
- Added mechanism for dynamic form construction based on available proxies
- Added EMQ proxy schema
- Added EMQ ACL rules `Add` and `Update` functionality

### TODO:
- [x] Add i18n tags to labels in a template
- [x] Fix "Get API key" button - related issue #2573
- [x] Show notification when updating ACLs
- [x] Fix delete proxy button
- [x] Make EMQ ACL field in schema optional


frenchbread:
POSTing ACL rules to EMQ-REST-API (postgres db) works. Next I'm going to add possibility to update rules remotely as well (in addition to meteordb). It's a bit more complex, since for example when adding rules, you can add as much as you like basically and this works fine. But when editing them, beyond regular update of single rule, some can add rules in addition to already existing one, and this needs some additional logic (e.g. need to pre-fetch existing rules and check if one is already added, so that it's not added second time and only updated, and add those which have not been added yet.

PS: Don't load your brain trying to understand this, in general, POSTing ACLs works, updating rules is on the way.

frenchbread:
Ready for review.

proxy_backends/client/form/autoform.js
```diff
@@ -27,59 +27,36 @@ AutoForm.hooks({
...
                     // return false;
-                    form.result(false);
-                  }
+                  form.result(false);
+                }
 
                   // If success, attach API Umbrella backend ID to API
-                  if (response.result && response.result.data &&
-                    response.result.data.api) {
+                if (response.result && response.result.data && response.result.data.api) {
```
brylie: Can we use `_.get()` here?
https://lodash.com/docs/4.17.4#get

proxy_backends/client/form/autoform.js
```diff
@@ -104,6 +81,19 @@ AutoForm.hooks({
               form.result(false);
             }
           });
+        } else if (proxyBackend.type === 'emq') {
+          // Before insert iterate through ACL rules
+          proxyBackend.emq.settings.acl.forEach((aclRule) => {
+            // Adding ID field for each rule separately is needed to differentiate
+            // add edit them
+            aclRule.id = new Meteor.Collection.ObjectID().valueOf();
```
brylie: Should we use `_id` instead of `id`?

marla-singer:
@frenchbread Do we have a real instance of MQT proxy to test it?

proxy_backends/client/form/form.html
```diff
@@ -29,150 +23,21 @@
...
-      }}
+    {{/ if }}
 
-      <!-- Request Headers -->
-      {{> afQuickField
-          name="apiUmbrella.settings.headers_string"
-          placeholder="X-Example-Header: value"
-      }}
+    <!-- EMQ -->
+    {{# if equals proxy.type "emq" }}
```
marla-singer: Did you try using helper from autoform? For example, 
```html
{{#if afFieldValueIs name="proxy.type" value="emq" }}
```

proxy_backends/client/form/form.html
```diff
@@ -29,150 +23,21 @@
...
-        {{ afFieldLabelText name='apiUmbrella.url_matches.0.frontend_prefix' }}
-      </label>
+    <!-- Check which proxy is selected and render related fields -->
 
-      <div class="input-group">
-      <span class="input-group-addon" id="proxy-url">
-        {{ proxy.apiUmbrella.url }}
-      </span>
+    <!-- API Umbrella -->
+    {{# if equals proxy.type "apiUmbrella" }}
```
marla-singer: Did you try using helper from autoform? For example, 
```html
{{#if afFieldValueIs name="proxy.type" value="apiUmbrella" }}
```

proxy_backends/client/form/form.js
```diff
@@ -176,6 +180,12 @@ Template.proxyBackend.helpers({
     // Return boolean value
     return !!(instance.data.proxyBackend);
   },
+  registeredProxies () {
```
marla-singer: I did search for PR and did not find using this helper. Do you need it?

proxy_backends/collection/schemas/emq.js
```diff
@@ -0,0 +1,84 @@
...
+    allowedValues: [
+      1,
+      2,
+      3,
+    ],
+  },
+  topic: {
+    type: String,
+    autoform: {
+      placeholder: 'Topic',
```
marla-singer: Add i18n tag for placeholder

proxy_backends/collection/schemas/emq.js
```diff
@@ -0,0 +1,84 @@
...
+            value: 'ipaddr',
+          },
+        ];
+      },
+    },
+  },
+  fromValue: {
+    type: String,
+    autoform: {
+      placeholder: 'Value',
```
marla-singer: Add i18n tag for placeholder

proxy_backends/server/methods.js
```diff
@@ -62,4 +68,133 @@ Meteor.methods({
...
+    // Re-construct URI instance without auth credentials
+    uri.normalize();
+
+    // Retrun object with both values
+    return {
+      auth,
+      url: `${uri.valueOf()}`,
+    };
+  },
+  emqAclRequest ({ method, proxyId, rules }) {
```
marla-singer: Why did you wrap arguments? Can you use it  just (method, proxyId, rules)?

frenchbread:
@marla-singer Yes, I've send you details in chat.

proxy_backends/client/form/autoform.js
frenchbread: I don't think so, `_id` is a generic "key" which is usually located in the root of the object/document, while `id` you are pointing at is related to ACL rule's fields that are stored in a list. At the same time, current `id` implementation is used to re-write `id` value on SailsJs side.

proxy_backends/client/form/form.html
frenchbread: Good catch!

frenchbread: I tried your suggestion and it does not work this way since `proxy` is attached to different collection which is not autoform-object.

proxy_backends/client/form/form.js
frenchbread: 👍  Removed.

proxy_backends/collection/schemas/emq.js
frenchbread: Done

proxy_backends/collection/schemas/emq.js
frenchbread: Done

proxy_backends/client/form/form.html
frenchbread: Same as above.

proxy_backends/client/form/autoform.js
```diff
@@ -206,59 +196,107 @@ AutoForm.hooks({
...
+
         // Get success message translation
         const message = TAPi18n.__('proxyBackendForm_successMessage');
 
         // Alert the user of success
         sAlert.success(message);
       }
     },
+    onError (formType, error) {
+      console.error(formType, error);
```
frenchbread: Yep, done

proxy_backends/client/form/autoform.js
```diff
@@ -206,59 +196,107 @@ AutoForm.hooks({
...
+      // Form Type - INSERT
       } else {
+        // Check what proxy backend is selected
+        if (proxyBackend.type === 'emq') {
+          Meteor.call('emqAclRequest', {
+            method: 'POST',
+            proxyId: proxyBackend.proxyId,
+            rules: proxyBackend.emq.settings.acl,
+          }, (err) => {
+            if (err) console.error(err);
```
frenchbread: Done

proxy_backends/client/form/autoform.js
```diff
@@ -206,59 +196,107 @@ AutoForm.hooks({
...
-        );
+          );
+        // **** EMQ *****
+        } else if (proxyBackend.type === 'emq') {
+          Meteor.call('emqAclRequest', {
+            method: 'PUT',
+            proxyId: proxyBackend.proxyId,
+            rules: proxyBackend.emq.settings.acl,
+          }, (err) => {
+            if (err) console.error(err);
```
frenchbread: Done

proxy_backends/client/form/autoform.js
frenchbread: Done

proxy_backends/server/methods.js
frenchbread: Either way is ok, but ES6 params passing is better IMO. It's easier to keep track of what params has been passed to a function & not strictly follow the order in which they are passed. Also, when calling the method, this might look a bit messy: 
```js
Meteor.call('emqAclRequest', 'POST', proxyBackend.proxyId, proxyBackend.emq.settings.acl, (err, res) => { ... });
```

marla-singer:
@frenchbread My test is fail.
I have proxy settings of MQT and couple settings of API Umbrella
joxi_screenshot_1494402686394
I was connection API to MQT proxy when I had an error in browser console:
`autoform.js:298 insert Error: Proxy id is required` in file `/pxoy_backend/client/form/autoform.js:93`
joxi_screenshot_1494403055038

frenchbread:
@marla-singer that's weird.. How many proxies do you have? That might be caused by having multiple Proxies added, for some reason. I'll check on that.

marla-singer:
@frenchbread Three proxies with names "MQT proxy", "Staging proxy", "proxy". The main idea is that MQT proxy is the first in list

frenchbread:
@marla-singer I tested with multiple proxies after `meteor reset` multiple times & was able to submit ACL rules. So cannot reproduce the error.

frenchbread:
I can though add a default option to the select box (with empty value), e.g. 

```html
<select>
  <option value="" >Select proxy</option>
  <!-- .. other options go here .. -->
</select>
```

brylie:
screenshot_20170510_134751

Access control rules fields are misaligned in Firefox.

frenchbread:
@brylie On it.

brylie:
screenshot_20170510_141354

Consider making default selections for the ACL rules. E.g.

- deny
- both

frenchbread:
Cannot reproduce in Firefox. 

screen shot 2017-05-10 at 14 14 49

@brylie What version are you using?

frenchbread:
> Consider making default selections for the ACL rules

@brylie  Why?

brylie:
I was using [this vversion](https://github.com/apinf/platform/pull/2483/commits/570240de4d738961f3ab3b84441ce79521e49f6e). We have pulled the latest 3 commits.

brylie:
screenshot_20170510_142031

Same problem, latest commit **on Firefox**.

marla-singer:
@frenchbread Itis my faul,  I tested onold version

Tested it again on new one and found:
1. After successful adding, I could not delete proxy backend information using button Delete. Nothing happends
2. After navigatiin to Details tab on API profile page I have error in browser console:
```
Exception in template helper: TypeError: Cannot read property 'url' of undefined
 at Object.proxyUrl (http://localhost:3030/app/app.js?hash=ed50ce90e8239866f33e29a69c6edc8314e72d4f:4695:35)
```
Perhaps it needs to update helpers:
https://github.com/apinf/platform/blob/a7846f838e1b2869933d5ef72241444b4dd332bc/apis/client/profile/details/details.js#L36

frenchbread:
@brylie I mean what version of the Firefox browser. I couldn't reproduce in on my side on both Chrome & Firefox

brylie:
We are confused about the meaning of fields, such as *value*. Consider adding inline help to the form, to help the user. Please explain the meaning of the fields to us as well.

frenchbread:
> After successful adding, I could not delete proxy backend information using button Delete. Nothing happends

@marla-singer This issues existed before adding EMQ proxy, if I recall correctly there is an issue for it. Deleting ALC can be done y removing it manually using `-` button & submitting the form again.

frenchbread:
I had a discussion with @bajiat about this, and will create a unified document that describes overall architecture and all the components around it (in day or two). 

For now, the ACL schema on apinf side is used to map data to schema on EMQ which looks like [this](https://github.com/emqtt/emq-auth-pgsql#acl-table)

marla-singer:
@frenchbread More details about "bad markup"
OS: Debian 8
Browser: Firefox 45.8.0

frenchbread:
@brylie There is no such wildcard in MQTT as `*`, there are only `+` & `#` as noted in [here](http://www.hivemq.com/blog/mqtt-essentials-part-5-mqtt-topics-best-practices)

brylie:
> rules should be in specific order, and executed one-by-one from top to bottom.

What does a user do, if they want to change the order of the rules? Can you describe the user flow?

screenshot_20170510_144127

brylie:
> There is no such wildcard in MQTT as *, there are only + & # 

The point is not the syntax. Is it possible for *Tampere City Traffic* to accidentally deny *Helsinki City Traffic* from publishing their bus locations, for example?

frenchbread:
> What does a user do, if they want to change the order of the rules? Can you describe the user flow?

As of now, since meteor autoform does not support drag & drop functionality in array lists, user would have to remove rules & add the in correct order

frenchbread:
> The point is not the syntax. Is it possible for Tampere City Traffic to accidentally deny Helsinki City Traffic from publishing their bus locations, for example?

That's a good question. Gotta ask @phanimahesh about this.

frenchbread:
I was able to reproduce the "bad markup" issue. Would like some help with that. @marla-singer Could you look into this one?

marla-singer:
@frenchbread I can try but I'm not good at HTML

frenchbread:
@marla-singer Just need a fresh look from the side, because IMO, css & html looks correct. Might need some css hack for this.

phanimahesh:
That's a good question, I had this written down but somehow forgot to
reference this in acl PR. Sorry about that.

We have a frontend prefix in api umbrella to prevent path collisions.
Similarly we'll have a prefix for emq per api, and all rules implicitly
will prepend this prefix to topics. That will prevent collisions. This also
means we'll need to track used prefixes and organisations via emq rest api
and underlying postgres store.

Also, can we do something about drag and drop? Having to manually delete
and readd things would be very bad UX.

Nazarah:
> As of now, since meteor autoform does not support drag & drop functionality in array lists, user would have to remove rules & add them in correct order

@frenchbread can we add a controller for each rule panel so that it can be push up or down? Sincemeteor won't be supporting drag and drop, this is the fail-safe solution we can adopt. Otherwise, if we have 5 rules and I wish to make the 5th rule as the default one, then deleting all the 4 rules above and readding them is a really bad usability.

marla-singer:
@brylie @frenchbread I suggest thinking about **which versions of browsers do we need to support?** And write this information in our Guide or something else. 

I searched in the Internet and found [this post ](https://meta.stackexchange.com/questions/56161/which-browsers-are-officially-supported-and-what-else-do-i-need) where the main idea is
>  **"we support the last two versions of the browsers that we see the vast majority of our visitors actually use. This does not include beta/dev releases, which are not supported."**

The last version of Firefox is 53 [(Link to wiki)](https://en.wikipedia.org/wiki/Firefox_version_history). Hence we can test the markup only in 5x version of Firefox and our case is not bug

55:
@brylie @frenchbread @marla-singer 
I can help with markup.

marla-singer:
@55 It is reproduced in 45 Firefox and is not in a new version :confused:

55:
@marla-singer how I get to this view?

marla-singer:
@55 
1. Create a new proxy with type "emq'
2. Go to API profile -> Proxy tab -> Select early added proxy

55:
@marla-singer okay, I am there and able to reproduce it with 53.0.2 on macOS.

frenchbread:
Thanks a lot @55 and @marla-singer. I actually had the same issue in last Firefox v53 as well. Tested, and works on my side.

> can we add a controller for each rule panel so that it can be push up or down

@Nazarah @phanimahesh I think drag and drop would be easier to implement at some point than adding buttons. There is a good [library](https://bevacqua.github.io/dragula/) for it. But unlike in the example, I'd suggest to add a UI button that allow to grad & drop specific rule, NOT the entire <div>.

marla-singer:
@frenchbread I tested again and have questions:
1. After connecting API to emq proxy, Dashboard page contains this API in general list and there is "Fetching latest data.." on screen. As I understand, this kind of API will have own Dashboard page or not?
2. "Get API key" button is appeared for this API and doesn't work.
3. As Enhancement show users message about updating Proxy Backend configuration
4. How to remove saved Proxy Backend configuration? Can you fix Delete button at least for EMQ?

frenchbread:
@marla-singer 

1. Unlikely it will have it's own page, but own component, yes. It's meant that the dashboard should contain some kind of switch (e.g. `<selectbox/>`) between API Umbrella/EMQ template & data contexts since datasets are different for each of them.
2. Ok
3. Good point.
4. Removing an ACL can be done by manually removing it from the list (by clicking "-" btn) and submitting the form again. But that's not really good flow.

Updated def. of done

marla-singer:
@frenchbread 
> Removing an ACL can be done by manually removing it from the list (by clicking "-" btn) and submitting the form again. But that's not really good flow.

if I have just one ACL and click on "-" button then there is the message about "ACL is required field"

frenchbread:
@marla-singer Thanks for reporting this. Might need to remove requirement of this field then.

frenchbread:
I opened issue #2573 for fixing "GET API Button".

frenchbread:
Added suggested changes, please review.

marla-singer:
@frenchbread Cool, deleting for emq works :+1: Alerts as well. I will open some follow up issue and can merge it

brylie:
Does this PR close #2573 for fixing "GET API Button"?

frenchbread:
@brylie No. It was created as a follow up

PR #2490, opened 2017-04-24
Add imports standardization, and other info

mauriciovieira:
* Standardize how to import packages
* Add instructions of how to run eslint locally
* Also add CSS styleguide information that was in the docs repository (DRY principle).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -311,6 +311,8 @@ disgard(trash);
 
 [eslint](https://eslint.org) is configured to run on every commit on Continuous Integration server.
 
+The project is configured with the Airbnb styleguide and meteor recommendation for eslint.
+
 To run locally, do:
 
 ```bash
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -266,10 +306,16 @@ let trash = ["tattered shoe", "broken pencil", "crumpled paper"];
 eat(food);
 disgard(trash);
 ```
+
 ### Code standard(s) and Lint
-Configure your IDE to use eslint with the Airbnb styleguide.
```
frenchbread: - Shouldn't we mention styleguide that is used?

Beyond that, looks good to me.

mauriciovieira: My point is that before adding eslint to the build, the IDE configuration was not enough. It simply did not work - when I added eslint to travis, there was 1800+ lint errors - so I don't think that mentioning IDE is worth here...

I will add a note about the airbnb plugin. Thanks :-)

frenchbread:
Awesome. Merging.

PR #2495, opened 2017-04-25
Feature/organization pinned apis

### LINKED ISSUE #2311: Pinning featured APIs to the top of the API catalog

matleppa:
Closes #2311


apis/client/card/apiCardPin/api_card_pin.html
```diff
@@ -0,0 +1,15 @@
+<template name="apiCardPin">
+  {{# if isApiFeatured }}
```
brylie: Which sounds more natural in a sentence:

- "if is API featured"
- "if API is featured"

apis/client/card/apiCardPin/api_card_pin.js
```diff
@@ -0,0 +1,64 @@
...
+
+// Collection imports
+import Organizations from '/organizations/collection';
+
+Template.apiCardPin.events({
+  'click .pin-button': function () {
+    const instance = this;
+    const organizationId = instance.organization._id;
+    const featuredApiId = instance.api._id;
+    const featuredApiList = instance.organization.featuredApiIds;
```
brylie: Should this be `instance.data.organization`?

apis/client/card/apiCardPin/api_card_pin.js
```diff
@@ -0,0 +1,64 @@
...
+import { sAlert } from 'meteor/juliancwirko:s-alert';
+
+// Collection imports
+import Organizations from '/organizations/collection';
+
+Template.apiCardPin.events({
+  'click .pin-button': function () {
+    const instance = this;
+    const organizationId = instance.organization._id;
+    const featuredApiId = instance.api._id;
```
brylie: Should this be `instance.data.api._id`?

apis/client/card/apiCardPin/api_card_pin.js
```diff
@@ -0,0 +1,64 @@
...
+    const featuredApiList = instance.organization.featuredApiIds;
+    // If API was already featured, remove it from featured list
+    if (Organizations.find({ featuredApiIds: featuredApiId }).count() > 0) {
```
brylie: Try not to do so many things in a single line. Right now, this line does these things:

1. finds organizations
2. counts organizations
3. checks that there are more than zero organizations

Try breaking it in to more lines for readability.

```js
const organizationIsFeatured = Organizations.findOne(featuredApiId);

if (organizationIsFeatured) {
  // do some stuff
}
```

brylie: In general, if you see so many symbols `.`, `{`, `}`, `:`, `>`, `(`, `)`, it might be a sign that the line of code is working too hard.

apis/client/card/apiCardPin/api_card_pin.js
```diff
@@ -0,0 +1,64 @@
...
+        { $push:
+          { featuredApiIds: featuredApiId },
+        }
+      );
+      // Get localized success message
+      const message = TAPi18n.__('apiCardPin_featuredSuccessfully');
+
+      // Display success message to user
+      sAlert.success(message);
+    } else {
```
brylie: Make sure this `else` clause is catching the conditions you are explaining to the user. In other words, is the "featured list already full" whenever we get into this `else` block? Are there any other conditions that might cause us to get into the `else` block?

apis/client/card/apiCardPin/api_card_pin.js
```diff
@@ -0,0 +1,64 @@
...
+        // Display success message to user
+      sAlert.error(message);
+    }
+  },
+});
+
+Template.apiCardPin.helpers({
+  apiIsFeatured () {
+    const instance = this;
+    const featuredApiId = instance.api._id;
```
brylie: This variable could just as well be called `apiId`, since it may or may not be featured.

Also, should it be `instance.data.api._id` rather than `instance.api.Id`?

apis/client/card/apiCardPin/api_card_pin.js
```diff
@@ -0,0 +1,64 @@
...
+    let apiIsFeatured = false;
+    // If API was already featured, remove it from featured list
+    if (Organizations.find({ featuredApiIds: featuredApiId }).count() > 0) {
```
brylie: Consider rewriting as:

```js
// Check if this API has been featured by any organization
const apiIsFeatured = Organizations.findOne({ featuredApiIds: apiId });

if (apiIsFeatured) {
  // do some stuff
}
```

organizations/client/profile/featured_apis/featured_apis.html
```diff
@@ -0,0 +1,24 @@
+<!-- Copyright 2017 Apinf Oy
+This file is covered by the EUPL license.
+You may obtain a copy of the licence at
+https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence-eupl-v11 -->
+
+<template name="organizationFeaturedApis">
+  <!-- Show or not content-->
+  {{# if hasFeaturedApi }}
```
brylie: Move this check in to parent template (organization profile):

```html
<!-- in organization.html -->
{{# if organization.featuredApiIds }}
  {{> organizationFeaturedApis organization=organization }}
{{/ if }}
```

organizations/client/profile/featured_apis/featured_apis.html
```diff
@@ -0,0 +1,24 @@
...
+    <div class="block-header">
+      <!-- Title -->
+      <h3 class="apis-block-title">
+        {{_ 'organizationProfile_bigText_featuredApisCount' }}
+        <small>
+          {{_ 'organizationProfile_smallText_apisCount' count=featuredApisCount }}
+        </small>
+      </h3>
+      {{> apiFilteredBy }}
```
brylie: What does this `apiFilteredBy` template do? Should it be part of the parent template? Is there a different name that is more self-explanatory?

organizations/client/profile/featured_apis/featured_apis.js
```diff
@@ -0,0 +1,45 @@
...
+    return featuredApis;
+  },
+  featuredApisCount () {
+    const featuredApiList = Template.currentData().organization.featuredApiIds;
+
+    const apis = Apis.find({
+      _id: { $in: featuredApiList },
+    }).fetch();
+
+    return apis.length;
```
brylie: Template.currentData().organization.featuredApiIds` should be an array, if it exists. So, you can do something like:

```js
// Get featured API IDs
const featuredApiList = Template.currentData().organization.featuredApiIds;

// Check if featured API IDs are defined
if (featuredApiList) {
  // Return count of featured APIs
  return featuredApiList.length;
}

// Otherwise return zero, no featured API IDs
return 0;
```

Database query is unnecessary.

organizations/client/profile/featured_apis/no_featured_apis.html
```diff
@@ -0,0 +1,15 @@
+<!-- Copyright 2017 Apinf Oy
+This file is covered by the EUPL license.
+You may obtain a copy of the licence at
+https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence-eupl-v11 -->
+
+<template name="organizationNoFeaturedApis">
+  <div class="alert alert-warning" style="display: inline-block;">
+    <h4>
```
brylie: Is there a header level three (`h3`) tag directly above this? Avoid using header tags to make text bigger or smaller. Header tags are semantic, and should be nested in proper order.

organizations/client/profile/apis/apis.js
```diff
@@ -37,23 +37,25 @@ Template.organizationApis.onCreated(function () {
 
   // Watching on changes of managed APIs after connection to/disconnection from
   instance.autorun(() => {
-    // reactive solution to update pagination with template instant data
-    const updatedApisPerPage = Template.currentData().organization.apisPerPage || 10;
+    if (Template.currentData()) {
+      // reactive solution to update pagination with template instant data
+      const updatedApisPerPage = Template.currentData().organization.apisPerPage || 10;
```
brylie: We need to also **make sure `organization` exists before trying to access its properties and methods**.

organizations/client/profile/featured_apis/featured_apis.js
```diff
@@ -15,31 +15,40 @@ import { Template } from 'meteor/templating';
...
-  hasFeaturedApi () {
-    const featuredApiList = Template.currentData().organization.featuredApiIds;
-
-    const apis = Apis.find({
-      _id: { $in: featuredApiList },
-    }).fetch();
-
-    return apis.length;
-  },
+  // featuredApis () {
```
brylie: Remove commented code.

apis/client/card/apiCardPin/api_card_pin.js
matleppa: The instance object here does not contain .data level.

apis/client/card/apiCardPin/api_card_pin.js
matleppa: The instance object here does not contain .data level.

apis/client/card/apiCardPin/api_card_pin.js
brylie: Alright, then it is not probably the template instance.  We are just giving it the name `instance`, but that does not mean it is actually a Template instance. This is an example of how names can help and hinder our understanding of the code.

apis/client/card/apiCardPin/api_card_pin.js
matleppa: In the first branch there are checked, whether the API id is already in featured list (regardless of list length). In case the API id is not found, in second branch it is checked, whether the featured list does not exist or if the length of featured list is not maximum value. In case these conditions are not met, the remaining alternative is that the list exists and it is full.

brylie:
I am getting a console error when navigating to the Organization Profile:

```js
Error: $in needs an array
```

Where are you using the `{{# for ... in ...}}`? Make sure it is always getting an array, even if an empty array.

brylie:
When I comment out the [`organizationFeaturedApis` from the Organization profile.html](https://github.com/apinf/platform/blob/feature/organization_pinned_apis/organizations/client/profile/profile.html#L14), this error goes away.

brylie:
The error likely originates in the `Organizations.featuredApis()` collection helper. On line 104, we are querying the Apis collection where `{_id: {$in: featuredApiList }}`. There are cases in the code where `featuredApiList` may be `undefined`. Two fixes would be good here:

1. make sure `featuredApiList` exists before using it 
  - possibly making sure it is an array
2. add a default value to the `featuredApiIds` field in the Organizations schema, so that it always at least contains an empty array

core/migrations/server/8-add-featured-apis-organization.js
```diff
@@ -0,0 +1,25 @@
...
+You may obtain a copy of the licence at
+https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence-eupl-v11 */
+
+import { Migrations } from 'meteor/percolate:migrations';
+
+import Organizations from '/organizations/collection/';
+
+Migrations.add({
+  version: 8,
+  name: 'Adds featuredApiIds to contain a list of featured APIs',
```
brylie: Perhaps a better explanation would be 'Ensure all Organizations have featuredApiIds array'.

marla-singer: @matleppa Why did you create a new migration step?

matleppa: @marla-singer I created the migration step because of Brylies's advice related to this error "Error: $in needs an array"

apis/client/card/apiCardPin/api_card_pin.js
```diff
@@ -0,0 +1,69 @@
...
+
+Template.apiCardPin.events({
+  'click .pin-button': function () {
+    const instance = this;
+    const organizationId = instance.organization._id;
+    const featuredApiId = instance.api._id;
+    const featuredApiList = instance.organization.featuredApiIds;
+
+    // Check what can be done for API card in question
+    if (Organizations.find({ featuredApiIds: featuredApiId }).count() > 0) {
```
brylie: How many organizations do you expect to find here? If you only expect **one organization at the most**, you can use `findOne`, which will return either
- one document
- `undefined`

In the case of using `findOne` you will simplify the code, since you don't have to count an array length.

apis/client/card/apiCardPin/api_card_pin.js
```diff
@@ -0,0 +1,69 @@
...
+// Meteor packages imports
+import { Template } from 'meteor/templating';
+import { TAPi18n } from 'meteor/tap:i18n';
+import { sAlert } from 'meteor/juliancwirko:s-alert';
+
+// Collection imports
+import Organizations from '/organizations/collection';
+
+Template.apiCardPin.events({
+  'click .pin-button': function () {
```
brylie: Template events take two arguments:

- event
- templateInstance

Since you are attempting to get the Template instance on line 16, you can simply catch the `templateInstance` parameter of the event function. Note, you will need also to refactor the code to use the `templateInstance.data.api` instead of `instance.api`, since the current solution has a semantic error.

PR #2505, opened 2017-05-02
swagger document is not valid

### LINKED ISSUE #2296: "Swagger document is not valid" after updating apiDoc
Early nightly.apinf.io site has the APIs with uploaded Swagger documentation. About month ago we implemented the new version of Documentation Manage dialog which allows adding Swagger documentation both File and URL. 

Now it displays"Swagger document is not valid" with the old correct documentation. I changed the type to File and the Swagger UI was updated after refreshing.

Verified on nigthly.apinf.io (v 0.41.0)
joxi_screenshot_1489590896408

### Suggested reason
It happened because of the incorrect migration. [We have to specify](https://github.com/apinf/platform/blob/develop/core/migrations/server/3-move-api-docs.js#L15-L20) the `type: "file"` as well

### LINKED ISSUE #2532: Swagger documentation not displayed on apinf.io
# Steps to reproduce 

1. Log into apinf.io (as an admin user)
2. Select an API with a Swagger file (e.g. Hello APinf team API)
3. Go to Documentation tab

# Expected result
Swagger documentation is displayed in Swagger UI.

# Actual result
There is a spinner and the Swagger file does not get displayed

# Console error

```
Exception in delivering result of invoking 'parsedDocument': TypeError: Cannot read property 'schemes' of undefined
    at https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:436:14907
    at c._callback (https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:3:7449)
    at c._maybeInvokeCallback (https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:123:9139)
    at c.receiveResult (https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:123:9447)
    at o._livedata_result (https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:123:21587)
    at o (https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:123:6985)
    at https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:123:606
    at Array.forEach (native)
    at Function.A.each.A.forEach (https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:1:865)
    at C.t.socket.onmessage (https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:123:564) ```

### PR DESCRIPTION
**Pull request by jawidahmadi**
close #2296
close #2532

### CODE REVIEW

File: core/migrations/server/7-move-api-docss.js [hunk 1]
```diff
@@ -0,0 +1,28 @@
+import { Migrations } from 'meteor/percolate:migrations';
```
brylie: Rename this file to `7-move-api-docs.js`

core/migrations/server/7-move-api-docss.js
```diff
@@ -10,19 +10,26 @@ Migrations.add({
     // Iterate through Apis collection
     Apis.find().forEach((api) => {
       // Check if fields exist
```
brylie: Update the comments to describe what is going on here. Why are we only checking `api.documentationFileId` in one place and `api.documentation_link` in another. Comments help us to understand what is going on in the code, why it was written a certain way.

core/migrations/server/3-move-api-docs.js
```diff
@@ -9,7 +9,7 @@ import Apis from '/apis/collection';
 import ApiDocs from '/api_docs/collection';
 
 Migrations.add({
-  version: 3,
+  version: 4,
```
marla-singer: Return number 3

marla-singer:
@jawidahmadi 



As you can see the problem is that **apiDoc document already exists** and we don't need to create another one apiDoc document. (But you create it)

There is the problem because **type** is set as **"url"** but file was uploaded as file

core/migrations/server/7-move-api-docs.js
```diff
@@ -20,6 +20,22 @@ Migrations.add({
...
         ApiDocs.insert(apiDoc);
+      } else if (api.documentationFileId) {
+        // New apiDoc object
+        const apiDoc = {
+          apiId: api._id,
+          type: 'file',
+          fileId: api.documentationFileId,
+          submit_methods: api.submit_methods,
+        };
+        ApiDocs.insert(apiDoc);
```
marla-singer: We don't need to create new apiDoc. We need to fix exists apiDoc

core/migrations/server/7-move-api-docs.js
```diff
@@ -0,0 +1,24 @@
...
+import ApiDocs from '/api_docs/collection';
+
+Migrations.add({
+  version: 7,
+  name: 'Move API documentation from Apis to ApiDocs collection',
+  up () {
+    // Iterate through Apis collection
+    Apis.find().forEach((api) => {
+      // Iterate through ApiDocs collection
+      ApiDocs.find().forEach((apiDoc) => {
```
brylie: This could be a lot simpler, I reckon. Why do we loop through *all* ApiDocs, when we are only looking for one? Is there a way to `findOne` ApiDocs that matches the `api.documentationFileId`?

core/migrations/server/8-move-api-docs.js
```diff
@@ -0,0 +1,24 @@
...
+import ApiDocs from '/api_docs/collection';
+
+Migrations.add({
+  version: 8,
+  name: 'Move API documentation from Apis to ApiDocs collection',
+  up () {
+    // Iterate through Apis collection
+    Apis.find().forEach((api) => {
+      // Iterate through ApiDocs collection
+      ApiDocs.find().forEach((apiDoc) => {
```
brylie: This could be a lot simpler, I reckon. Why do we loop through *all* ApiDocs, when we are only looking for one? Is there a way to `findOne` ApiDocs that matches the `api.documentationFileId`?

core/migrations/server/8-move-api-docs.js
```diff
@@ -0,0 +1,22 @@
...
+import Apis from '/apis/collection';
+import ApiDocs from '/api_docs/collection';
+
+Migrations.add({
+  version: 8,
+  name: 'Move API documentation from Apis to ApiDocs collection',
+  up () {
+    // Iterate through Apis collection
+    Apis.find().forEach((api) => {
+      // Iterate through ApiDocs collection
```
brylie: Remember to update the comments to explain what the code is doing.

PR #2506, opened 2017-05-02
Add margin between API page header and navbar

marla-singer:
After implement footer on every page I found that API page header doesn't have margin-top


mauriciovieira:
LGTM

PR #2514, opened 2017-05-03
Feature/analytics code

### LINKED ISSUE #2288: Footer code for platform admin

marla-singer:
Closes #2288

### DIFF
--- a/core/client/footer/footer.js
+++ b/core/client/footer/footer.js
@@ -7,6 +7,29 @@
 import { Template } from 'meteor/templating';
 import { Modal } from 'meteor/peppelg:bootstrap-3-modal';
 
+Template.footer.onCreated(function () {
+  this.autorun(() => {
+    // Get branding document
+    const branding = Template.currentData().branding;
+
+    // Make sure footer code exists
+    if (branding && branding.footerCode) {
+      //
+      FlowRouter.watchPathChange();
+      // Get previously and current route
+      const oldRoute = FlowRouter.current().oldRoute;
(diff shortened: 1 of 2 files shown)


brylie:
After saving my footer code (`console.log('footer code');`), it seems to work correctly on all pages. However, when navigating from the Organizations page to the Add API page, the footer code does not seem to evaluate:

peek 2017-05-03 15-09

brylie:
After discussing with @marla-singer, we agreed to open a follow-up issue to fix some of the edge cases with this feature. Overall, it is working.

philippeluickx:
@brylie did you open a new issue for this?

Navigation does not seem to be an edge case to me. Is it only when navigating to "Add API" that the footer code is not evaluated?

brylie:
> Navigation does not seem to be an edge case to me. Is it only when navigating to "Add API" that the footer code is not evaluated?

@philippeluickx By 'navigating', I meant 'going from X to Y'. It can be read as 'when going from Organizations page to the Add API page, the footer code does not seem to evaluate'. Hence it was agreed that the work could be continued in a follow-up task.

@marla-singer did you open the follow up task for this issue?

marla-singer:
@philippeluickx @brylie 
#2601

PR #2518, opened 2017-05-03
Add padding to all settings pages containers

### LINKED ISSUE #2516: Navbar covers content of Setting pages
The Navbar covers content of Setting pages like Profile, Account, Branding, Settings and Proxies.

- Go to https://nightly.apinf.io and sign in
- Visit https://nightly.apinf.io/settings/profile

selection_151

manzapanza:
Closes #2516

### DIFF
--- a/branding/client/branding.html
+++ b/branding/client/branding.html
@@ -4,7 +4,7 @@ You may obtain a copy of the licence at
 https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence-eupl-v11 -->
 
 <template name="branding">
-  <div class="container">
+  <div class="container padded-container">
     <div class="row">
       {{> settingsMenu }}
 
--- a/core/client/style/global.less
+++ b/core/client/style/global.less
@@ -113,3 +113,7 @@ hr {
     border-color: transparent;
   }
 }
+
+.padded-container{
+  margin-top: 36px;
+}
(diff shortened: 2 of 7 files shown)


marla-singer:
@manzapanza I suggest another way.
1. add only `margin-top: 2em;` syle instead of padding
2. add new class for `row` element not for container

What do you think?

manzapanza:
I analyzed a bit better the problem now. Looking on the other pages, for example Organizations, the top margin between the navbar and the title h1 is obtained by the margin top of the h1 element: 1 em. Because the h1 has a font-size 36px, the calculated margin top is 36px.

I think that is better use as unit px than em because we wont apply this property to block elements (and not inline elements like h1) and the right top value should be 36px. (and not 32px as I did)

I agree with you that margin top is a better choice. Between container and row I supposed that padded-container was more easy to understand than something like first-row or first-row-margin, but ther's not much difference between one or the other.

I placed a bottom padding to solve a problem to the settings page, because doesn't there a main panel, there's not space between the button form save and the footer. 
Much better apply the same panel margin bottom (20px) to the class .settings-fields that wraps all panels on the page.

My time for today is gone, tomorrow I will send an update!

bajiat:
@marla-singer Can I assign you as the reviewer?

marla-singer:
@manzapanza I removed the label "Ready for review". Please add it again when it will be ready
@bajiat Self-assigned

PR #2545, opened 2017-05-10
Endpoints for Apis collection

### LINKED ISSUE #2441: APInf Catalog (APIs) API implementation
# Goal
API consumer can call the **API** endpoints and use the methods defined in the Catalog API design document.

# Related resources
We have produced a Swagger file with Organization API endpoints, which can be accessed on our [APInf deployment](https://apinf.io/apis/TYf33t5B6D6fwamKt).

This is also related to issue #2440

marla-singer:
Closes #2441 
Endpoints for Apis collection

## Known issues
- Swagger file doesn't have `/login` and `/logout` path
- After using POST managerIds doesn't contain ID

### DIFF
--- a/apis/server/api.js
+++ b/apis/server/api.js
@@ -3,12 +3,15 @@
  You may obtain a copy of the licence at
  https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence-eupl-v11 */
 
+// Meteor packages imports
+import { Meteor } from 'meteor/meteor';
+
 // Collection imports
 import Apis from '/apis/collection';
 import ApiV1 from '/core/server/api';
 import Organizations from '/organizations/collection';
 
-// Request /apinf-rest/v1/apis for Apis collection
+// Request /rest/v1/apis for Apis collection
 ApiV1.addCollection(Apis, {
   routeOptions: {
     authRequired: false,
@@ -20,7 +23,7 @@ ApiV1.addCollection(Apis, {
         tags: [
           ApiV1.swagger.tags.api,
         ],
-        description: 'List and search APIs.',
+        description: 'List and search public APIs.',
         parameters: [
           ApiV1.swagger.params.optionalSearch,
           ApiV1.swagger.params.organization,
@@ -30,7 +33,7 @@ ApiV1.addCollection(Apis, {
         ],
         responses: {
           200: {
-            description: 'Returns list of APIs',
+            description: 'Returns list of public APIs',
           },
           400: {
             description: 'Bad query parameters',
@@ -103,16 +106,16 @@ ApiV1.addCollection(Apis, {
         tags: [
           ApiV1.swagger.tags.api,
         ],
-        description: 'Returns result is one or zero (no match found) APIs',
+        description: 'Returns one API with specified ID or nothing if there is not match found',
         parameters: [
           ApiV1.swagger.params.apiId,
         ],
         responses: {
           200: {
-            description: 'Returns API entity',
+            description: 'Returns API',
           },
           404: {
-            description: 'Bad parameters',
+            description: 'Bad parameter',
           },
         },
       },
@@ -123,13 +126,13 @@ ApiV1.addCollection(Apis, {
         tags: [
           ApiV1.swagger.tags.api,
         ],
-        description: 'Adds an API to catalog. On success methods return added object with details.',
+        description: 'Adds an API to catalog. On success, returns newly added API object.',
         parameters: [
           ApiV1.swagger.params.api,
         ],
         responses: {
           200: {
-            description: 'API was successful added',
+            description: 'API successfully added',
           },
           401: {
             description: 'Authentication is required',
@@ -143,8 +146,7 @@ ApiV1.addCollection(Apis, {
         ],
       },
     },
-    // Partially modify the entity with the given :id with the data contained in the request body.
-    // Only fields included will be modified.
+    // Modify the entity with the given :id with the data contained in the request body.
     put: {
       authRequired: true,
       // manager role is required. If a user already has an API then the user has manager role
@@ -153,20 +155,23 @@ ApiV1.addCollection(Apis, {
         tags: [
           ApiV1.swagger.tags.api,
         ],
-        description: 'This is a post route to post things. Updates API to the system.',
+        description: 'Update an API',
         parameters: [
           ApiV1.swagger.params.apiId,
           ApiV1.swagger.params.api,
         ],
         responses: {
           200: {
-            description: 'API was successful edited.',
+            description: 'API successfully edited.',
           },
           401: {
             description: 'Authentication is required',
           },
+          403: {
+            description: 'User does not have permission',
+          },
           404: {
-            description: 'API is not found with specified ID',
+            description: 'API is not found',
           },
         },
         security: [
@@ -181,26 +186,39 @@ ApiV1.addCollection(Apis, {
         const bodyParams = this.bodyParams;
         // Get ID of API
         const apiId = this.urlParams.id;
+        const userId = this.userId;
         const api = Apis.findOne(apiId);
 
-        // Make sure API exists
-        if (api) {
+        // Make sure API exists & user can manage
+        if (api && api.managerIds.indexOf(userId) > 0) {
           // Update API document
           Apis.update(apiId, { $set: bodyParams });
 
           return {
             statusCode: 200,
             body: {
-              status: 'Success updating 12345',
+              status: 'Success updating',
               data: Apis.findOne(apiId),
             },
           };
         }
 
+        // Make sure API exists but user can not manage
+        if (api && api.managerIds.indexOf(userId) === -1) {
+          return {
+            statusCode: 403,
+            body: {
+              status: 'Fail',
+              message: 'You do not have permission for editing this API',
+            },
+          };
+        }
+
+        // API doesn't exist
         return {
           statusCode: 404,
           body: {
-            status: 'fail',
+            status: 'Fail',
             message: 'API is not found with specified ID',
           },
         };
@@ -220,10 +238,16 @@ ApiV1.addCollection(Apis, {
         ],
         responses: {
           200: {
-            description: 'API was successful removed.',
+            description: 'API successfully removed.',
           },
           401: {
             description: 'Authentication is required',
+          },
+          403: {
+            description: 'User does not have permission',
+          },
+          404: {
+            description: 'API is not found',
           },
         },
         security: [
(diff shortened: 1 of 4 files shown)


apis/server/api.js
```diff
@@ -1,98 +1,98 @@
...
-
-// Npm packages imports
-import _ from 'lodash';
+ This file is covered by the EUPL license.
+ You may obtain a copy of the licence at
+ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence-eupl-v11 */
 
 // Collection imports
 import Apis from '/apis/collection';
+import Organizations from '/organizations/collection';
```
mauriciovieira: Move to the last line so that we have imports sorted.

marla-singer: @mauriciovieira Ok

core/server/api.js
```diff
@@ -10,7 +10,7 @@ import { Meteor } from 'meteor/meteor';
 import { Restivus } from 'meteor/nimble:restivus';
 
 const ApiV1 = new Restivus({
-  apiPath: 'api',
+  apiPath: 'apinf-rest',
```
mauriciovieira: I wonder if this path should be a configuration from Branding. This way we whitelabel also the rest-api.

apis/server/api.js
mauriciovieira: Sort the imports.  Organizations would go after ApiV1

core/server/api.js
mauriciovieira: I wonder if this _path_ should be a Branding configuration. This way we whitelabel also the rest-api.

@kyyberi will we need to have a configurable API path? `GET /apinf-rest/v1/users/:id` for APInf installation, and `GET /my-org-rest/v1/users/:id` for a my-org customer, for instance.

mauriciovieira: @marla-singer I talked to @kyyberi and I think it's fine, and this path can be hardcoded... Anyways, we can change later :-)

apis/server/api.js
```diff
@@ -1,97 +1,237 @@
...
+          ApiV1.swagger.params.organization,
+          ApiV1.swagger.params.skip,
+          ApiV1.swagger.params.limit,
+          ApiV1.swagger.params.lifecycle,
         ],
-        description: 'Returns all APIs.',
         responses: {
           200: {
-            description: 'List of APIs.',
+            description: 'Returns list of APIs',
```
brylie: Should this say 'Returns list of *public* APIs'?

apis/server/api.js
```diff
@@ -1,97 +1,237 @@
...
+      },
+    },
+    // Return the entity with the given :id
+    get: {
+      authRequired: false,
+      swagger: {
+        tags: [
+          ApiV1.swagger.tags.api,
+        ],
+        description: 'Returns result is one or zero (no match found) APIs',
```
brylie: This description is confusing. The code comment is easier to understand. Could you explain a bit more here?

apis/server/api.js
```diff
@@ -1,97 +1,237 @@
...
+        },
+      },
+    },
+    post: {
+      authRequired: true,
+      swagger: {
+        tags: [
+          ApiV1.swagger.tags.api,
+        ],
+        description: 'Adds an API to catalog. On success methods return added object with details.',
```
brylie: Consider describing as:

> Adds an API to catalog. On success, returns newly added API object.

apis/server/api.js
```diff
@@ -1,97 +1,237 @@
...
+        parameters: [
+          ApiV1.swagger.params.api,
+        ],
+        responses: {
+          200: {
+            description: 'API was successful added',
```
brylie: Change to:

> API successfully added

apis/server/api.js
```diff
@@ -1,97 +1,237 @@
...
+          },
+        ],
+      },
+    },
+    // Partially modify the entity with the given :id with the data contained in the request body.
+    // Only fields included will be modified.
+    put: {
+      authRequired: true,
+      // manager role is required. If a user already has an API then the user has manager role
+      roleRequired: ['manager', 'admin'],
```
brylie: What is the reason to include 'manager' role here?

apis/server/api.js
```diff
@@ -1,97 +1,237 @@
...
+      swagger: {
+        tags: [
+          ApiV1.swagger.tags.api,
+        ],
+        description: 'This is a post route to post things. Updates API to the system.',
```
brylie: Change to

> Update an API

... or something similar.

apis/server/api.js
```diff
@@ -1,97 +1,237 @@
...
+        parameters: [
+          ApiV1.swagger.params.apiId,
+          ApiV1.swagger.params.api,
+        ],
+        responses: {
+          200: {
+            description: 'API was successful edited.',
```
brylie: Successful -> successful**ly**

apis/server/api.js
```diff
@@ -1,97 +1,237 @@
...
+        if (api) {
+          // Update API document
+          Apis.update(apiId, { $set: bodyParams });
+
+          return {
             statusCode: 200,
             body: {
-              status: 'success',
-              data: allApis,
+              status: 'Success updating 12345',
```
brylie: 12345?

apis/server/api.js
```diff
@@ -1,97 +1,237 @@
...
+            message: 'API is not found with specified ID',
+          },
+        };
       },
     },
-    get: {
+    delete: {
+      authRequired: true,
+      // manager role is required. If a user already has an API then the user has manager role
+      roleRequired: ['manager', 'admin'],
```
brylie: Why do we check for manager role here? What does it accomplish?

apis/server/api.js
```diff
@@ -1,97 +1,237 @@
...
         ],
-        description: 'Returns one API with given ID.',
+        description: 'Deletes the identified API from the system.',
         parameters: [
           ApiV1.swagger.params.apiId,
         ],
         responses: {
           200: {
-            description: 'One API.',
+            description: 'API was successful removed.',
```
brylie: Successful -> successful**ly**

core/server/api.js
brylie: Yes, just remove the 'apinf' part of the path.

core/server/api.js
```diff
@@ -20,41 +20,126 @@ const ApiV1 = new Restivus({
...
   tags: {
-    apis: 'Apis',
+    api: 'APIs',
   },
   params: {
     apiId: {
       name: 'id',
       in: 'path',
-      description: 'Api ID',
+      description: 'Pass ID of the API',
```
brylie: Change to 

> ID of API

No verb ('pass') is necessary.

core/server/api.js
```diff
@@ -20,41 +20,126 @@ const ApiV1 = new Restivus({
...
       required: true,
       type: 'string',
     },
+    optionalSearch: {
+      name: 'q',
+      in: 'query',
+      description: 'Pass an optional search string for looking up inventory.',
```
brylie: Change to:

> optional search string for finding APIs.

core/server/api.js
```diff
@@ -20,41 +20,126 @@ const ApiV1 = new Restivus({
...
-  // Generates: POST on /api/v1/users and GET, DELETE /api/v1/users/:id for
   // Meteor.users collection
   ApiV1.addCollection(Meteor.users, {
-    excludedEndpoints: ['getAll', 'put'],
+    excludedEndpoints: ['getAll', 'put', 'patch'],
     routeOptions: {
       authRequired: true,
     },
     endpoints: {
+      // GET /apinf-rest/v1/users/:id
```
brylie: Remember to update all of these comments if changing base path (removing `apinf` from base path).

apis/server/api.js
marla-singer: Look comment above
>If a user already has an API then the user has manager role

apis/server/api.js
marla-singer: Look comment above
If a user already has an API then the user has manager role

apis/server/api.js
brylie: Right, but what do we achieve from using the `manager` role here?

core/server/api.js
brylie: > I wonder if this path should be a Branding configuration.

Also, as a general guideline **less is more**. When possible, we should find simple, generic solutions, without adding more options to the user interface. This impacts both **maintainability** and **usability**.

apis/server/api.js
```diff
@@ -181,26 +186,39 @@ ApiV1.addCollection(Apis, {
         const bodyParams = this.bodyParams;
         // Get ID of API
         const apiId = this.urlParams.id;
+        const userId = this.userId;
         const api = Apis.findOne(apiId);
 
-        // Make sure API exists
-        if (api) {
+        // Make sure API exists & user can manage
+        if (api && api.managerIds.indexOf(userId) > 0) {
```
brylie: Can we use the `api.currentUserCanManage()` helper here?

apis/server/api.js
```diff
@@ -1,97 +1,303 @@
...
+        tags: [
+          ApiV1.swagger.tags.api,
+        ],
+        description: 'Returns one API with specified ID or nothing if there is not match found',
+        parameters: [
+          ApiV1.swagger.params.apiId,
+        ],
+        responses: {
+          200: {
+            description: 'Returns API entity',
```
brylie: Maybe we don't need to say 'entity' here.

apis/server/api.js
```diff
@@ -1,97 +1,303 @@
...
+          200: {
+            description: 'API successfully added',
+          },
+          401: {
+            description: 'Authentication is required',
+          },
+        },
+        security: [
+          {
+            userSecurityToken: [],
```
brylie: What do these do?

apis/server/api.js
```diff
@@ -1,97 +1,303 @@
...
+      action () {
+        // Get data from body parameters
+        const bodyParams = this.bodyParams;
+        // Get ID of API
+        const apiId = this.urlParams.id;
+        const userId = this.userId;
+        const api = Apis.findOne(apiId);
+
+        // Make sure API exists & user can manage
+        if (api && api.managerIds.indexOf(userId) > 0) {
```
brylie: Same as above. Lets re-use the `api.currentUserCanManage()` helper, since it does this same thing in a re-usable and self-describing manner.

apis/server/api.js
marla-singer: Do authorization. Read docs about creation Swagger specification

apis/server/api.js
marla-singer: @brylie Not, we can not
```
Error: Meteor.userId can only be invoked in method calls. Use this.userId in publish functions.
    at AccountsServer.userId (packages/accounts-base/accounts_server.js:82:13)
    at Object.Meteor.userId (packages/accounts-base/accounts_common.js:252:19)
    at Document.Apis.helpers.currentUserCanManage (apis/collection/helpers.js:33:27)
    at Object.ApiV1.addCollection.endpoints.put.action (apis/server/api.js:193:24)
    at Route.share.Route.Route._callEndpoint (packages/nimble_restivus/lib/route.coffee:149:25)
    at packages/nimble_restivus/lib/route.coffee:59:33
    at packages/simple_json-routes.js:98:9
```

marla-singer:
@brylie Ready

apis/server/api.js
brylie: Ah, I thought they were related to Restivus. Also, it is helpful to provide links to related documentation.

marla-singer: @brylie http://swagger.io/specification/#securityRequirementObject

Issue #2528, opened 2017-05-04
apinf.io dashboard empty


bajiat:
# Steps to reproduce
1. Log into apinf.io (as admin user)
2. Go to dashboard

# Result

<img width="908" alt="production dashboard empty 0 43 0" src="https://cloud.githubusercontent.com/assets/11626388/25701431/0a44930c-30d5-11e7-9e88-33bade3d76b5.png">

Console error:
```
TypeError: Cannot read property 'name' of undefined
  at e.apiName (https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:449:24269)
  at https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:448:22152
  at h (https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:17:22322)
  at Function.ys [as map] (https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:18:31451)
  at Object.proxyBackends (https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:448:22115)
  at https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:153:31360
  at https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:153:17194
  at https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:153:31993
  at Function.e._withTemplateInstanceFunc (https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:154:5707)
  at https://apinf.io/bb2540909451ddaee73743b11b0dac9549b8624d.js?meteor_js_resource=true:153:31900
bb25409….js?meteor_js_resource=true:3 Connection timeout. No sockjs heartbeat received. 
```

marla-singer:
@bajiat I updated console message, left the only one which related to problem

bajiat:
If I log into apinf.io as a non-admin, I see the dashboard:

<img width="925" alt="production dashboard as non-admin 0 43 0" src="https://cloud.githubusercontent.com/assets/11626388/25701657/f40ae252-30d5-11e7-954d-4dfcfa088da5.png">

[...]

PR #2540, opened 2017-05-09
Ensure API document is available before returning name

brylie:
Closes #2528 

Note, we will need to make a new patch-level release for this hotfix.

PR #2577, opened 2017-05-17
Hotfix/fix lint

brylie:
Don't hit me :fearful:

### DIFF
--- a/proxy_backends/collection/helpers.js
+++ b/proxy_backends/collection/helpers.js
@@ -15,7 +15,7 @@ ProxyBackends.helpers({
     const api = Apis.findOne(apiId);
 
     // placeholder for API name
-    const apiName;
+    let apiName;
 
     // Make sure API was found before accessing name property
     if (api) {
--- a/proxy_backends/collection/helpers.js
+++ b/proxy_backends/collection/helpers.js
@@ -14,10 +14,15 @@ ProxyBackends.helpers({
     // Get API
     const api = Apis.findOne(apiId);
 
+    // placeholder for API name
+    const apiName;
+
     // Make sure API was found before accessing name property
     if (api) {
-      // Return API Name
-      return api.name;
+      // Set API Name from API
+      apiName = api.name;
     }
+
+    return apiName;
   },
 });


marla-singer:
@brylie It doesn't work. Now if list contains proxy backend with unexisted API then there is an empty place
joxi_screenshot_1495031201568

brylie:
We need to clean up that proxy backend. The original fix was a band-aid, we discussed this multiple times.

marla-singer:
> We need to clean up that proxy backend. Please refer to the discussion in the other PR, where we tried doing due diligence and decided to go with the basic fix.
This PR is a lint fix.

But the basic hotfix doesn't work

brylie:
Well, it does work, but there may be corrupted data on the server. How did you test?

marla-singer:
> How did you test?
@brylie Create Proxy Backend with non-existent API ID. The same case as apinf.io has

mauriciovieira:
@marla-singer The logic is unaltered. So this empty name should happen also before this hotfix.

brylie:
> Create Proxy Backend with non-existent API ID. The same case as apinf.io has

Steps to reproduce?

Remember, there were three of us [discussing this issue](https://github.com/apinf/platform/issues/2528). We found out there were [migration problems](https://github.com/apinf/platform/issues/2528#issuecomment-300409079). I tested the [cascading delete](https://github.com/apinf/platform/issues/2528#issuecomment-300710154) successfully.

It [might just be a botched migration that caused inconsistent data](https://github.com/apinf/platform/issues/2528#issuecomment-300711791), but the patch will prevent the originally reported issue of the Dashboard being blank.

marla-singer:
@mauriciovieira I missed this bug in this PR https://github.com/apinf/platform/pull/2540

proxy_backends/collection/helpers.js
```diff
@@ -14,10 +14,15 @@ ProxyBackends.helpers({
...
+
     // Make sure API was found before accessing name property
     if (api) {
-      // Return API Name
-      return api.name;
+      // Set API Name from API
+      apiName = api.name;
     }
+
+    return apiName;
```
mauriciovieira: I mean, @marla-singer, before this `return` here, it was returning `undefined` anyways.

marla-singer:
@brylie But now apinf.io contains Proxy Backend document with non-existent API ID. And because of that Dashboard page is empty

brylie:
In the case of the non-existant API, the data needs to be cleaned up. This hotfix did not address that issue, by design. We can continue the work in a different pull request, but will need steps to reproduce the data inconsistency other than manually creating a proxy backend with malformed API ID.

brylie:
> But now apinf.io contains Proxy Backend document with non-existent API ID. And because of that Dashboard page is empty

The whole dashboard is empty?

mauriciovieira:
Ah, I see https://github.com/apinf/platform/pull/2540/commits was not checked by travis.

proxy_backends/collection/helpers.js
brylie: What should this code block look like @mauriciovieira?

brylie:
Jep, and I wasn't reading my lint messages closesly. My bad.

marla-singer:
@mauriciovieira What do you think? This PR fixes ESlint error but it seems that found bug can repeat on apinf.io
joxi_screenshot_1495031201568

mauriciovieira:
@brylie, @marla-singer I think that if the problem of #2528 is still there, even after #2540, we must not amend things. We must revert https://github.com/apinf/platform/commit/c3cb04a44945ff588e1a39b927c4a1aefeee1f29

mauriciovieira:
But if #2540 really closes #2528, this hotfix is appropriate here.

brylie:
> I think that if the problem of #2528 is still there

Based on what? How can we reproduce?

Our strongest hypothesis so far is that it was a failed migration, please review the comments:

- https://github.com/apinf/platform/issues/2528#issuecomment-299157381
- https://github.com/apinf/platform/issues/2528#issuecomment-300137672'
- https://github.com/apinf/platform/issues/2528#issuecomment-300409079

mauriciovieira:
@brylie I see your point. @marla-singer I agree with @brylie.

brylie:
> But if #2540 really closes #2528, this hotfix is appropriate here.

The problem we face is that we have no clear answer as to how the data became inconsistent. We have discussed this, double checked that the ProxyBackend (and other related data) gets deleted when the API is deleted, and issued a band-aid fix to make sure we don't try to access the *name* property of `undefined`.

> @brylie I see your point. @marla-singer I agree with @brylie.

Thank you. I am not trying to be stubborn here. We can definitely continue researching this issue, but it is just not clear to me what to do differently.

@marla-singer thank you very much for being so thorough with your reviews :-)

mauriciovieira:
> The problem we face is that we have no clear answer as to how the data became inconsistent

I wonder if it's related to migration changing https://github.com/apinf/platform/pull/2431#discussion_r110453761

mauriciovieira:
I did merge master back to develop https://github.com/apinf/platform/commit/1eaa1ede41599886601471ad7f7cd802120732a6 after this hotfix was applied to master

PR #2597, opened 2017-05-19
Hotfix/hide get api key

### LINKED ISSUE #2573: Hide "Get API Key" button when EMQ is selected as a proxy for API

frenchbread:
Closes #2573 

### Changes
- Fixed indentation
- Added template helper
- Added check to template

### DIFF
--- a/api_keys/client/api_key.js
+++ b/api_keys/client/api_key.js
@@ -111,6 +111,15 @@ Template.apiKey.helpers({
     // Get proxyBackend from template data
     const proxyBackend = Template.currentData().proxyBackend;
 
-    return proxyBackend.type === 'apiUmbrella';
+    // Check is proxy backend exists and contains type field
+    if (proxyBackend && proxyBackend.type) {
+      // Check if proxy backend type is "apiUmbrella"
+      const proxyBackendIsApiUmbrella = proxyBackend.type === 'apiUmbrella';
+
+      return proxyBackendIsApiUmbrella;
+    }
+
+    // Don't show "Get API Key" button if proxtBackend is undefined
+    return false;
   },
 });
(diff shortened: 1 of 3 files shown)


frenchbread:
Please review 👀

api_keys/client/api_key.js
```diff
@@ -107,4 +107,10 @@ Template.apiKey.helpers({
 
     return apiKey;
   },
+  showButton () {
+    // Get proxyBackend from template data
+    const proxyBackend = Template.currentData().proxyBackend;
+
+    return proxyBackend.type === 'apiUmbrella';
```
brylie: What happens if `proxyBackend` is `undefined`?

frenchbread: Currently both proxies are set in a way to contain either "apiUmbrella" or "emq" proxy backend type. And `type` is required field.

brylie: Would it be possible for `proxyBackend` to be `undefined`?

frenchbread: I can't think of any scenario when `proxyBackend` is `undefined`

brylie: Alright, we just get a bug about 'cannot access *property* of *undefined*' about once per sprint.

brylie: So, just make sure to check for object before using a property. For safety.

brylie: You can, for example, use the Meteor check() package, or just an `if (thing && thing.property) {// access property }`

frenchbread: Ok, on it.

frenchbread: `undefined` might have occurred when API is added and `proxyBackend` has not been configured yet. I've added check that hides the button if no proxy is attached to API.

brylie:
OK, please resolve conflicts.

frenchbread:
Done 👍

PR #2600, opened 2017-05-23
Emq users

### LINKED ISSUE #2599: Add UI for EMQ Users management via EMQ-REST-API
### Def. of done
- [x] Add emq-user
- [x] Update emq-user
- [x] Remove emq-user
- [x] i18n
- [x] comments

frenchbread:
Closes #2599 

`WIP`


bajiat:
@phanimahesh Could you mentor and review?

frenchbread:
Ready for review.

proxies/collection/schemas/emq.js
```diff
@@ -35,7 +35,7 @@ export default {
   },
   'emq.httpApi': {
     type: String,
-    regEx: SimpleSchema.RegEx.Url,
+    // regEx: SimpleSchema.RegEx.Url,
```
mauriciovieira: Should be this removed?
If so, remove also the import of SimpleSchema package. This is currently breaking lint.

proxy_backends/client/form/proxies/emq/users/add/add.js
```diff
@@ -0,0 +1,58 @@
+// Meteor packages imports
```
mauriciovieira: Add copyright notice.

proxy_backends/client/form/proxies/emq/users/add/add.js
```diff
@@ -0,0 +1,58 @@
...
+import { Meteor } from 'meteor/meteor';
+import { Template } from 'meteor/templating';
+
+// Meteor contributed packages imports
+import { Modal } from 'meteor/peppelg:bootstrap-3-modal';
```
mauriciovieira: sort these 3 lines

proxy_backends/client/form/proxies/emq/users/remove/remove.js
```diff
@@ -0,0 +1,53 @@
+// Meteor packages imports
```
mauriciovieira: Add copyright notice.

proxy_backends/client/form/proxies/emq/users/remove/remove.js
```diff
@@ -0,0 +1,53 @@
...
+import { Meteor } from 'meteor/meteor';
+import { Template } from 'meteor/templating';
+
+// Meteor contributed packages imports
+import { Modal } from 'meteor/peppelg:bootstrap-3-modal';
```
mauriciovieira: sort these 3 lines

proxy_backends/client/form/proxies/emq/users/users.js
```diff
@@ -0,0 +1,87 @@
+// Meteor packages imports
```
mauriciovieira: Add copyright notice

proxy_backends/client/form/proxies/emq/users/users.js
```diff
@@ -0,0 +1,87 @@
...
+import { Meteor } from 'meteor/meteor';
+import { Template } from 'meteor/templating';
+import { ReactiveVar } from 'meteor/reactive-var';
+
+// Meteor contributed packages imports
+import { Modal } from 'meteor/peppelg:bootstrap-3-modal';
+import { sAlert } from 'meteor/juliancwirko:s-alert';
+
+// APINF imports
```
mauriciovieira: This comment should be 'Collection imports'

proxy_backends/server/emq.js
```diff
@@ -0,0 +1,112 @@
+// Meteor packages imports
```
mauriciovieira: Add copyright notice.

proxy_backends/client/form/proxies/emq/users/add/add.js
frenchbread: Why and how? If sorting is needed here, it should be added in contributing guidelines.

proxy_backends/client/form/proxies/emq/users/add/add.js
frenchbread: Done

proxy_backends/client/form/proxies/emq/users/remove/remove.js
frenchbread: Done

proxy_backends/client/form/proxies/emq/users/users.js
```diff
@@ -0,0 +1,87 @@
...
+  'click #add-emq-user-modal': function (event, templateInstance) {
+    // Prevent default form submit
+    event.preventDefault();
+
+    // Get proxy data from reactive variable
+    const emqProxy = templateInstance.emqProxy.get();
+    // Get reactive variable instance that stores EMQ users
+    const emqUsersReactive = templateInstance.emqUsers;
+
+    // Show moda, and pass data to it
```
frenchbread: Done

proxy_backends/server/emq.js
frenchbread: Done

proxy_backends/client/form/proxies/emq/users/add/add.js
frenchbread: Done

proxy_backends/client/form/proxies/emq/users/users.js
frenchbread: Done

Nazarah:
A few **questions** from the UX review meeting of this PR:
1. Can we show an example value of "Topic" as disabled text in the related field?
2. Is there any way to know which users are already under EMQ? This is related when we are adding specific username and password under a specific ACL under EMQ proxy in API profile.
3. What is the need of keeping masked password under EMQ user table? Can we hide them and have a control (e.g. button) to show them when needed?
(same goes for id column)
4. Is it safe to list all EMQ users (and topic) under API profile proxy tab? If multiple organizations are using the same API, then organization managers would be able to see user credentials of other organizations apart from their own. This may cause issues with security.
5. Do we need to add an Email field in the Add User modal to let users know about their credentials?
(part of this question is that can we use random password generation than API owner giving passwords to added users?)

**Feedback** from the UX meeting

6. In the PR branch, there were some delay in saving an API proxy configuration from profile (or it might not have saved at all). So saving proxy configuration should be faster.
7. Buttons: Save and Publish, Cancel should appear after EMQ users table
8. When using username attribute on creating an ACL. adding users should be made using auto-complete.
Suggestion of username should appear from the EMQ Users table.

@bajiat can you please see if I missed out something?

proxy_backends/client/form/proxies/emq/users/add/add.js
mauriciovieira: Sorting is not needed but is aesthetically more beautiful. Just my opinion, not a rule. Whenever you ask a compiled language IDE to organize imports, it should also sort them. I will add something to the contributing guidelines so that we have a standard :-)

frenchbread:
1 - Not really sure how's it connected to EMQ users.
2 - By default, the table lists all existing users on EMQ (does not have any filter though, fetches all the records), might need to attach an `api_id` to emq_user schema on emq, to be able to see uses for specific API.
3 - Agreed, `password` is not needed there. `id` might be needed for question 4
4 - Good point.
5 - Agreed
6 - There will be delay due to network latency, for example. `or it might not have saved at all` - sounds like a bug. What are the exact steps for reproducing this? And is it related to EMQ users or ACLs?
7 - This is because EMQ users & emq ACLs are implemented differently, EMQ ACLs works through meteor autoform while EMQ Users works directly with `emq-rest-api`, so `Buttons: Save and Publish, Cancel` are related to EMQ ACLs only.
8 - Good point, but calls for an enhancement feature, rather than implementing it right in this PR.

Nazarah:
@frenchbread @phanimahesh went through your comments in both this PR and rocketchat.
5 - is required when we start using actual users apart from the dummy ones. 
6 - was reviewing the UI from a specific dev branch, so can't produce it in nightly. The reproduction is pretty simple. Simply add an API and configure its proxy from the profile and try to save it. See if it takes a while for the configuration to be saved or it doesn't get saved at all. (no S alert appearing)
7 - I'll go with your logic here
8 - Can I get a screenshot of proxy configuration from API profile for EMQ, with some users added under the ACL? Then  i can open an enhancement issue.

frenchbread:
@Nazarah 

> is required when we start using actual users apart from the dummy ones

`emq-users` are different from `apinf` users and will not have emails, @phanimahesh correct me if I am wrong

proxy_backends/client/form/proxies/emq/users/add/add.js
```diff
@@ -0,0 +1,63 @@
...
+    const emqProxy = templateInstance.data.emqProxy;
+    // Get EMQ HTTP API URL from EMQ Proxy
+    const emqHttpApi = emqProxy.emq.httpApi;
+
+    // Get username and password values
+    const username = $('#emq-user-username').val();
+    const password = $('#emq-user-password').val();
+
+    // Check if values are present
+    if (username && password) {
```
frenchbread: 👍  Done

frenchbread:
Here is a screenshot:

screen shot 2017-05-31 at 19 30 11

proxy_backends/client/form/proxies/emq/users/add/add.html
```diff
@@ -0,0 +1,39 @@
+<template name="addEmqUser">
+  <div class="modal fade">
+    <div class="modal-dialog">
+      <div class="modal-content">
+
+        <div class="modal-header">
+          <h4 class="modal-title">{{_ "proxyBackendForm_emqProxyForm_addEmqUser_modalTitle" }}</h4>
```
brylie: Place nested elements on own line, for readability. This allows a reviewer or contributor to scan the code vertically instead of deciphering a lengthy line of horizontal text.

proxy_backends/client/form/proxies/emq/users/add/add.html
```diff
@@ -0,0 +1,39 @@
...
+        </div>
+
+        <form class="form" id="add-emq-user-form">
+          <div class="modal-body">
+              <div class="form-group">
+                <label for="emq-user-username">{{_ "proxyBackendForm_emqProxyForm_addEmqUser_username" }}</label>
```
brylie: Place nested elements on own line, for readability. This allows a reviewer or contributor to scan the code vertically instead of deciphering a lengthy line of horizontal text.

proxy_backends/client/form/proxies/emq/users/add/add.html
```diff
@@ -0,0 +1,39 @@
...
+                <input id="emq-user-username" type="text" class="form-control" />
+              </div>
+              <div class="form-group">
+                <label for="emq-user-password">{{_ "proxyBackendForm_emqProxyForm_addEmqUser_password" }}</label>
```
brylie: Place nested elements on own line, for readability.

proxy_backends/client/form/proxies/emq/users/remove/remove.html
```diff
@@ -0,0 +1,32 @@
+<template name="removeEmqUser">
+  <div class="modal fade">
+    <div class="modal-dialog">
+      <div class="modal-content">
+
+        <div class="modal-header">
+          <h4 class="modal-title">{{_ "proxyBackendForm_emqProxyForm_removeEmqUser_modalTitle" }}</h4>
```
brylie: Place nested elements on own line, for readability.

brylie:
On a superficial level, the code here looks good. It is readable and makes sense if the assumption is that there will be a unique EMQ instance for each organization.

Given that we are aiming to support multi-tenancy, I am hesitant to merge the current solution. Publishing all EMQ users to every API Owner could easily lead to accidents or even abuse.

We can merge the solution as it is, but my intuition says we should think it through a bit more.

brylie:
One first attempt at an alternative solution would be to create an EMQ users collection in APInf. That collection might have the following fields (we can explore possible relationships):

- `proxyId` - ID of an EMQ proxy in the `Proxies` collection
- `apiID` - ID of an API in the `Apis` collection
- `managerId` - ID of an internal Meteor user who can manage this EMQ User
- `emqUser` - Object containing user details from the *remote EMQ Users collection*

screenshot_20170608_145508


With the above collection/schema in place, we would get the following benefits:

- code simplicity
  - fewer reactive variables to manage the array of EMQ users
    - instead, we rely on the Meteor publication and subscription of related documents (giving us reactivity by default)
- We could scope the EMQ users, so we only display users related to a specific API via `_id`

Some things to iron out would be how to ensure uniqueness of user details such as `email` or `username`. It is likely that we could use **Autoform hooks** (e.g. `onSubmit`) to check whether an EMQ user has already been created with a given `username`.

In general, the above details are very common in our project, so would be familiar to our developers. We work hard to keep our code consistent and readable, so following common design patterns is a helpful practice.

frenchbread:
@brylie thanks for your comments

> we intend to support multi-tenancy

Agreed, but unique EMQ instance for each organization is needed anyways since EMQ does not support "grouping by organization" feature when using only one instance (as far as I understand).

> create an EMQ users collection in APInf

Do not agree here. EMQ ACLs is currently using something similar to your suggestion and it causes some issues with flexibility while completely relying on **autoform**, such as re-ordering, emq<->apinf ACLs syncing, etc.

The current approach for EMQ users here is done differently, comparing to EMQ ACLs, and fetches items directly from  EMQ-REST-API (without APInf collection). This basically removes the need to sync items using a cron process, for example. However, in order to differentiate and "map" EMQ users to organization or API, we can add fields (as you suggesting) such as `apiID`, `managerId` and `proxyId`.

-----

My suggestion is to follow this approach and refactor EMQ ACLs implementation so that autoform is not used there. This way, it would be easy to add drag-and-drop functionality to re-order items, and correctly keep track of ACL rules for specific API.

frenchbread:
Pushed changes.

brylie:
> unique EMQ instance for each organization is needed anyways since EMQ does not support "grouping by organization"

By way of example, we don't require a unique API Umbrella instance per organization, although admittedly API keys are somewhat different than EMQ users.

In any case, we want to also support multi-tenancy for EMQ. This means that the APInf Platform will contain necessary logic to group by organization, regardless of whether EMQ has that functionality. In other words, **we can use the APInf collections and permissions to allow EMQ Users to be grouped by proxy, proxy backend, and even a 'manager' user**.

PR #2602, opened 2017-05-24
Add proxy enhancements 

### LINKED ISSUE #2553: UI Enhancement for Add proxy view
image

1. Add examples in texts (italic and appearing disable) to fields: (can be the same style as that of HTTP API URL
    - a. Host
    - b. Port
2. Rename HTTP API URL as **Configuration API Endpoint**
3. Give Infor Tip explaining what is TLS. The message should contain the full meaning of TLS.

55:
Closes #2553 

- [x] Add examples in texts (italic and appearing disable) to fields: (can be the same style as that of HTTP API URL
  - [x] a. Host
  - [x] b. Port
- [x] Rename HTTP API URL as Configuration API Endpoint
- [x] Give Info Tip explaining what is TLS. The message should contain the full meaning of TLS.

### DIFF
--- a/core/lib/i18n/en.i18n.json
+++ b/core/lib/i18n/en.i18n.json
@@ -820,10 +820,12 @@
                       "label": "Protocol"
                     },
                     "host": {
-                      "label": "Host"
+                      "label": "Host",
+                      "placeholder": "https:\/\/example.com\/"
                     },
                     "port": {
-                      "label": "Port"
+                      "label": "Port",
+                      "placeholder": "i.e. 224"
                     },
                     "tls": {
                       "label": "TLS"
@@ -831,10 +833,11 @@
                   }
               },
               "httpApi": {
-                  "label": "HTTP API URL"
+                  "label": "Configuration API Endpoint"
               },
               "elasticsearch": {
-                  "label": "ElasticSearch URL"
+                  "label": "ElasticSearch URL",
+                  "placeholder": "https:\/\/example.com\/"
               }
           }
       },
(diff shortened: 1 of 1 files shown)


core/lib/i18n/en.i18n.json
```diff
@@ -820,21 +820,24 @@
...
                     },
                     "host": {
-                      "label": "Host"
+                      "label": "Host",
+                      "placeholder": "https:\/\/example.com\/"
                     },
                     "port": {
-                      "label": "Port"
+                      "label": "Port",
+                      "placeholder": "i.e. 224"
```
matleppa: E.g.  instead of i.e.

core/lib/i18n/en.i18n.json
```diff
@@ -820,21 +820,24 @@
...
                     },
                     "tls": {
                       "label": "TLS"
```
matleppa: Should there be a tooltip explaining the meaning og TLS?

55: Yep, in a process of figuring out the best way.

55:
@Nazarah are there any other ways to explain the purpose of TLS but tooltip helper?
We have limitations here, since we are generating these fields all at once and not separately one by one.
Perhaps, transcript of it in brackets will work?

@brylie @marla-singer any thoughts on this?

marla-singer:
@55 I suggest using "title" attribute for that. What do you think?

55:
@marla-singer yes, I also think that this might be an overdo with help here.

Nazarah:
Ok with the changes. Good work Ilya!

PR #2651, opened 2017-06-12
REST API additional work

### LINKED ISSUE #2645: User API, a user is removed, but id remains in Organization manager list
Problem:
When a user, who is a manager in one or more Organizations, is removed, the user's id remains in Organization(s) manager list(s).

When Organizations manager list is displayed, it looks like the user id is removed, because the deleted users email address is not visible. However, the user id still remains in list.

What to do?
- when user is deleted, a new check is needed whether the user is a manager in any Organization. In case yes, the user's id must be removed from manager list(s).
- also it needs to be checked, if there are other similar lists, where deleted user's id needs also to be removed.
- it needs to be decided, what to do in case the deleted user is the last manager in Organization (or in another similar list): 
a) add extra confirmation ("Are you really removing the last manager?"), if yes, then remove. 
b) prohibit removal of user and inform ("User cannot be removed, because of being only manager in XXX Organization. Nominate another manager or remove Organization and then try again")
c) just remove id from list, which most likely causes problems afterwards.

Same checking functionality needs to be implemented both in UI operation and in REST API operation.

### LINKED ISSUE #2648: User API, request one user data with id
Although not described in User API document, Restivus offers possibility to request user data with id. However, because of structure of Users collection (there is declared a subschema for user profile before actual schema for user data), only user name and Company name are presented with that Restivus implicite functionality.

The functionality for requiring User data by id is needed to be implemented explicitly in order to get all necessary User data.

### LINKED ISSUE #2678: REST API: Users swagger finalization
### Background
In order to get REST API functionality included in delivery, some swagger file generation content was left out.

### What is missing
Swagger generation about Users route: 
- description of successful response message content in methods GET, GET{id}, GET updates and POST.
- some parameter descriptions needs to be adjusted

### LINKED ISSUE #2693: apinf API page "rest/v1/swagger.json" has no login info
when you call this API with postman first page you end up is https://nightly.apinf.io/rest/v1/swagger.json.
This page has no login info (eg how to login, what address to use...)
apinf api

You need login info to modify and create new APIs to the site.

Correct address is: https://nightly.apinf.io/rest/v1/login
POST
{
"username": "account",
"password": "example"
}

matleppa:
Closes #2648 
Closes #2645 
Closes #2678 
Closes #2693

### DIFF
--- a/core/server/api.js
+++ b/core/server/api.js
@@ -246,30 +246,7 @@ ApiV1.swagger = {
       },
 
       '/users/{id}': {
-        get: {
-          tags: [
-            UserFields.struct.tags.users,
-          ],
-          description: 'Returns user data with given ID.',
-          parameters: [
-            UserFields.struct.params.userId,
-          ],
-
-          responses: {
-            200: {
-              description: 'Data of identified user.',
-            },
-            401: {
-              description: 'Authentication is required',
-            },
-            403: {
-              description: 'User does not have permission.',
-            },
-            404: {
-              description: 'No user found with given UserID.',
-            },
-          },
-        },
+        get: {},
         delete: {
           tags: [
             UserFields.struct.tags.users,
--- a/core/server/api.js
+++ b/core/server/api.js
@@ -152,7 +152,7 @@ ApiV1.swagger = {
     },
     organizationId: {
       name: 'id',
-      in: 'query',
+      in: 'path',
       description: 'ID of Organization',
       required: true,
       type: 'string',
--- a/core/server/api.js
+++ b/core/server/api.js
@@ -65,15 +65,6 @@ UserFields.params = {
     required: true,
     type: 'string',
   },
-  skip: {
-    name: 'skip',
-    in: 'query',
-    description: 'Number of records to skip for pagination.',
-    required: false,
-    type: 'integer',
-    format: 'int32',
-    minimum: 0,
-  },
   since: {
     name: 'since',
     in: 'path',
@@ -82,6 +73,15 @@ UserFields.params = {
     type: 'integer',
     format: 'int32',
     minimum: 1,
+  },
+  skip: {
+    name: 'skip',
+    in: 'query',
+    description: 'Number of records to skip for pagination.',
+    required: false,
+    type: 'integer',
+    format: 'int32',
+    minimum: 0,
   },
   sortBy: {
     name: 'sort_by',
--- a/core/server/api.js
+++ b/core/server/api.js
@@ -26,25 +26,25 @@ UserFields.struct = {
 
   params: {
     addUser: {
-      name: 'Update user',
+      name: 'addUserData',
       in: 'body',
-      description: 'Data for adding User',
+      description: 'Data for adding a user account',
       schema: {
-        required: ['userName', 'password', 'email'],
+        required: ['username', 'password', 'email'],
         properties: {
-          userName: {
+          username: {
             type: 'string',
-            example: 'John Doe',
+            example: 'johndoe',
           },
           password: {
             type: 'string',
-            example: 'secrecy_is_everything#¤%',
+            example: 'password',
           },
           email: {
             type: 'string',
             format: 'email',
             description: 'E-mail address of user',
-            example: 'company-mail@gmail.com',
+            example: 'firstname.lastname@vendor.com',
           },
         },
       },
@@ -136,15 +136,15 @@ UserFields.struct = {
       type: 'string',
     },
     updateUser: {
-      name: 'Update user',
+      name: 'updateUserData',
       in: 'body',
-      description: 'Data for editing User',
+      description: 'Data for updating a user account',
       schema: {
         required: [],
         properties: {
-          userName: {
+          username: {
             type: 'string',
-            example: 'John Doe',
+            example: 'johndoe',
           },
           company: {
             type: 'string',
@@ -152,7 +152,7 @@ UserFields.struct = {
           },
           password: {
             type: 'string',
-            example: 'secrecy_is_everything#¤%',
+            example: 'updatedPassword',
           },
         },
       },
@@ -164,7 +164,7 @@ UserFields.struct = {
       required: true,
       type: 'string',
     },
-    userName: {
+    username: {
       name: 'username',
       in: 'body',
       description: 'Name of user',
@@ -460,7 +460,7 @@ ApiV1.swagger = {
       required: true,
       type: 'string',
     },
-    userName: {
+    username: {
       name: 'username',
       in: 'body',
       description: 'Username',
@@ -564,7 +564,7 @@ ApiV1.swagger = {
       required: ['name', 'email', 'password'],
       properties: {
         username:
-          UserFields.struct.params.userName,
+          UserFields.struct.params.username,
         email:
           UserFields.struct.params.email,
         password:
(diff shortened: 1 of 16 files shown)


bajiat:
@marla-singer Could you please review this PR?

organizations/server/methods.js
```diff
@@ -100,6 +100,23 @@ Meteor.methods({
       }
      );
   },
+  removeUserFromAllOrganizations (userId) {
+    // Make sure userId is an String
+    check(userId, String);
+    // Check if user to be deleted is a manager in any Organizations
+    const organizations = Organizations.find({
```
marla-singer: @matleppa First, field `managerIds` has type Array and your query isn't correct. You must use `$in` operator. [Link to doc](https://docs.mongodb.com/manual/reference/operator/query/in/#op._S_in)
```js
const organizations = Organizations.find({ managerIds: { $in: userId } }).fetch();
```
Fix it.

The second one, Comment above has a wrong semantic because *you don't check any data, you just get data*.
The comment should be "Get all organization where a user is a manager" or something like that.
Fix it

core/server/api.js
```diff
@@ -17,6 +17,95 @@ const ApiV1 = new Restivus({
   enableCors: true,
 });
 
+const UserFields = {};
+
+UserFields.params = {
```
marla-singer: @matleppa Can you please explain why did you create this list of parameters if these parameters already exist?
https://github.com/apinf/platform/blob/develop/core/server/api.js#L63-L162

core/server/api.js
```diff
@@ -26,6 +115,161 @@ ApiV1.swagger = {
       version: '1.0.0',
       title: 'Admin API',
     },
+    paths: {
+      '/users': {
```
marla-singer: Can you please explain why did you create these paths if these paths already exist?
https://github.com/apinf/platform/blob/develop/users/collection/server/api.js#L23-L29

marla-singer:
@matleppa Also don't forget to set the number of Sprint

core/server/api.js
matleppa: In restivus-swagger.js there are explicitly prohibited generation of swagger document about paths that contain string "users". 
However, there is a work-around: when paths are explicitly declared in Api1V1.swagger object, they are included in generated swagger document.
https://github.com/apinf/restivus-swagger#define-paths-outside-of-routes
Discussion about the matter: https://github.com/apinf/restivus-swagger/issues/22

core/server/api.js
matleppa: Because I had to add outside of route definition for paths containing string "users" (as explained in next comment), this was only way I could find in order to be able to reference to parameters from inside users path definition here.
Perhaps there is a way to reference , but I could not find it at the time. 
I'll look again.

core/server/api.js
marla-singer: Sorry, I can't see the whole picture. Can you write an example or steps for reproduce when you really need to call request to endpoints `'/users'` and `/users/update` and  the first solution doesn't work?
You already have implementation for path `/users/update` -> https://github.com/apinf/platform/blob/develop/users/collection/server/api.js#L550

What kind of data do you need from Users collection that doesn't available from standart solution? -> https://github.com/apinf/platform/blob/develop/users/collection/server/api.js#L152-L153

@brylie Or maybe you can explain me? I just see the duplication of code and don't see any reason why it is so.

organizations/server/methods.js
matleppa: @marla-singer I disagree using $in here. 
Isn't it so, that the $in operator is used in case there is an array of values as serach parameter, and the value in collection has to match with one or more of them. 
Here there is always only one value (userId) as search parameter and according to documentation of find, it finds match also in case the field in collection is an array. 
(Also the version with $in didn't work).

core/server/api.js
matleppa: There are two separate issues here.

The functionality of requests for ```/users``` path is implemented in https://github.com/apinf/platform/blob/develop/users/collection/server/api.js
It is working fine.

However, when the swagger document is generated from the program code, it is done with restivus-swagger.js.
Now, in restivus-swagger.js there is some restirictions concerning the generation of swagger document. One of them is to exclude all paths, which include string ```"users"```.
https://github.com/apinf/restivus-swagger/blob/master/server/restivus-swagger.js#L46-L51
So the ```users``` related routes are not included in generated swagger document.

However, there is a possibility to add path into generated swagger document.
https://github.com/apinf/restivus-swagger/blob/master/server/restivus-swagger.js#L80-

So being able to generate complete swagger document is the reason, why ```/users``` related paths are added into Api1V1.swagger object.

organizations/server/methods.js
matleppa: We also checked this matter with Brylie.
Perhaps it was $elemmatch operator you were thinking.
However, it is said in documentation, that in case $elemmatch specifies only one condition, the $elemmatch keyword need not to be used.
https://docs.mongodb.com/manual/reference/operator/query/elemMatch/

marla-singer:
@matleppa Okay, maybe you're right and restivus-swagger.js doesn't work correct with`/user`.
But anyway, PR is not ready for review. Swagger UI for Users part doesn't look as expected
joxi_screenshot_1497861187070

marla-singer:
@matleppa I tested the functionality and found some no good things.
1. Let's talk about **For which methods for Users collection authorization is required?**
Now I see **all methods** are required. As for me, it isn't correct for POST method.
As an unregistered user, I want to create an account using POST but I can't because it needs authorization. Looks like a vicious circle.

**My suggest is to set no required authorization for POST method**

2. Incorrect User model for POST and PUT methods
    - Rename name of model "Update user" to more general. For example, "Data", "User data" and so  on
    - Rename field "userName" to "username" because field names must be the same as in collection schema
    - Change the example of username because "John Doe" is a wrong username
    - Remove the strange symbols from an example of a password. Looks very strange

joxi_screenshot_1498053314171

**Fix all four points**

3. I found the method not duplicate the code. First, don't duplicate `params` and `tags` for Swagger. I will prepare a branch where I will show you. Please repeat it (or merge into yours) and clean up code

marla-singer:
Removing the user from Organization manager list works as expected :+1:

bajiat:
@mauriciovieira Could you help with the Travis errors related to Chimp.js?

@matleppa Is there anything else missing from this PR than the Travis errors?

matleppa:
@bajiat  While I could not get further because of Travis problem, I added also login functionality and use of login credential in other operations into generated swagger document. (Closing 2693).
I think that now this PR has to be done, because the line needs to be drawn somewhere.
There are other issues pending related to REST API, but it is more clear to continue with then by own branches.

marla-singer:
@bajiat @matleppa It should be resolved by another PR

marla-singer:
@matleppa When this PR will be ready to review, please set the label and ping me

marla-singer:
@matleppa I've started to code review but I thought this PR is too big and I just tested the features.

1. Removing user ID from Organization  works :+1: 
2. The swagger file contains information about Login path :+1: 
3. Login request works es expected :+1: 
4. The swagger file contains summary information for each endpoint :+1: 
5. The swagger file contains correct UI about Users :+1: 

I added a little fix for parameter `organizationId`

Also, I found redundant using a parameter with type "in: body". But I will create separated issue for that
And another "bugs" I will describe in another issue

PR #2655, opened 2017-06-13
bugfix missing ID's

### LINKED ISSUE #2623: Buttons in the UI must have an ID associated with them
Almost none of the buttons at apinf.io have valid way to locate them in automated testing. Common way to locate them is to add unique id to the buttons.

wuder4:
Closes #2623

Adding ID's to most of the buttons in the site. Rest of the ID's are coming soon.


marla-singer:
@wuder4 Why do we need this set of IDs? You don't reference to any issue.

Also if your PR is ready for review then please set the label "Ready for review"

wuder4:
This pull is still in progress but it will fix #2623

55:
@wuder4 also we are using "Closes #issue-number" to automatically close issue after PR got merged.

core/client/footer/footer.html
```diff
@@ -38,7 +38,7 @@
           <div class="row">
             {{# if currentUser }}
               <div class="col-xs-6 col-md-12">
-                <a href="{{ pathFor 'signOut' }}" id="bottom-signout">
+                <a class="sign-out" href="{{ pathFor 'signOut' }}" id="bottom-signout">
```
brylie: We might not need both a `class` and `id` here. Why do we add the `class="sign-out"`?

ashakunt: IMHO, you can still keep the class because class will classify what is the purpose (here, signout) and the ID will pin-point the place (here: bottom-sign-out).

ashakunt: Also, having the `class=signout` will help in handling all `signout` scenarios in similar way, e.g. 

```
$(.signout).click(function (e) {doSignout()})
```
and hence, will reduce redundancy. The purpose of having an ID is to support testing.

brylie: Are we using the `.sign-out` in any event handlers? If not, I don't think it is needed, as it just adds unnecessary clutter.

> Also, having the class=signout will help in handling all signout scenarios in similar way, e.g.
>
> $(.signout).click(function (e) {doSignout()})

For what it's worth, we don't have many/any, global event handlers that I am aware of.

brylie: In this specific case, the 'sign-out' event handler is part of a third party module.

The code in question is just an [hyperlink to the sign-out route](https://github.com/apinf/platform/blob/develop/core/client/footer/footer.html#L41-L43), so it is unlikely that we would use the class in a JavaScript event handler.

In this case, I respectfully maintain that having both the class and ID is redundant, and recommend we only use an ID to accomplish the related goal of improving testability of our code.

ashakunt: if we dont have, then its ok to remove class.. :)

brylie: Cool beans. It is a small nit, so not really worth too much disagreement in any case :-)

One thing to note, though, is we are trying to keep this PR relatively small for now, and continue the work of adding IDs in one or more follow-up PRs.

ashakunt: Small drops fill the ocean!

apis/client/add/add.html
```diff
@@ -13,12 +13,12 @@ <h1 class="page-header">
         <!--invisible auto-value field-->
         {{> afQuickField name='latestMonitoringStatusCode' value="-1" type="hidden"}}
         <fieldset>
-          {{> afQuickField name='name' }}
-          {{> afQuickField name='description' rows=2 }}
-          {{> afQuickField name='url' }}
-          {{> afQuickField name='lifecycleStatus' }}
+          {{> afQuickField name='name' id="api-name"  }}
```
55: Extra space here

core/client/footer/footer.html
```diff
@@ -38,18 +38,18 @@
           <div class="row">
             {{# if currentUser }}
               <div class="col-xs-6 col-md-12">
-                <a class="sign-out" href="{{ pathFor 'signOut' }}">
+                <a class="sign-out" href="{{ pathFor 'signOut' }}" id="bottom-signout">
```
55: I think it's better to use` id="footer-..."`, since footer is common term.

core/client/navbar/navbar.html
```diff
@@ -15,7 +15,7 @@
         </a>
         {{ else }}
         <div class="project-name">
-          <a href="{{ pathFor 'home' }}" class="project-name-link">
+          <a href="{{ pathFor 'home' }}" class="project-name-link" id="frontpage-button">
```
55: Navbar is on every page, so perhaps` id="navbar-projecttitle-button"` or ` id="projecttitle-button"`?

brylie:
@55 are these changes alright now?

55:
@wuder4 is this PR ready to be merged or you are still have ids to add?

wuder4:
> @wuder4 is this PR ready to be merged or you are still have ids to add?

I still have few pages left.

55:
@wuder4 just put `Ready to review` label if it so.

PR #2702, opened 2017-06-29
Changing site title and slogan id's

### LINKED ISSUE #2700: UI, Branding: wrong text color in input fields
### Wrong input text color
When values in Branding page for following input fields
- Site title 
- Site slogan

 are inputted, the color of *input text* is the same as color defined as Primary text.

E.g. In case the Primary text color is set to white, the inputted characters can not be seen.
#### What is expected 
Color must be the default color for input fields.

#### Probable cause
Same html element id is used for input fields and for actual Site title and Site slogan, thus color set by css is set also for input fields.

#### Correction
Change html element ids for input fields to different from actual Site title and Site slogan.

wuder4:
Closes #2700 

This pr fixes color bug in branding settings

### DIFF
--- a/branding/client/branding.html
+++ b/branding/client/branding.html
@@ -26,8 +26,8 @@ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence
                 type="update"
                 collection=brandingCollection
                 doc=branding }}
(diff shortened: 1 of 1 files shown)


marla-singer:
@wuder4 @matleppa Please set also the Milestone in PR and the related issue

PR #2705, opened 2017-06-30
Remove deleted Users userid from from Organization manager list

### LINKED ISSUE #2664: Using UI, removed user's id remains in Organization manager list
### Problem:
When a user, who is a manager in one or more Organizations, is removed, the user's id remains in Organization(s) manager list(s).
When Organizations manager list is displayed, it looks like the user id is removed, because the deleted users email address is not visible. However, the user id still remains in list.

### What to do?

When user is deleted, a new check is needed whether the user is a manager in any Organization. In case yes, the user's id must be removed from manager list(s).

It needs to be decided, what to do in case the deleted user is the last manager in Organization (or in another similar list):
a) add extra confirmation ("Are you really removing the last manager?"), if yes, then remove.
b) prohibit removal of user and inform ("User cannot be removed, because of being only manager in XXX Organization. Nominate another manager or remove Organization and then try again")
c) just remove id from list, which most likely causes problems afterwards.

### Other lists?
Also it needs to be checked, if there are other similar lists, where deleted user's id needs also to be removed.

matleppa:
Closes  #2664

### DIFF
--- a/users/server/methods.js
+++ b/users/server/methods.js
@@ -21,6 +21,9 @@ Meteor.methods({
 
     let user;
     if (this.userId === userId) {
+      // Remove user from all Organizations
+      Meteor.call('removeUserFromAllOrganizations', userId);
+
       user = Meteor.users.remove({
         _id: this.userId,
       });


marla-singer:
@matleppa Is PR ready for review? If it is, please set the label and Milestone for PR and issue

matleppa:
Ok, let's leave the functionality when Admin is deleting a user outside of scope of this PR.
Perhaps it needs to be studied later (via another issue).

PR #2717, opened 2017-07-05
adding ids to buttons that I missed last PR

wuder4:
adding ids to viewmode buttons

### DIFF
--- a/organizations/client/catalog/toolbar/toolbar.html
+++ b/organizations/client/catalog/toolbar/toolbar.html
@@ -77,14 +77,14 @@ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence
                 class="btn btn-default toolbar-tooltip active"
                 data-original-title="{{_ 'organizationCatalog_toolbar_viewMode_grid' }}">
               <input type="radio" class= "filter-sort-control" name="viewMode" value="grid">
-              <i class="fa fa-th-large" aria-hidden="true"></i>
+              <i class="fa fa-th-large" id="viewmode-grid" aria-hidden="true"></i>
             </label>
             <label
                 for="table-view"
                 class="btn btn-default toolbar-tooltip"
                 data-original-title="{{_ 'organizationCatalog_toolbar_viewMode_table' }}">
               <input type="radio" class= "filter-sort-control" name="viewMode" value="table">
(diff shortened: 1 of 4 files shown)


organizations/client/catalog/toolbar/toolbar.html
```diff
@@ -77,14 +77,14 @@
                 class="btn btn-default toolbar-tooltip active"
                 title="{{_ 'organizationCatalog_toolbar_viewMode_grid' }}">
               <input type="radio" class= "filter-sort-control" name="viewMode" value="grid">
-              <i class="fa fa-th-large" aria-hidden="true"></i>
+              <i class="fa fa-th-large" id="viewmode-grid" aria-hidden="true"></i>
```
marla-singer: I'm not sure that you should add ID attribute to `<i>` tag because it is **just an icon**.
Try to add ID attribute to `<input>` element

organizations/client/catalog/toolbar/toolbar.html
```diff
@@ -77,14 +77,14 @@
...
             </label>
             <label
                 for="table-view"
                 class="btn btn-default toolbar-tooltip"
                 title="{{_ 'organizationCatalog_toolbar_viewMode_table' }}">
               <input type="radio" class= "filter-sort-control" name="viewMode" value="table">
-              <i class="fa fa-table" aria-hidden="true"></i>
+              <i class="fa fa-table" id="viewmode-table" aria-hidden="true"></i>
```
marla-singer: The same as above. Add ID attribute to `<input>` element

wuder4:
how bad would it be is we kept the id on the icon? test case kind of needs it to be there 

testcase

organizations/client/catalog/toolbar/toolbar.html
wuder4: fixed this problem by changing to label

organizations/client/catalog/toolbar/toolbar.html
wuder4: fixed this problem by changing to label

marla-singer:
@wuder4 Please make the rebase and squash all commit into one with the name "Add missing IDs for testing"
Don't make multiple copies of commits for small changes

PR #2754, opened 2017-07-17
Feature/separate swagger files

### LINKED ISSUE #2737: REST API: Generate own swagger files of Apis API / Management API (organizations + Users)
Instead of generating one swagger file containing all REST API endpoints, several swagger files needs to be generated.
- one about Apis handling API 
- one about Organizations and Users handling API

matleppa:
Closes #2737

### DIFF
--- a/core/server/api.js
+++ b/core/server/api.js
@@ -22,9 +22,9 @@ ApiV1.swagger = {
   meta: {
     swagger: '2.0',
     info: {
-      description: 'APinf is open source API management and catalog. ',
+      description: 'APinf is open source API management and catalog.. ',
       version: '1.0.0',
-      title: 'Admin API',
+      title: 'Admin API for API handling',
     },
     // Create  placeholder for storage paths for Users collection
     paths: {},
@@ -43,9 +43,6 @@ ApiV1.swagger = {
   },
   tags: {
     api: 'APIs',
-    organization: 'Organizations',
-    users: 'Users',
-    login: 'Login',
   },
   params: {
     api: {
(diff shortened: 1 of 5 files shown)


core/server/api.js
```diff
@@ -22,9 +22,9 @@ ApiV1.swagger = {
   meta: {
     swagger: '2.0',
     info: {
-      description: 'APinf is open source API management and catalog. ',
+      description: 'APinf is open source API management and catalog.. ',
```
marla-singer: Delete the last point on the sentence

core/server/api.js
```diff
@@ -43,9 +43,6 @@ ApiV1.swagger = {
   },
   tags: {
     api: 'APIs',
-    organization: 'Organizations',
-    users: 'Users',
-    login: 'Login',
```
marla-singer: i think the endpoint §login must be declared for each swagger specification because the endpoints `/apis` and `/apis{id}` also supposes the authorization.

**Recover login tags and endpoint /login for swagger-api.json file**

core/server/maintenance.js
```diff
@@ -0,0 +1,501 @@
...
+            },
+          },
+        },
+      },
+    },
+
+  },
+};
+
+// Generate Swagger to route /rest/v1/swagger.json
```
marla-singer: Change the comment, should be:
`// Generate Swagger to route /rest/v1/swagger_maintenance.json`

core/server/api.js
```diff
@@ -438,196 +203,10 @@ ApiV1.swagger = {
...
-              example: 'My Company Ltd',
-            },
-          },
-        },
-      },
-    },
   },
 };
 
 // Generate Swagger to route /rest/v1/swagger.json
```
marla-singer: Change the comment, should be 
`// Generate Swagger to route /rest/v1/swagger_api.json`

PR #2781, opened 2017-07-21
REST API: handle APIs, admin to perform PUT and DELETE, duplicate error response in POST

### LINKED ISSUE #2721: REST API: duplicate error,  POST method API endpoint
When API is created with REST API with POST method, it needs to  check, whether an API with same name already exists and return appropriate controlled response.

### LINKED ISSUE #2723: REST API: API endpoint, PUT and DELETE, allow methods for admin
When API endpoint methods PUT and DELETE are used, they must be allowed for admin user also.
It is not enough to allow them only for users included in API manager list.
*Additional information to correct original poor choice of words*:
Here allowed means, that operations has to be performed also in case admin sends a request via REST API.

matleppa:
Closes #2723 
Closes #2721

### DIFF
--- a/apis/collection/helpers.js
+++ b/apis/collection/helpers.js
@@ -28,14 +28,14 @@ import ApiMetadata from '/metadata/collection';
 import Apis from './';
 
 Apis.helpers({
-  currentUserCanManage () {
+  currentUserCanManage (apiUserId) {
     // Get current userId
-    const userId = Meteor.userId();
-
+    // parameter apiUserId is used only in case of REST API
+    const userId = apiUserId || Meteor.userId();
     // Check that user is logged in
     if (userId) {
       // Check if user is manager of this API
-      const userIsManager = this.currentUserIsManager();
+      const userIsManager = this.currentUserIsManager(apiUserId);
 
       // Check if user is administrator
       const userIsAdmin = Roles.userIsInRole(userId, ['admin']);
@@ -47,7 +47,7 @@ Apis.helpers({
       let userIsOrganizationManager;
 
       if (parentOrganization) {
-        userIsOrganizationManager = parentOrganization.currentUserCanManage();
+        userIsOrganizationManager = parentOrganization.currentUserCanManage(userId);
       }
 
       // if user is manager or administrator, they can edit
@@ -69,9 +69,10 @@ Apis.helpers({
     // Only user who can edit, can view private APIs
     return (this.isPublic || userIsAuthorized || this.currentUserCanManage());
   },
-  currentUserIsManager () {
-    // Get current User ID
-    const userId = Meteor.userId();
+  currentUserIsManager (apiUserId) {
+    // Get current userId
+    // parameter apiUserId is used only in case of REST API
+    const userId = apiUserId || Meteor.userId();
 
     // Check if user is manager of this API
     const isManager = _.includes(this.managerIds, userId);
--- a/apis/server/api.js
+++ b/apis/server/api.js
@@ -311,24 +311,26 @@ ApiV1.addCollection(Apis, {
         // Get ID of API
         const apiId = this.urlParams.id;
         const userId = this.userId;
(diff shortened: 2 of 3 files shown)


apis/server/api.js
```diff
@@ -311,24 +322,26 @@ ApiV1.addCollection(Apis, {
...
+        const IsAdmin = Roles.userIsInRole(userId, ['admin']);
+
         const api = Apis.findOne(apiId);
 
         // Make sure API exists & user can manage
-        if (api && api.managerIds.includes(userId)) {
-          // Update API document
-          Apis.update(apiId, { $set: bodyParams });
+        if (api) {
+          if (api.managerIds.includes(userId) || IsAdmin) {
```
marla-singer: @matleppa No, it is not a good idea because it is not a simple way to check how can really managed API. Please look at this collection helper https://github.com/apinf/platform/blob/develop/apis/collection/helpers.js#L31
As I remember this collection helper doesn't work when is run in Swagger part. I recall @brylie suggested creating server method for that and re-use it.

Please create a server method for checking permission. Logic copy from the related collection helper

apis/server/api.js
```diff
@@ -387,25 +400,26 @@ ApiV1.addCollection(Apis, {
...
+        const IsAdmin = Roles.userIsInRole(userId, ['admin']);
         // Get API document
         const api = Apis.findOne(apiId);
 
         // Make sure API exists & user can manage
-        if (api && api.managerIds.includes(userId)) {
-          // Remove API document
-          Meteor.call('removeApi', api._id);
+        if (api) {
+          if (api.managerIds.includes(userId) || IsAdmin) {
```
marla-singer: The same way is here. Use the server method for checking

apis/server/api.js
matleppa: Ok, I see your point.

While checking the helper currentUserCanManage, I was thinking, that instead of writing two almost identical helpers, could we reuse the existing one?

Current implementation: 
```
  currentUserCanManage () {
    // Get current userId
    const userId = Meteor.userId();

    // Check that user is logged in
    if (userId) {
```
With additional parameter, which is filled only in case helper is called from REST API.
```
  currentUserCanManage (apiUserId) {
    // Get current userId
    let userId = Meteor.userId();

    if (apiUserId) {
      userId = apiUserId;
    }
    // Check that user is logged in
    if (userId) {
```
Only difference is the way the user ID is gotten, the rest of functionality remains the same.
@marla-singer  What is your opinion about this?

marla-singer: @matleppa Let's try this way
```js
currentUserCanManage(apiUserId) {
  let userId = apiUserId || Meteor.userId();
...
}
```

matleppa: Thanks, it is much more elegant than my original proposal.

matleppa: It seems, that same kind of handling is needed to be added also in helpers, which are called from the first one.

marla-singer:
@matleppa Works fine. There leaves a little changes. Also I created a new Sprint, please, set the sprint both PR and issues

apis/collection/helpers.js
```diff
@@ -28,14 +28,14 @@ import ApiMetadata from '/metadata/collection';
...
     // Get current userId
-    const userId = Meteor.userId();
-
+    // parameter apiUserId is used only in case of REST API
+    const userId = apiUserId || Meteor.userId();
     // Check that user is logged in
     if (userId) {
       // Check if user is manager of this API
-      const userIsManager = this.currentUserIsManager();
+      const userIsManager = this.currentUserIsManager(apiUserId);
```
matleppa: I disagree.
The point is, that ONLY in case the request is gotten from REST API, the parameter is given in helper call. So here in this helper, it the REST API parameter is got, it is named apiUserId. In case it is not got, it is empty also in calling next helper.

marla-singer: Okay, I got you. But my inner esthetic tells me "It will be better with userId params!" :D

PR #2787, opened 2017-07-25
Integrate OpenAPI Designer into the platform

### LINKED ISSUE #2780: Add OpenAPI designer iframe integration
Wireframes available [here](https://github.com/apinf/platform/issues/1998).

Example OpenAPI designer in iframe with authtoken sending support available [here](https://dev.openapi.design/iframe.html).

marla-singer:
Closes #2780

### Know issue
if API has a uploaded swagger documentation then Editor doesn't "catch" this file.


marla-singer:
@matleppa Will you have time for review that?

matleppa:
Sorry, I noticed this just now. I'll look at it tomorrow.

matleppa:
I wonder whether the functionality is correct:

1. open API A
2. open OpenAPIDesigner editor
3. create document "doc_API_A"
4. open API B
5. open editor

The on step 5 the editor has same document ("doc_API_A") open, which was created on step 3.
I would expect, that the document should somehow be attached to APIs. 
I.e.
On step 4 I'll expect an empty document.
In case re-opening API A, I expect to see "doc_API_A" again.

What do you think?

matleppa:
I saw some discussion about whether to use iframe or not. 
Is the final decision that iframe is used?

marla-singer:
> On step 4 I'll expect an empty document.
In case re-opening API A, I expect to see "doc_API_A" again.
What do you think?

I think this issue is related to OpenAPI Designer developers (@tulir, @xylix ). I had a quick view on releasing and probably it happens because data is stored in localStorage. If you delete it and go to API B then iframe will not contain any data but API A will not as well

> I saw some discussion about whether to use iframe or not.
Is the final decision that iframe is used?

I suppose it is not the final decision. But like a beta release - why not?

Nazarah:
@marla-singer @matleppa can you please ask @kyyberi about integrating OpenAPI Designer with iframe? as far as I know, he would be making the decision about it.

api_docs/client/view/documentation.html
```diff
@@ -5,33 +5,47 @@
...
-          {{/ if }} <!-- /if documentation exists -->
-        {{/ if }}<!-- /if codegen server exists -->
-      {{/ if }}<!-- /if template subscriptions ready -->
-      <h2 id="documentation-title" class="panel-title clearfix">
-        {{_ "apiDocumentation_Title" }}
-      </h2>
-    </div>
+          {{/ if }} <!-- /if currentUserCanManage -->
+        </div>
+        <h2 id="documentation-title" class="panel-title clearfix">
```
brylie: Why is this heading at the bottom of it's containing element?

brylie:
It would be nice if the changes related to internationalization were in their own pull request, with an accompanying issue. That would make this PR a bit cleaner.

brylie:
Please resolve conflicts.

.editorconfig
```diff
@@ -27,7 +27,7 @@ charset = utf-8
 indent_style = space
 indent_size = 2
 
-[core/**/*.i18n.json]
+[core/**/fi.i18n.json]
```
brylie: Can we put i18n related changes in a separate PR?

marla-singer: Ok

brylie: Cool, thanks. Also, open an issue, since it is work you are doing. Anything worth doing is worth an issue (within reason) :-)

api_docs/client/view/documentation.html
marla-singer: The buttons have style `float:right` and markup works properly if buttons are at the top.
I changed the order and markup works bad
joxi_screenshot_1501574255402

.editorconfig
marla-singer: It is not a critical change just "nice to have". It is related to spacebars in the file

api_docs/client/view/documentation.html
brylie: Alright. It is just a bit unconventional for a header to be at the bottom of a block element.

.editorconfig
brylie: Right, I like it.

api_docs/client/view/documentation.html
marla-singer: Agree.

marla-singer:
@brylie 
- Resolved conflicts
- Remove changes related to the file `.editorconfig`

api_docs/client/editor/editor.html
```diff
@@ -0,0 +1,9 @@
+<!-- Copyright 2017 Apinf Oy
+This file is covered by the EUPL license.
+You may obtain a copy of the licence at
+https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence-eupl-v11 -->
+
+<template name="openApiEditor">
+  <iframe id="designer-iframe" src="https://openapi.design" frameborder="0"></iframe>
```
brylie: We should probably consider exposing this in the platform settings page, like we do for the codegen.

cc: @bajiat thoughts?

marla-singer: @brylie I thought about it as well.

brylie: It would be consistent, and we generally try to avoid hard-coding these kind of things.

bajiat: I agree, this is what we did with Swagger Editor and this actually could replace Swagger editor setting in platform settings view.

brylie: Should we deprecate Swagger editor in place of OpenAPI Designer?

brylie:
@bajiat @marla-singer will the following be included in the definition of done for this PR?

- add 'OpenAPI Designer URL' to configuration settings page
- deprecate Swagger Editor setting

I just need to know what to look for before this PR should be merged.

api_docs/client/editor/editor.html
bajiat: I think we should deprecate Swagger editor.

bajiat:
Should we create another follow-up issue for Settings page? If yes, can we merge without changes?

brylie:
Who will create the follow-up issue? I will merge this PR when the issue has been created and linked to this PR by reference.

marla-singer:
@brylie Done

PR #2797, opened 2017-07-26
Feature/rest api manage organization manager

### LINKED ISSUE #2768: [FIWARE] REST-API: Add endpoint for Organizations part to manipulate managers
Create the endpoint into Organizations part which allows to add/delete manager of current organization

matleppa:
Closes #2768

### DIFF
--- a/organizations/server/api.js
+++ b/organizations/server/api.js
@@ -241,16 +241,12 @@ MaintenanceV1.addRoute('organizations/:id', {
       summary: 'Update Organization.',
       description: `
    ### Update an Organization ###
-   
+
    You can update Organization data with parameters listed on example.
 
    Parameters can be given one by one or several ones at a time.
    When several parameters are given, in successful case all of them are
    updated and in unsuccessful case none of them is updated.
-
-   Only an existing user can be added as a new manager for Organization.
-   In case the manager to be added already exists on list, the operation
-   is considered successful.
       `,
       parameters: [
         MaintenanceV1.swagger.params.organizationId,
@@ -258,7 +254,7 @@ MaintenanceV1.addRoute('organizations/:id', {
       ],
       responses: {
         200: {
-          description: 'Organization successfully edited.',
+          description: 'Organization data successfully updated.',
           schema: {
             type: 'object',
             properties: {
@@ -315,28 +311,6 @@ MaintenanceV1.addRoute('organizations/:id', {
           if (bodyParams.description) {
             organizationData.description = bodyParams.description;
           }
-          // We need to get old manager list and push new user IDs into it
-          if (bodyParams.manager) {
-            // Check if user ID for manager exists
-            if (!Meteor.users.findOne(bodyParams.manager)) {
-              return {
-                statusCode: 404,
-                body: {
-                  status: 'Fail',
-                  message: 'User with given manager ID not found.',
-                },
-              };
-            }
-            // Add new manager only if same user ID was not on list before
-            const alreadyManager = organization.managerIds.includes(bodyParams.manager);
-            if (!alreadyManager) {
-              const managerIds = organization.managerIds;
-              // Add user ID to manager IDs list
-              managerIds.push(bodyParams.manager);
-              organizationData.managerIds = managerIds;
-            }
-          }
-
 
           if (bodyParams.contact_name ||
               bodyParams.contact_phone ||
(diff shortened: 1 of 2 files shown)


organizations/server/api.js
```diff
@@ -291,23 +291,47 @@ MaintenanceV1.addRoute('organizations/:id', {
...
-            socialMedia: {
-              facebook: bodyParams.facebook,
-              instagram: bodyParams.instagram,
-              twitter: bodyParams.twitter,
-              linkedIn: bodyParams.linkedin,
-            },
-          };
+          const organizationData = {};
+
+          if (bodyParams.name) {
```
marla-singer: @matleppa These separating doesn't need. Look at the comment above
```js
// If bodyParams doesn't contain any fields
// then organizationData JSON doesn't contain it as well
```
You can test it

marla-singer:
@matleppa I think it is a bad idea to add managers by IDs. I suggest using e-mail and it assumes the available to provide couple e-mails.

organizations/server/api.js
matleppa: But in that case the other fields in ```organizationData``` are empty. And when it is updated to Organization, the other fields in Organization are emptied. So after giving only ```name``` and ```url``` the field description is empty. contact fields are empty and social media fields are empty.

It means, that in case user wants to update only one field, he/she needs to fill all fields in request  with old values in order not to make them empty.

Is that really the wanted functionality?

matleppa:
OK, I see your point on using e-mail addresses. 
Could you please what do you mean with the rest of the sentence. 

"it assumes the available to provide couple e-mails."


"

organizations/server/api.js
marla-singer: Tested it. The problem was because of invalid structure of organizationData fields.

The valid structure is:
```js
const organizationData = {
            name: bodyParams.name,
            url: bodyParams.url,
            description: bodyParams.description,

            'contact.person': bodyParams.contact_name,
            'contact.phone': bodyParams.contact_phone,
            'contact.email': bodyParams.contact_email,
            
            'socialMedia.facebook': bodyParams.facebook,
            'socialMedia.instagram': bodyParams.instagram,
            'socialMedia.twitter': bodyParams.twitter,
            'socialMedia.linkedIn': bodyParams.linkedIn,
          };
```
### Experiment result
1. Create an organization with contact and social media data
joxi_screenshot_1501165201564

2. Send PUT request with one field
joxi_screenshot_1501165201564 1

3. Other fields are the same
joxi_screenshot_1501165201564 2

marla-singer:
> Could you please what do you mean with the rest of the sentence.
"it assumes the available to provide couple e-mails."

I mean to allow a user adds couple e-mails at one time using a comma. (If it will be possible)

organizations/server/api.js
matleppa: After additional testing it seems, that the values in fields name, url and description remain in DB although they are not given in request.
However, the subfields in fields contact and socialMedia are emptied, in case their values are not given in request. 

1) Organization data, contact person included in

image

2) Change socialMedia data only

image

3) SocialMedia included, but contact person missing

image

4) Alter description

image

5) Description altered, but socialMedia data missing

image

So I think, that the way Organization data change requests are updated to DB, needs to be corrected. 
Do you agree?

marla-singer: @matleppa Did you test the old code (on `develop` branch) or after suggested updating variable `organizationData` ?
Please commit the changes where changes don't work, I want to test it

matleppa: With old code.
Because I did not see your correction comment until now. 
I think there is something wrong in my computer or in system, because this is not the first time I do not see updates on time.

But anyway, I make the correction the way you solved the problem.

matleppa:
About adding managers.
So you prefer adding managers functionality to be a new parameter in PUT /organizations/{:id} endpoint rather than a new endpoint PUT /organization/{:id}/managers/<email>?
(Although I'm not sure, if it can be implemented with email address)
What about deleting a certain manager from Organization?

marla-singer:
My feeling is to create a new endpoint `/organizations/{id}/managers`
As for me, It is a big feature for Organization component. We need to create/edit/delete manager by e-mail and return list with managers.

**GET** `/organizations/{id}/managers` -- return list with managers. Provide information as e-mail, username, id of users.
**Query parameters** - username, e-mail. If parameters are filled then return list of users with specified e-mail OR username

**GET** `/organizations/{id}/managers/{id}` -- return a specified manager
**Path parameter**: ID of user to return

**POST** `/organizations/{id}/managers` -- add one or few managers into Organization.
**Body parameters**: e-mails field'

**DELETE** `/organizations/{id}/managers/{id} -- delete a specified manager
**Path parameter**: ID of user to remove

In that case PUT method isn't needed. 

Permissions:
- Administrator or Organization mamanager can manipulate.
- Organization manager or Administrator can't delete himself.

@matleppa Want your review on my suggestion

matleppa:
You've been thinking this Organization manager issue well through. I agree almost completely.
However, there is one thing which maybe could be discussed.

At the beginning you say: 
> We need to create/edit/delete manager by e-mail and return list with managers.
--

In GET and DELETE methods the user ID is used, though.
I wonder, if the user email address is needed to be used here in POST method at all. 
When user operates via UI, I understand the use of email address, because  the user ID is not visible (and it is hard for human to use).
But because APIs are used programmatically, isn't is more useful to use user ID also in POST method instead?

marla-singer:
> We need to create/edit/delete manager by e-mail and return list with managers.
> In GET and DELETE methods the user ID is used, though.
> I wonder, if the user email address is needed to be used here in POST method at all.

Sorry, my mind wasn't completed. For DELETE method and some GET request we will provide ID to remove a specified user or get details information. As an organization manager I have access to managers IDs in my organization and can delete one or get info about this user.

But for POST method we need to use email because if  I am just Organization manager how I can add my college? I don't have access to college user ID but I know his e-mail.

> When user operates via UI, I understand the use of email address, because the user ID is not visible (and it is hard for human to use).

User ID is stillnot  visible for ordinary users . You can know only owned

matleppa:
OK, well explained. Now I agree completely.

PR #2799, opened 2017-07-31
Feature/rest api get owned api list

### LINKED ISSUE #2773: [FIWARE] REST-API: User must get the list of owned APIs
Now a user can get the list of all APIs or with the specific APIs

### Goal
- Create endpoint to return all managed APIs for this user. It should use header "X-User-id"

### LINKED ISSUE #2790: REST API: GET /apis, parameter organization needs fixing
Matter sot be fixed:
### Parameter name
* in swagger document parameter name is ```organization```, however in code there are prepared for parameter name ```organizationApi```.
More descriptive name were ```organizationId```
### Functionality
Without any parameter all APIs are returned in response message.
* in case parameter organizationId is given and Organization is found, only APIs related to that organization are returned. This is correct.
* In case parameter is given, but organization is not found, all APIs are returned, which is erroneous. Expected response is an error message ```Bad query parameters```

matleppa:
Closes #2773 
Closes #2790 

REST API for APIs.
- new parameter for requesting APIs managed by requestor
- fixes in parameter Organization

### DIFF
--- a/apis/server/api.js
+++ b/apis/server/api.js
@@ -70,7 +70,7 @@ ApiV1.addCollection(Apis, {
    Result: returns maximum of 200 APIs which are managed by equesting user.
 
 
-   Note! When using parameter managedAPIS, the requestor user ID is read
+   Note! When using parameter managedAPIs, the requestor user ID is read
    from X-User-Id field in message header.
         `,
 
@@ -114,22 +114,22 @@ ApiV1.addCollection(Apis, {
 
         // Get APIs managed by user requesting operation
         if (queryParams.managedAPIs) {
-          // Reaquestor User ID needed in header
-          const requestorId = this.request.headers['x-user-id'];
+          // Get Manager ID from header
+          const managerId = this.request.headers['x-user-id'];
 
-          // Response with error in case requestorId is missing
-          if (!requestorId) {
+          // Response with error in case managerId is missing
+          if (!managerId) {
             return {
               statusCode: 400,
               body: {
                 status: 'Fail',
-                message: 'Bad query parameters. Managing User ID missing, expected in header.',
+                message: 'Bad query parameters. Manager ID expected in header (X-User-Id).',
               },
             };
           }
 
           // Set condition for a list of managed APIs
-          query.managerIds = requestorId;
+          query.managerIds = managerId;
         }
 
         // Parse query parameters
--- a/apis/server/api.js
+++ b/apis/server/api.js
@@ -58,13 +58,29 @@ ApiV1.addCollection(Apis, {
           ApiV1.swagger.tags.api,
         ],
         summary: 'List and search public API.',
-        description: 'List and search public APIs.',
+        description: `
+   ### List and search public APIs ###
+
+   Parameters are optional and combinations or parameters can be used.
+
+   Example call:
+
+       GET /apis?limit=200&amp;&managedApis=true
+
+   Result: returns maximum of 200 APIs which are managed by equesting user.
+
+
+   Note! When using parameter managedAPIS, the requestor user ID is read
+   from X-User-Id field in message header.
+        `,
+
         parameters: [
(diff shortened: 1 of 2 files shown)


apis/server/api.js
```diff
@@ -58,13 +58,29 @@ ApiV1.addCollection(Apis, {
...
+   Parameters are optional and combinations or parameters can be used.
+
+   Example call:
+
+       GET /apis?limit=200&amp;&managedApis=true
+
+   Result: returns maximum of 200 APIs which are managed by equesting user.
+
+
+   Note! When using parameter managedAPIS, the requestor user ID is read
```
brylie: Should be `managedAPIs`. Case sensitive.

apis/server/api.js
```diff
@@ -92,21 +108,47 @@ ApiV1.addCollection(Apis, {
...
+          // Reaquestor User ID needed in header
+          const requestorId = this.request.headers['x-user-id'];
+
+          // Response with error in case requestorId is missing
+          if (!requestorId) {
+            return {
+              statusCode: 400,
+              body: {
+                status: 'Fail',
+                message: 'Bad query parameters. Managing User ID missing, expected in header.',
```
brylie: Go ahead and document the header parameter (x-user-id) in the error message. Also, determine whether the parameter is case sensitive. If so, we may want to capitalize the first letter of each 'word': X-User-Id or something similar.

apis/server/api.js
```diff
@@ -92,21 +108,47 @@ ApiV1.addCollection(Apis, {
...
         const queryParams = this.queryParams;
-
         // Only Public APIs are available for unregistered user
         const query = { isPublic: true };
         const options = {};
 
+        // Get APIs managed by user requesting operation
+        if (queryParams.managedAPIs) {
+          // Reaquestor User ID needed in header
+          const requestorId = this.request.headers['x-user-id'];
```
brylie: Lets **keep the variable name consistent** with the parameter name. In this code, we first call something by one name ('user id') and then by another name ('requestor id') and later by a third name ('manager id'). 

Inconsistent code is hard to follow.

apis/server/api.js
matleppa: When header parameter is read in API, it can be referred only with lower case letters.
In Postman the parameter name is not case sensitive. However, in swagger login, the parameter name is capitalized.
So, I choose to write it in documentation by capitalizing the first letters of each word.

apis/server/api.js
```diff
@@ -58,13 +58,29 @@ ApiV1.addCollection(Apis, {
...
+   from X-User-Id field in message header.
+        `,
+
         parameters: [
           ApiV1.swagger.params.optionalSearch,
           ApiV1.swagger.params.organizationApi,
           ApiV1.swagger.params.skip,
           ApiV1.swagger.params.limit,
           ApiV1.swagger.params.lifecycle,
+          ApiV1.swagger.params.managedApis,
```
brylie: I still see a bit of inconsistency with this parameter. Choose one of the following and try to use it consistently, unless there is some specific reason not to (in which case add a comment):

- managedApis
- managedAPIs

apis/server/api.js
```diff
@@ -92,21 +108,47 @@ ApiV1.addCollection(Apis, {
...
         const queryParams = this.queryParams;
-
         // Only Public APIs are available for unregistered user
         const query = { isPublic: true };
         const options = {};
 
+        // Get APIs managed by user requesting operation
+        if (queryParams.managedAPIs) {
+          // Get Manager ID from header
+          const managerId = this.request.headers['x-user-id'];
```
brylie: Remember consistency. Is it `X-User-Id` or `X-User-id`?

matleppa: In program code it is x-user-id, because JS seems to accept it only without any capitalization of letters. In documentation X-User-Id, according to way Swagger document uses the parameter name.

PR #2800, opened 2017-07-31
Manage organization manager with REST API

### LINKED ISSUE #2768: [FIWARE] REST-API: Add endpoint for Organizations part to manipulate managers

### LINKED ISSUE #2818: Rename REST APIs
In order to make APIs more consistent, the API handling API should be renamed to Catalog API and Maintenance API to Management API.

Also changes in Project directory and in program code needed.
* rename swagger structure accordingly
  * ApiV1 to CatalogV1
  * MaintenanceV1 to ManagerV1
* move swagger generation APIs from /core/server to own directory /rest_apis on main level

matleppa:
Closes #2768 
Closes #2818

New Endpoint -- /organization/{id}/managers
* GET /organizations/{id}/managers -- return list with managers. Provide information as e-mail, username, id of users.
<s>Query parameters: username, e-mail. If parameters are filled then return list of users with specified e-mail OR username</s>
* POST /organizations/{id}/managers -- add one <s>or several</s> new manager into Organization.
Body parameters: e-mails field'

New Endpoint -- /organization/{id}/managers({id}
* GET /organizations/{id}/managers/{id} -- return a specified manager
Path parameter: ID of user to return
* DELETE /organizations/{id}/managers/{id} -- delete a specified manager
Path parameter: ID of user to remove

PUT method isn't needed.
Permissions:
* Administrator or Organization mamanager can manipulate.
* Organization manager or Administrator can't delete himself.

Also includes correction in PUT /organization/{id} method functionality


organizations/server/api.js
```diff
@@ -413,8 +421,599 @@ MaintenanceV1.addRoute('organizations/:id', {
...
+          body: {
+            title: 'Not found',
+            detail: 'Queried User is not a Manager in Organization.',
+          },
+        };
+      }
+
+      // Do not include password in response
+      const options = {};
+      const excludeFields = {};
```
marla-singer: Actually, You can provide include fields using 1

brylie:
We should probably re-organize the files in `/core/server` related to the API:

- `/core/server/api.js` --> `/core/server/api/index.js`
- `/core/server/maintenance.js` --> `/core/server/api/maintenance.js`

brylie:
Lets consider a consistent way to organize our auto-generated API docs, i.e. without the word `swagger` in the path or filename.

The current solution might be changed
```js
// Generate Swagger to route /rest/v1/swagger_maintenance.json
MaintenanceV1.addSwagger('swagger_maintenance.json');
```

to use a namespace like

- /rest/v1/docs/maintenance.json
- /rest/v1/docs/another.json

brylie:
@marla-singer are you going to continuing with review for this PR?

marla-singer:
@brylie No, I just left one comment

organizations/server/api.js
```diff
@@ -413,8 +421,599 @@ MaintenanceV1.addRoute('organizations/:id', {
...
+          },
+        };
+      }
+
+      const updateManagerList = [];
+      // Get User IDs for Manager list
+
+      managerEmails.forEach((email) => {
+        // Get user account with matching email
+        const newManager = Accounts.findUserByEmail(email);
```
brylie: This should probably be called `existingUser`, since that is technically what you are checking for.

organizations/server/api.js
```diff
@@ -413,8 +421,599 @@ MaintenanceV1.addRoute('organizations/:id', {
...
+          const alreadyManager = organization.managerIds.includes(newManager._id);
+          // Check if the user is already a manager
+          if (!alreadyManager) {
+            // Add user ID to manager IDs field
+            updateManagerList.push(newManager._id);
+          }
+        }
+      });
+
+      if (managerEmails.length !== updateManagerList.length) {
```
brylie: Could this simply be an `else` block, as opposed to another `if`? E.g.

```js
if (existingUser) {
  if (!alreadyManager) {
    // add to managers list
  } else {
    // return error that user is already manager
  }
} else {
  // return error that user does not exist
}
```

brylie: Or, the negative conditions could be up front:

```js
if (!existingUser) {
  // return error saying user doesn't exist
} else {
  // check if user is manager
  if (!alreadyManager) {
    // add user ID to managerIds array
  } else {
    // return error that  user is already manager
  }
}

Comment (brylie) [inline organizations/server/api.js]: In the second example, the conditions are both negated. That may be a bit more consistent. I am not sure how to make this any more concise.

File: organizations/server/api.js [hunk 4]
```diff
@@ -413,8 +421,599 @@ MaintenanceV1.addRoute('organizations/:id', {
...
+    roleRequired: ['admin', 'manager'],
+    swagger: {
+      tags: [
+        MaintenanceV1.swagger.tags.organization,
+      ],
+      summary: 'Add one ore more new Managers into Organization.',
+      description: `
+   Adds one or more new Managers into Organization.
+   * Managers are identified with email address.
+   * Several managers can be given at a time by separating email addresses with a space.
```
brylie: For simplicity, I think we should drop the requirement that this endpoint accept multiple emails.

brylie: In effect, the operation should be atomic if possible, involving only one user at a time. That way our error handling could be more informative about which email was invalid (i.e. not a registered user or already a manager).

organizations/server/api.js
brylie: I just realized that this endpoint is probably a bit too complex. I added the note above about allowing only one email at a time.

brylie: Imagine a situation where a manager is attempting to add five, ten, or a hundred emails to their Organization managers list. All they get back is an error saying 'one or more users have no account or are already a Manager', with no further details.

Alternatively, we could tell them exactly which email is not a user, or already a manager with simpler implementation and explicit error messages

- user@domain.com is not a registered user
- user@domain.com is already a manager in this organization

organizations/server/api.js
```diff
@@ -413,8 +421,599 @@ MaintenanceV1.addRoute('organizations/:id', {
...
+          data: Meteor.users.find({ _id: { $in: organization.managerIds } }, options).fetch(),
+        },
+      };
+    },
+  },
+
+});
+
+// Request /rest/v1/organizations/:id/managers/:managerId
+MaintenanceV1.addRoute('organizations/:id/managers/:managerId', {
```
brylie: What is the purpose of this endpoint? Why doesn't the `/users/:id` API suffice for this purpose?

organizations/server/api.js
```diff
@@ -413,8 +421,599 @@ MaintenanceV1.addRoute('organizations/:id', {
...
+        200: {
+          description: 'Organization Manager successfully removed.',
+        },
+        401: {
+          description: 'Authentication is required',
+        },
+        403: {
+          description: 'User does not have permission',
+        },
+        404: {
```
brylie: How about documenting the responses for the following errors?
-  `managerId` not provided
- `organizationId` not provided
- `managerId` not found
- `organizationId` not found
- user is not organization manager

organizations/server/api.js
```diff
@@ -413,8 +421,599 @@ MaintenanceV1.addRoute('organizations/:id', {
...
+          body: {
+            title: 'Forbidden operation',
+            detail: 'You do not have permission for editing this Organization',
+          },
+        };
+      }
+      // Use list in final query
+      let managerSearchList = [];
+      // Is query limited by manager email or username
+      if (this.queryParams.managerEmail || this.queryParams.managerUsername) {
```
brylie: In other words, iterating over an array of manager emails should happen in the downstream application. That way, our REST endpoint stays simple and atomic with explicit error messages.

organizations/server/api.js
```diff
@@ -413,8 +421,596 @@ MaintenanceV1.addRoute('organizations/:id', {
...
+          statusCode: 404,
+          body: {
+            title: 'Bad parameter',
+            detail: 'One or more Users have no account or already are a Manager in Organization',
+          },
+        };
+      }
+
+      // Update Organization manager list
+      updateManagerList.forEach((newManagerid) => {
```
brylie: In other words, iterating over an array of manager emails should happen in the downstream application. That way, our REST endpoint stays simple and atomic with explicit error messages.

organizations/server/api.js
marla-singer: @brylie I suggested these parameters. The idea was to return the user with specified username or e-mail. What do you think? 

But now I had a quick mind about why did we need it if we have the same logic on Users endpoint :thinking:

brylie: Well yes, our endpoints should be as simple as possible. Do one thing well and allow output to be used with other endpoints (like the unix philosophy somewhat). So, the output of this endpoint (an array of userIds) can/should be used subsequently the Users API.

matleppa:
```Changes``` contains following changes:
* ApiV1 renamed to CollectionV1
   * /core/server/api/api.js moved and renamed /rest_apis/collection.js
* MaintenanceV1 renamed to ManagemantV1
   * /core/server/api/maintenance.js moved and renamed to /rest_apis/management.js
* Removed query with manager's username or user email
* added error response parameter documentation
* added comments (especially in cases organization :id parameter is missing and execution goes to wrong path)

File: organizations/server/api.js [hunk 9]
```diff
@@ -413,8 +421,599 @@ MaintenanceV1.addRoute('organizations/:id', {
...
+          userSecurityToken: [],
+          userId: [],
+        },
+      ],
+    },
+    action () {
+      // Get data from body parameters
+      const bodyParams = this.bodyParams;
+      // Get ID of Organization
+      const organizationId = this.urlParams.id;
```
matleppa: Added error cases. However, in case Organization id is missing from path, the execution goes to another (wrong) path, because Restivus can not exactly find out, when a parameter in middle of path is missing, but it tries to interprete the path.

organizations/server/api.js
matleppa: Added list of possible error cases in response documentations.

apis/server/api.js
```diff
@@ -11,20 +11,20 @@ import { Roles } from 'meteor/alanning:roles';
 
 // Collection imports
 import Apis from '/apis/collection';
-import ApiV1 from '/core/server/api/api';
+import CatalogV1 from '/rest_apis/collection';
```
brylie: Is CatalogV1 a 'collection'? Since we are developing APIs here, the pattern with other modules won't be exactly the same. So, we can change the folder structure to more accurately reflect what this module is doing.

matleppa: I'll change the name from collection to catalog.

brylie: Consider also renaming the directory, which is currently `collection`.

organizations/server/api.js
matleppa: There are two methods under this endpoint.
DELETE is needed to just remove certain manager from certain organization. 
With ```/users/:id``` certain user account is removed.

GET method functionality here is quite is near to corresponding method in ```/users/:id```, but the difference is, that in ```/users/:id``` only admin can query user data, here also Organization manager is allowed to query same Organization's managers' data.

brylie:
APIv1 should be renamed to **CatalogV1**, not collection. A collection in our project is a database collection, like a database table.

apis/server/api.js
matleppa: Renamed

PR #2817, opened 2017-08-08
Error: Configuration options 

### LINKED ISSUE #2465: Uncaught Error: Configuration options must be set before AccountsTemplates.init!
Error is thrown in the browser console during local development. As meteor server restarts automatically when any change is made to the project code, it causes browser to force page refresh automatically as well, after the first page refresh (right after some code change is made), following error is thrown:

### Error message in the browser console
```
Uncaught Error: Configuration options must be set before AccountsTemplates.init!
```

Second page refresh "resolves" the issue.

## Steps to reproduce
  1. When having app running locally and browser tap opened on any page (for example - `/api/<api_name>` open editor and make some change to force meteor restart server & refresh browser page.
  2. The age will be refresh with an error in the console. (This does not happen every time)

## ScreenShot
screen shot 2017-04-18 at 17 55 31

## Observed
   1. Error in the browser console
   2. API Data is not loaded

## Expected
   No errors in the console  & data API data loading fine (on first page auto refresh)

## Software versions
  - macOS `10.12.4`
  - Node `v7.4.0`
  - v8 - `5.4.500.45`
  - Meteor - `1.4.3.1`
  - Chrome - `60.0.3073.0 canary (64-bit)`

marla-singer:
Closes #2465

A possible solution of issue with `Error: Configuration options must be set before AccountsTemplates.init!`

### DIFF
--- a/apinf_packages/branding/client/lib/router.js
+++ b/apinf_packages/branding/client/lib/router.js
@@ -7,7 +7,10 @@ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence
 import { BlazeLayout } from 'meteor/kadira:blaze-layout';
 import { FlowRouter } from 'meteor/kadira:flow-router';
 
-FlowRouter.route('/settings/branding', {
+// APInf imports
+import signedIn from '/core/client/lib/router';
+
+signedIn.route('/settings/branding', {
   name: 'branding',
   action () {
     BlazeLayout.render('masterLayout', { main: 'branding' });
--- a/apinf_packages/branding/client/lib/router.js
+++ b/apinf_packages/branding/client/lib/router.js
@@ -8,7 +8,7 @@ import { BlazeLayout } from 'meteor/kadira:blaze-layout';
 import { FlowRouter } from 'meteor/kadira:flow-router';
 
 // APInf imports
-import signedIn from '/core/client/lib/router';
+import signedIn from '/apinf_packages/core/client/lib/router';
 
 signedIn.route('/settings/branding', {
   name: 'branding',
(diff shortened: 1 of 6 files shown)


core/client/lib/router.js
```diff
@@ -5,29 +5,12 @@ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence
...
 
 // Meteor contributed packages imports
 import { BlazeLayout } from 'meteor/kadira:blaze-layout';
 import { FlowRouter } from 'meteor/kadira:flow-router';
 import { Roles } from 'meteor/alanning:roles';
 
-FlowRouter.wait();
-
-Tracker.autorun(() => {
-  // The Roles package actually depends on a subscription.
```
brylie: Is the roles subscription necessary on any route? Are there routes where access control checks will fail without a Roles subscription?

marla-singer: > Is the roles subscription necessary on any route?

Not any. The routes `/settings/branding`, `/settings`, `/settings/proxies`, `/users` need access to Roles subscription checking is a user admin or not.

As you remember we have the issue where any user has access to route `/users` directly. And part of code with `FlowRouter.wait()` and `Tracker.autorun()` needed to resolve problem with un-ready subscription. 

But! I was thoughting. Why did we not have the same problem for other routes? (settings, branding and proxies). And I found that we're using the global trigger `requireAdminRole` that works fine and no problems.

In this PR I added the route `/users` to group with global trigger `requireAdminRole`

marla-singer: Unfortunately, there doesn't throw the error about AccountTemplate on my worked laptop and I can't test does it work properly by myself

brylie: @phanimahesh can you try this branch to see if your error goes away?

PR #2831, opened 2017-08-21
adding missing ids

wuder4:
adding few more id's that i missed

### DIFF
--- a/apis/client/profile/settings/details/details.html
+++ b/apis/client/profile/settings/details/details.html
@@ -26,10 +26,10 @@ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence
           doc=api
           id="apiDetailsForm"
           type="update" }}
-        {{> afQuickField name='name' }}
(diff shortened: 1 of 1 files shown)


apis/client/profile/settings/details/details.html
```diff
@@ -26,10 +26,10 @@ <h2 class="panel-title">
           doc=api
           id="apiDetailsForm"
           type="update" }}
-        {{> afQuickField name='name' }}
-        {{> afQuickField name='description' }}
-        {{> afQuickField name='url' }}
-        {{> afQuickField name='lifecycleStatus' }}
+        {{> afQuickField name='name' id="API-name" }}
```
marla-singer: @wuder4 Value of attributes, in your case ID, doesn't have to be upper case. Please use lowercase and full word
```html
id="api-name"
id="api-description"
...
```

matleppa: I agree

PR #2820, opened 2017-08-11
Integrate new Dashboard for API-Umbrella proxy

marla-singer:
Closes #2636 Implement server-aggregated aver. response time chart
Closes #2637 
Closes #2644 Implement usage figures for single API
Closes #2744 (Link to see more details about an API )
Closes  #2745 Show API information in Dashboard view for last 7 days
Closes #2740 Show Statistics of APIs owned by an user 
Closes #2742 Show detailed overview information for each REST API listed in Dashboard
Closes #2746 Date range
Closes #2753 Case when user has 1 API 
### API Analytic page
Closes #2756 API Analytics View 
Closes #2803 Show No of API calls chart in detailed view 
Closes #2763 Show Average Response Time chart in detailed view 
Closes #2760 Show KPIs for REST APIs in API details view 
Closes #2762 Most Frequent Users
Closes #2761 Errors list table

## Second iteration
- Create Groups (My Apis, Other Apis, Managed APis) ##2828
- Sorting #2758
- Filtering #2757
- Search box to find a specific API #2752

### Usability issues
- [ ] Create an issue about Add button "Watch Analytic" for APIs if it is connected to proxy and redirect to page "API Analytics View" for this API 
- [ ] Create a button "Go to API" from API Analytics View page to API Profile page

PR #2853, opened 2017-08-25
Show all backends

marla-singer:
Related to PR #2820

- Add case when proxy_backend path doesn't have a metrics but anyway display it on the table


apinf_packages/dashboard/client/view/overview/overview.html
```diff
@@ -4,69 +4,74 @@
...
+              {{_ 'dashboardOverviewStatistic_text_responseTime' time=responseTime }}
+            </p>
+            <div class="form-group">
             <span class="overview-col-title">
               {{_ 'dashboardOverviewStatistic_chartTitle_responseTime' }}
             </span>
-            {{> averageResponseTime buckets=requestOverTime proxyBackendId=proxyBackendId }}
-            <div class="{{ arrowDirection 'time' }}"></div>
-            <span>{{ overviewComparing 'time' }}</span>
+              {{> averageResponseTime buckets=requestOverTime proxyBackendId=proxyBackendId }}
```
brylie: Perhaps call this `medianResponseTime`, so it is explicit what we mean by 'average'.

marla-singer: Done

PR #2867, opened 2017-08-31
Case a user has an API is connected to current proxy

marla-singer:
Related to PR #2820

- If a user has an API is connected to current proxy then display directly API Analytic Page
- Date range picker: implement Yesterday, Last 7 days, Last 30 days; Implement query param
- Date range picker as independent component
- UI tweaks


brylie:
Fix conflicts.

PR #2876, opened 2017-09-01
Unify REST API response contents and HTTP codes 

### LINKED ISSUE #2667: [REST API] unify error responses content
When requested operation can not be fullfilled, REST API returns an error response.
However, at the moment the contents of error responses vary quite a lot. 

In some cases the returned error code is enough, but in some cases additional information is needed in order to help REST API user.

It needs to be decided, on which level additional error information is provided, so that the error message contents are unified in all methods of REST API.

More information of use of error response. 
https://tools.ietf.org/html/rfc7807

### LINKED ISSUE #2873: Unify HTTP response code use: 200, 201 and 204
Unify the success response codes for APInf REST APIs according to APInf REST API design guide.

# Goal
* [ ] 201 response is given, when adding a resource was successful.
* [ ] 204 response is given, when operation was successful, but no data can be returned. For example, the resource was deleted.
* [ ] 200 is used for all other successful operations.

### LINKED ISSUE #2874: If API cannot be added to catalog with POST, Catalog API returns an error
If adding an API to the catalog via the Catalog API fails because of database field validations, an uncontrolled exception error is got as response. 

The API must check the data against the allowed values (e.g. length) in DB definition and return a 400 response code with appropriate error text, if the API cannot be added to the database.

# Goal
* [ ] Check all fields in POST and PUT requests for adding an API to the Catalog against database validation rules / field definitions
* [ ] If the API cannot be added, return an error response

### LINKED ISSUE #2881: APInf API management.json doesn't tell that email is required
If id says its required then email should say that too.
# Screenshot 
management

### LINKED ISSUE #2894: Catalog API methods errors
Several errors in API functionality
- GET method: Admin can not see private APIs
- PUT & POST: uncontrolled error message as a response for request containing parameter isPublic
- GET method: only parameter presence is checked, not its value

matleppa:
Closes #2874 
Closes #2873 
Closes #2667 
Closes #2881 
Closes #2894 

Corrections and additions of functionality in both Catalog and Management APIs
- refactor order of checkings
- unify response messages
   - field names and contents 
   - unify use of HTTP codes 
- add missing checkings 
- add field validations in PUT and POST
- complete generated swagger documentation to be up to date

Note! Because there are lot of changes both in functionality (some changes in almost every endpoint) and in generated documentation. So in reviewer's point of view it may be easier to go through all code instead of trying to follow only commits.

### DIFF
--- a/apinf_packages/apis/server/api.js
+++ b/apinf_packages/apis/server/api.js
@@ -408,6 +408,23 @@ CatalogV1.addCollection(Apis, {
           }
         }
 
+        // Is the API set to public or private
+        if (this.bodyParams.isPublic) {
+          if (this.bodyParams.isPublic === 'true') {
+            this.bodyParams.isPublic = true;
+          } else if (this.bodyParams.isPublic === 'false') {
+            this.bodyParams.isPublic = false;
+          } else {
+            return {
+              statusCode: 400,
+              body: {
+                status: 'fail',
+                message: 'Parameter isPublic has erroneous value',
+              },
+            };
+          }
+        }
+
         // Add manager IDs list into
         const apiData = Object.assign({ managerIds: [userId] }, this.bodyParams);
 
@@ -561,6 +578,23 @@ CatalogV1.addCollection(Apis, {
               body: {
                 status: 'fail',
                 message: 'Parameter lifecycleStatus has erroneous value',
+              },
+            };
+          }
+        }
+
+        // Is the API set to public or private
+        if (this.bodyParams.isPublic) {
+          if (this.bodyParams.isPublic === 'true') {
+            this.bodyParams.isPublic = true;
+          } else if (this.bodyParams.isPublic === 'false') {
+            this.bodyParams.isPublic = false;
+          } else {
+            return {
+              statusCode: 400,
+              body: {
+                status: 'fail',
+                message: 'Parameter isPublic has erroneous value',
               },
             };
           }
--- a/apinf_packages/apis/server/api.js
+++ b/apinf_packages/apis/server/api.js
@@ -23,7 +23,7 @@ CatalogV1.swagger.meta.paths = {
       summary: 'Logging in.',
       description: `
    ### Logging in ###
-   
+
    By giving existing username and password you get login credentials,
    which you can use in authenticating requests.
 
@@ -211,6 +211,8 @@ CatalogV1.addCollection(Apis, {
         ],
         summary: 'Fetch API with specified ID.',
         description: `
+   ### Fetching API with specified ID ###
+
    Returns the API with specified ID, if a match is found.
 
    Example call:
@@ -254,6 +256,8 @@ CatalogV1.addCollection(Apis, {
         ],
         summary: 'Add new API to catalog.',
         description: `
+   ### Adding a new API to Catalog ###
+
    Adds an API to catalog. On success, returns the added API object.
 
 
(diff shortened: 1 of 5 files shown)


marla-singer:
@matleppa What status of this PR? Is ready for review?

bajiat:
@marla-singer It's WIP, but could we assign you to review it?

matleppa:
>  /login endpoint in catalog.json file: chagne status code to 201

I disagree with this one.
When a new user account is created, the HTTP code in response is 201. However, when user just logs in, the HTTP code 200 is used, because no new resource is created.

marla-singer:
Review per issues:
2873: Don't totally complete. Some endpoints don't have updated 2xx status code.
2667: Unify error response for users but it is very bad for developers. The first, you have a lot of repeated pieces of code with 400 status code.
1. Create a function for error messages with parameters statusCode, statusText, message to return object
```js
{ statusCode,
  body: {
   status: statusText,
  message: message
  }
}
```
2. The file is general for all rest api endpoints then put it in general place
3. The name of file would be `rest_api_helpers.js`
4. Change all places with `return { statusCode: 400 ...}` to function `return functionName(400, 'fail','Error message')`

2881: Complete
2894:
> GET method: Admin can not see private APIs

Complete
> PUT & POST: uncontrolled error message as a response for request containing parameter isPublic

Complete

> GET method: only parameter presence is checked, not its value

Doesn't work. Set the "skip", "limit" parameters as string and lifecycle as any value and it didn't throw error,  see below screen

marla-singer:
Responses messages. Change text for `/users` endpoints
PUT, DELETE Methods, GET method with ID path parameter
- 403 code: No permission
- 404 code: User is not found

apinf_packages/apis/server/api.js
```diff
@@ -101,27 +116,35 @@ CatalogV1.addCollection(Apis, {
...
+
         const options = {};
 
         // Get APIs managed by user requesting operation
-        if (queryParams.managedAPIs) {
-          // Get Manager ID from header
-          const managerId = this.request.headers['x-user-id'];
-
-          // Response with error in case managerId is missing
+        if (queryParams.managedAPIs === 'true') {
```
marla-singer: Handling on non-existing parameter in code. See pic below
joxi_screenshot_1505305684194

matleppa:
@marla-singer 
>   What about 500 error? I found it in one place, should we duplicate it for other endpoints?

I think that at some point it will be needed. Currently in API code there is only one place, where there is some preparation for DB errors, because there are several updates. (And that code needs to be refactored). 
In places, where there is only one DB write, we just have an enormous trust on MongoDB

```
      return {
        statusCode: 200,
        body: {
          title: 'Manager found successfully',
          data: Meteor.users.findOne({ _id: managerId }, options),
        },
      };

```

However, I think that adding of 5xx errors is to be done on some later release.

matleppa:
@marla-singer 

>  403 code: No permission
>  404 code: User is not found

The goal was to go with reason phrases [https://tools.ietf.org/html/rfc7231#section-6](https://tools.ietf.org/html/rfc7231#section-6)
and provide additional information in response messages field message.

matleppa:
@marla-singer 

>   Create a function for error messages with parameters statusCode, statusText, message to return object

Created issue #2909

marla-singer:
> @marla-singer

403 code: No permission
404 code: User is not found
The goal was to go with reason phrases https://tools.ietf.org/html/rfc7231#section-6
and provide additional information in response messages field message.
Update:
Indeed it is more informative to have larger description than plain reason phrase in documentation.

I told about consistent. Endpoints `/apis` have `403 code: No permission, 404 code: API is not found`. Endpoints `/organization` have `403 code: No permission, 404 code: Organization is not found`. 
And I encourage you make the same messages for `/users` endpoints

matleppa:
@marla-singer 

>   Doesn't work. Set the "skip", "limit" parameters as string and lifecycle as any value and it didn't throw error

Created issue #2913

matleppa:
Created issue #2908 about putting descriptions meant for swagger document into a separate file.

Issue #2656, opened 2017-06-13
API settings positioning bug


wuder4:
when you go to your api and dont have enough room on your screen to display all settings in 1 row. it changes settings tabs position to the other side.
![apinf_settings](https://user-images.githubusercontent.com/28555452/27068752-34460d7a-501a-11e7-80a8-98d209b55354.png)

# Steps to reproduce
1. go to apinf.io
2. go to api that you own
3. open console on side
4. click export tab

# Expected result
Site stays visually the same.

# Actual result
Settings jumps to the other side of the site.

bajiat:
@wuder4 Can you also add information about which browser you used?

wuder4:
Used Google Chrome Version 58.0.3029.110 (64-bit)

bajiat:
@amardeep-deligence  Is this task in progress? If yes, is there a connected PR?

amardeep-deligence:
@bajiat I am sorry, I thought this issue was closed, but only the PR was closed because no changes were there. I have raised the PR for this issue.

bajiat:
@amardeep-deligence  in the PR please mention "Closes #2656". This links the issue and the PR together If the PR gets merged, the issue is automatically closed.

Linking the issue and the PR also helps us to view them together on our Waffle board.

PR #2878, opened 2017-09-02
Fixed: API settings positioning bug #2656

amardeep-deligence:
#Fixed Issue #2656

closes: #2656


apinf_packages/core/client/style/global.less
```diff
@@ -103,6 +103,7 @@ hr {
     border: solid transparent;
     border-width: 3px 1px 1px;
     border-radius: 3px 3px 0 0;
+    height: 44px;
```
mauriciovieira: Wouldn't it be better as a relative size (.em or %) ?

PR #2914, opened 2017-09-14
Update Finnish translations

bajiat:
Update based on a new translation round


bajiat:
@marla-singer, @matleppa or @brylie Please review. I would like this to be included in the release.

PR #2928, opened 2017-09-15
Add logout endpoint into generated swagger document

### LINKED ISSUE #2897: REST API, logout endpoints are missing from documentation
In generated swagger document there needs to be  a logout endpoint. so user can log out from Catalog while examining API functionality.

Logout endpoint must be (a pair with login endpoint) in every API.

matleppa:
Closes #2897


marla-singer:
@matleppa Add a number of related issue

Issue #2428, opened 2017-04-07
APIs count on the frontend page has to depend on user permission 


marla-singer:
### Goal 
The count of APIs should correspond with the number that are visible/accessible to the user, based on permissions. 

### Screenshots
![screen shot 2017-04-05 at 17 21 13](https://cloud.githubusercontent.com/assets/95258/24725226/65bd3a1c-1a24-11e7-93cc-2f06d5b3d156.png)

![screen shot 2017-04-05 at 17 23 09](https://cloud.githubusercontent.com/assets/95258/24725393/2919699a-1a25-11e7-92ac-f8e0e0d2dcaf.png)

amardeep-deligence:
@marla-singer I have done changes for visible/accessible API counts, but have some other questions:

1. Right now API owner can not see his private api in "Latest APIs" list in home page as well as it does'nt gets counted in API Counts.
Q. Do we have to count and show owners Private APIs in "Latest APIs" list?

2. Right now if owner authorizes some user for his private API, then authorized user can see APIs Count for this private API as well, but can not see this private API in "Latest APIs" list.
Q. Do we have to show authorized private API to authorized Users in "Latest APIs" list ?

marla-singer:
@bajiat Can you please answer @amardeep-deligence questions? I don't know about User Usability in this cases

Nazarah:
@marla-singer seeing this issue has been closed. Has there been any solution to the query Amardeep asked?
@amardeep-deligence : I try to answer the questions u ask. @bajiat can validate me
> Right now API owner can not see his private api in "Latest APIs" list in home page as well as it does'nt gets counted in API Counts.
Q. Do we have to count and show owners Private APIs in "Latest APIs" list?

Yes we do. Owner should see his private API in the latest API list.

>Right now if owner authorizes some user for his private API, then authorized user can see APIs Count for this private API as well, but can not see this private API in "Latest APIs" list.
Q. Do we have to show authorized private API to authorized Users in "Latest APIs" list? When  the owner is viewing the latest API list, his private API listed there should also be count.

Yes. APInf users authorized to view the private API should also see that API in "Latest API" list in homepage.

marla-singer:
@Nazarah This issues has been closed because @amardeep-deligence resolved a bug with APIs count on the front page. If found behaviour with API list doesn't work as expected then we should open a new bug report for this

Nazarah:
>If found behaviour with API list doesn't work as expected then we should open a new bug report for this

[...]

PR #2951, opened 2017-09-19
Feature/Apis count on the frontend page

amardeep-deligence:
# Fixed issue #2428

closes:  #2428


apinf_packages/home/server/methods.js
```diff
@@ -48,7 +48,26 @@ ${doc.message}`;
...
+        isPublic: true,
+      },
+      {
+        authorizedUserIds: {
+          $in: [Meteor.userId()],
+        },
+      },
+      {
+        managerIds: {
+          $in: [Meteor.userId()],
```
marla-singer: 1. You used `Meteor.userId()` twice. Create a new variable for that
2. It has to be one more case: if a user role is admin then publish count of all apis

PR #2968, opened 2017-09-22
REST API: a function to fill error response payload

### LINKED ISSUE #2909: REST API: a function tor returning error responses
Instead of filling error response in program code, a common error function can be used to fill and return error response.

Function parameters: HTTP error code, description of error situation.

matleppa:
Closes #2909 

Instead of filling REST API error response message directly a common function is called.

### DIFF
--- a/apinf_packages/rest_apis/rest_api_helpers.js
+++ b/apinf_packages/rest_apis/rest_api_helpers.js
@@ -1,0 +1,19 @@
+/* Copyright 2017 Apinf Oy
+ This file is covered by the EUPL license.
+ You may obtain a copy of the licence at
+ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence-eupl-v11 */
+
+ // Fill and return error response message body
+ function errorMessagePayload (statusCode, messageText) {
+   // Get branding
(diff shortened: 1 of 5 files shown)


apinf_packages/rest_apis/rest_api_helpers.js
```diff
@@ -0,0 +1,19 @@
+/* Copyright 2017 Apinf Oy
```
marla-singer: Change folder place to `rest_apis/lib/` or `rest_apis/server`

apinf_packages/rest_apis/rest_api_helpers.js
```diff
@@ -0,0 +1,19 @@
...
+ This file is covered by the EUPL license.
+ You may obtain a copy of the licence at
+ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence-eupl-v11 */
+
+ // Fill and return error response message body
+ function errorMessagePayload (statusCode, messageText) {
+   // Get branding
+   const errorPayload = {
```
marla-singer: The comment doesn't match with logic below.
Change comment or remove it

Issue #2588, opened 2017-05-19
Multiple clicks on 'Delete' API button shows multiple successful messages  


saralavanip:
# Steps  
1.Visit https://staging.apinf.io  v 0.44.1  
2.Login as admin  
3.Add API  
4.Go to 'Settings'  
5. Delete' API   
6.Click on 'Delete' button multiple times in confirmation dialog box   

# Findings  
Multiple clicks on 'Delete' button shows multiple successful messages 

# Operating system 
Windows 10 

# Browsers    
1.Mozilla Firefox  v 53.0.2 (32-bit)  
2.Google Chrome  v 58.0.3029.110 (64-bit)    

# Screenshot    

![delete_api_multiple_succesful_msgs](https://cloud.githubusercontent.com/assets/13309358/26236266/263f4ef4-3c78-11e7-9526-21e9a2577c5e.PNG)

![multiple_clicks_api_delete_button_error](https://cloud.githubusercontent.com/assets/13309358/26236030/fd1349aa-3c76-11e7-812f-f852e218f4ec.PNG)

saralavanip:
@bajiat , verified on staging v 0.51.0 and it is reproducible. 

Screenshot: 

![multiple_clicks_deleteapi_messages](https://user-images.githubusercontent.com/13309358/31487523-98ed1fba-af43-11e7-9b53-89d77ff9b036.PNG)

PR #2979, opened 2017-09-25
Fixed : Multiple clicks on 'Delete' API button shows multiple success messages.

amardeep-deligence:
# Fixed issue #2588 

closes:  #2588

### DIFF
--- a/apinf_packages/apis/client/delete/delete.js
+++ b/apinf_packages/apis/client/delete/delete.js
@@ -15,6 +15,12 @@ import { sAlert } from 'meteor/juliancwirko:s-alert';
 
 Template.deleteApiConfirmation.events({
   'click #modal-delete-api': function (event, templateInstance) {
+    // Get button instance
+    const clickedButton = event.currentTarget;
+
+    // Disable delete button
+    clickedButton.disabled = true;
+
     // Get API ID
     const apiId = templateInstance.data.api._id;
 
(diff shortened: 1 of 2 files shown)


apinf_packages/apis/client/delete/delete.js
```diff
@@ -21,7 +21,13 @@ Template.deleteApiConfirmation.events({
     // Route to API Catalog
     FlowRouter.go('apiCatalog');
 
+    // Disable delete button
+    $('#modal-delete-api').attr('disabled', true);
```
marla-singer: 1. Don't use jQuery if the target is already available. You have access via `event` parameter.
```js
const clickedButton = event.currentTurget;
```

2. Change place of disabling. It must come **before** redirect to Catalog

apinf_packages/apis/client/delete/delete.js
```diff
@@ -15,13 +15,22 @@ import { sAlert } from 'meteor/juliancwirko:s-alert';
 
 Template.deleteApiConfirmation.events({
   'click #modal-delete-api': function (event, templateInstance) {
+    // Get button instance
+    const clickedButton = event.currentTarget;
+
+    // Disable delete button
+    $(clickedButton).attr('disabled', true);
```
marla-singer: @amardeep-deligence Why do you use jQuery? You can manipulate with DOM attribute w/out jQuery.

https://www.w3schools.com/jsref/prop_pushbutton_disabled.asp

### Example
> document.getElementById("myBtn").disabled = true;

In your case `clickedButton` already has "link" to element with specified ID

marla-singer:
@amardeep-deligence Good. Now squash all your commit to one

Issue #2939, opened 2017-09-18
Organization card: Misaligment elements


marla-singer:
### Place
Organization catalog
![joxi_screenshot_1505737236422](https://user-images.githubusercontent.com/20643403/30541724-3d070a7e-9c85-11e7-8311-4b913f4e685c.png)

### Expected
The alignment of Items must be in one line like for API card
![joxi_screenshot_1505738117892](https://user-images.githubusercontent.com/20643403/30542181-fe5515da-9c86-11e7-9b50-6002f1403778.png)

bajiat:
@marla-singer Not sure what this issue is about. Can you please explain?

marla-singer:
@bajiat 

### Actual result
![joxi_screenshot_1505811476637 1](https://user-images.githubusercontent.com/20643403/30584263-e5c14056-9d31-11e7-869a-6879e01b1276.png)

### Expected result
![joxi_screenshot_1505811476637](https://user-images.githubusercontent.com/20643403/30584271-e9cecae2-9d31-11e7-94e9-966aa61a31b3.png)

PR #2982, opened 2017-09-26
issue fixed Organization card: Misaligment elements #2939

saransh-dev:
# Fixed issue #2939 

Closes: #2939

### DIFF
--- a/apinf_packages/organizations/client/card/organization_card.html
+++ b/apinf_packages/organizations/client/card/organization_card.html
@@ -22,13 +22,13 @@ https://joinup.ec.europa.eu/community/eupl/og_page/european-union-public-licence
     </div>
     <div class="organization-card-action-bar">
       <ul>
-        <li class="time">
+        <li class="time pull-left">
           <span title='{{ localeFormatDate organization.createdAt }}'>
             <i class="fa fa-clock-o"></i>
             {{_ "organizationCatalogGrid_organizationCardCreated_addedText" organization.relativeCreatedAt }}
           </span>
         </li>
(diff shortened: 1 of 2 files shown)


marla-singer:
joxi_screenshot_1506424730515

brylie:
Perhaps just put the 'days ago' text and 'apis count' on their own row, so we don't have to wrestle with this issue. That way, they will always be stacked, regardless of locale.

marla-singer:
@brylie Or we can use CSS style to "hidden" part
joxi_screenshot_1506425374399

saransh-dev:
We can do one thing that if text length larger then specific limit then we can convert text like some part of that text + .... 
text is limit is 15 characters then,
Text:- `Lisatty 15 paivaa sitten`
then it will be `Lisatty 15 paiv...`

Please advise.

marla-singer:
> We can do one thing that if text length larger then specific limit then we can convert text like some part of that text + ....
text is limit is 15 characters then,
Text:- Lisatty 15 paivaa sitten
then it will be Lisatty 15 paiv...

@saransh-dev Awful. 
1. As I said above you must use CSS --> https://www.w3schools.com/cssref/css3_pr_text-overflow.asp
2. If your CSS skill doesn't allow you manipulate with text then remove "Added" world as we have in API catalog.

marla-singer:
@saransh-dev Squash your commits into one

brylie:
@marla-singer we don't have a mandate to squash commits here.

marla-singer:
@saransh-dev Awesome :confetti_ball:

PR #2992, opened 2017-09-28
Return public URL to logo instead of internal id

### LINKED ISSUE #2816: REST API: Provide public image url for APIs and organizations in APInf Catalog API
**Description**:

When listing APIs and organisations via API internal image ids have no signifigance. What is needed is public URL for the images. Current REST API implementation for this part is only for internal use. Good practice is to design and implement APIS from the outside-in. 

**User Story**:

"As an API consumer, I need to get public URLs for API logos and organisation logos in responses"

### LINKED ISSUE #2986: REST API: Replace tags 'login' and 'logout' with 'Authentication'
### Description ###
At the moment both login and logout endpoints have their own tags. 
They are also located after or in between other endpoints.

### Correction ###
It would be more clear to have both endpoints under one tag 'Authentication' and position it at the beginning of API documentation.

matleppa:
Closes #2816 
Closes #2986 

## 2816
When API or Organization data is returned via REST API, it contains also logo id, it a logo exists. 
Instead of internal logo id the link to picture must be returned. 

## 2986
In REST API documentation endpoints Login and Logout should both be under Authentication tag not under separate tags.

### DIFF
--- a/apinf_packages/apis/server/api.js
+++ b/apinf_packages/apis/server/api.js
@@ -210,12 +210,23 @@ CatalogV1.addCollection(Apis, {
           ];
         }
 
+        // Fetch the list of APIs matching with conditions
+        const apiList = Apis.find(query, options).fetch();
+        // Replace internal logo id with correct link
+        if (apiList) {
+          apiList.forEach((api) => {
+            if (api.apiLogoFileId) {
+              api.apiLogoFileId = api.logoUrl();
+            }
+          });
+        }
+
         // Construct response
         return {
           statusCode: 200,
           body: {
             status: 'success',
-            data: Apis.find(query, options).fetch(),
+            data: apiList,
           },
         };
       },
--- a/apinf_packages/apis/server/api.js
+++ b/apinf_packages/apis/server/api.js
@@ -212,14 +212,12 @@ CatalogV1.addCollection(Apis, {
 
         // Fetch the list of APIs matching with conditions
         const apiList = Apis.find(query, options).fetch();
-        // Replace internal logo id with correct link
-        if (apiList) {
-          apiList.forEach((api) => {
-            if (api.apiLogoFileId) {
-              api.apiLogoFileId = api.logoUrl();
-            }
-          });
-        }
+        // Replace internal logo id (when exists) with correct link
+        apiList.forEach((api) => {
+          if (api.apiLogoFileId) {
+            api.apiLogoFileId = api.logoUrl();
+          }
+        });
 
         // Construct response
         return {
@@ -263,6 +261,43 @@ CatalogV1.addCollection(Apis, {
             description: 'API is not Found',
           },
         },
+      },
+      action () {
+        const apiId = this.urlParams.id;
+
+        let query = { _id: apiId };
+
+        // Get Manager ID from header
+        const managerId = this.request.headers['x-user-id'];
+
+        // Check if requestor is administrator
+        const requestorIsAdmin = Roles.userIsInRole(managerId, ['admin']);
+
+        if (!requestorIsAdmin) {
+          // Only Public APIs are available for unregistered user
+          query = { isPublic: true };
+        }
+
+        console.log('query', query);
+        const options = {};
+
+        // Fetch the list of APIs matching with conditions
+        const api = Apis.findOne({ _id: apiId }, query, options);
+        // Replace internal logo id (when exists) with correct link
+        if (api) {
+          if (api.apiLogoFileId) {
+            api.apiLogoFileId = api.logoUrl();
+          }
+        }
+
+        // Construct response
+        return {
+          statusCode: 200,
+          body: {
+            status: 'success',
+            data: api,
+          },
+        };
       },
     },
     // Create a new API
--- a/apinf_packages/apis/server/api.js
+++ b/apinf_packages/apis/server/api.js
@@ -257,6 +257,9 @@ CatalogV1.addCollection(Apis, {
               },
             },
           },
+          204: {
+            description: 'No data to return',
+          },
           404: {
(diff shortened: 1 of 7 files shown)


apinf_packages/apis/server/api.js
```diff
@@ -210,12 +210,23 @@ CatalogV1.addCollection(Apis, {
           ];
         }
 
+        // Fetch the list of APIs matching with conditions
+        const apiList = Apis.find(query, options).fetch();
+        // Replace internal logo id with correct link
+        if (apiList) {
+          apiList.forEach((api) => {
```
marla-singer: 1. `[]` is true. If you want to check array existing then you have to do `apiList.length > 0`
2. This condition is redundant because `forEach` loop will not run with an empty array

apinf_packages/apis/server/api.js
```diff
@@ -210,12 +210,21 @@ CatalogV1.addCollection(Apis, {
           ];
         }
 
+        // Fetch the list of APIs matching with conditions
+        const apiList = Apis.find(query, options).fetch();
+        // Replace internal logo id (when exists) with correct link
+        apiList.forEach((api) => {
+          if (api.apiLogoFileId) {
+            api.apiLogoFileId = api.logoUrl();
```
marla-singer: I don't like this way. 
1. Field name "apiLogoFileId" means ID of logo not URL
2. You changed a collection that assumes immutable structure. I suggest creating a new structure. 

That means
```js
// Create new API list that is based on APIs collection with extended field logoUrl
const apiList = Apis.find(query, options).map((api) => {
      // Make sure logo is uploaded
      if (api.apiLogoFileId) {
        // Create a new field to store logo URL
        api.logoUrl = api.logoUrl();
      }
      return api;
    });
```

apinf_packages/apis/server/api.js
```diff
@@ -248,11 +257,62 @@ CatalogV1.addCollection(Apis, {
...
+            return {
+              statusCode: 204,
+              body: {
+                status: 'success',
+              },
+            };
+          }
+        }
+
+
```
marla-singer: Remove needless an empty line

apinf_packages/apis/server/api.js
```diff
@@ -248,11 +257,62 @@ CatalogV1.addCollection(Apis, {
...
+        // Replace internal logo id (when exists) with correct link
+        if (api.apiLogoFileId) {
+          api.apiLogoFileId = api.logoUrl();
```
marla-singer: The same stuff. Create a new field `logoUrl` to store logo URL

apinf_packages/organizations/server/api.js
```diff
@@ -95,12 +95,21 @@ ManagementV1.addRoute('organizations', {
         ];
       }
 
+      // Fetch the list of Organizations matching with conditions
+      const organizationList = Organizations.find(query, options).fetch();
+      // Replace internal logo id (when exists) with correct link
+      organizationList.forEach((organization) => {
+        if (organization.organizationLogoFileId) {
+          organization.organizationLogoFileId = organization.logoUrl();
```
marla-singer: The same stuff. Create a new Organization list is based on Organization collection

```js
const organizationList = Organizations.find(query, options).map((organization) => {
      // Make sure logo is uploaded
      if (organization.organizationLogoFileId) {
        // Create a new field to store logo URL
        organization.logoUrl = organization.logoUrl();
      }
      return organization;
    });
```

apinf_packages/organizations/server/api.js
```diff
@@ -346,6 +355,11 @@ ManagementV1.addRoute('organizations/:id', {
         return errorMessagePayload(404, detailLine);
       }
 
+      // Replace internal logo id (when exists) with correct link
+      if (organization.organizationLogoFileId) {
+        organization.organizationLogoFileId = organization.logoUrl();
```
marla-singer: The same stuff. Create a new field `logoUrl`

apinf_packages/rest_apis/server/catalog.js
```diff
@@ -229,6 +229,14 @@ CatalogV1.swagger = {
             },
           },
         },
+        updated_at: {
+          type: 'string',
+          example: '2017-07-14T01:00:00+01:00',
+        },
+        apiLogoFileId: {
```
marla-singer: Update response structure after implementing changes. 
1. Change example for this field
2. Create a new field "logoUrl"
3. Update response for Organization endpoints as well

marla-singer:
@matleppa Squash all commits that are above "Use tag Authentication instead of Login and Logout" into one with text "Return public URL to logo instead of internal id " or similar

matleppa:
I don't understand that squash commits matter.
I tried to do it locally, after several attempts I thought I found a way to squash the previous but not the last commit. However it seems to have no effect and perhaps that's why there seems to be nothing to pull to server. 
Of course the reason is my lack of experience in Git, but I'm quite close to have doubts, if it really was worth to use hours just to fail in fixing a matter, which has no effect on actual functionality.

marla-singer:
> there seems to be nothing to pull to server.

You should do `git push -f` after rebasing

> Of course the reason is my lack of experience in Git, but I'm quite close to have doubts, if it really was worth to use hours just to fail in fixing a matter, which has no effect on actual functionality.

I think you don't have enough experience with Git because you've not done it before :smile: It's a vicious circle

In the future, article about squashing https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history#squash-commits-together

marla-singer:
@matleppa I don't like the way that we have double implementation of Authentication routes (login, logout). The idea is to store it in one place and import that it will be needed.
I will create a issue for that

PR #2997, opened 2017-09-29
Fixed: API settings positioning bug

amardeep-deligence:
#Fixed Issue #2656

closes: #2656

### DIFF
--- a/apinf_packages/core/client/style/global.less
+++ b/apinf_packages/core/client/style/global.less
@@ -103,6 +103,7 @@ hr {
     border: solid transparent;
     border-width: 3px 1px 1px;
     border-radius: 3px 3px 0 0;
+    height: 42px;
   }
 
   .nav-tabs > li > a {


apinf_packages/core/client/style/global.less
```diff
@@ -103,6 +103,7 @@ hr {
     border: solid transparent;
     border-width: 3px 1px 1px;
     border-radius: 3px 3px 0 0;
+    height: 42px;
```
marla-singer: Why 42px? What is the magic number?

amardeep-deligence: I observed that non-selected <li> 's height was 42px, but selected <li> 's height was 41px, and if we select any <li> nearest to right window border, then this reduced height was causing next <li> node to shift to its bottom. That's why I specified <li> 's height 42px, So that in all cases it takes the same height. Eventually this solution worked.

Issue #2998, opened 2017-09-29
REST API: automate testing


matleppa:
### Description ###
Each REST API has several endpoints and requests sent to those endpoints have several parameters.

When an existing endpoint functionality is changed, whole endpoint functionality needs to be tested in order to be sure, that no unwanted changes occurred.

Testing endpoint's all functionality by hand requires lot of work and is time consuming.

### Solution ###
At the moment the Postman testing application is used to test single request messages.

However, it is possible to use Postman to automatize testing by 
* defining which data is  expected to exist in response message 
* form sets of tests, which can be run by pressing a button.

This would speed up the implementing new functionalities, when it were easier to cover testing old and new functionality faster.

Nazarah:
Related to #2862

Nazarah:
need to see about other test frameworks for continuous testing.

PR #3028, opened 2017-10-09
adding APIs OK and Users OK test cases

wuder4:
Adding documents I received from @matleppa to platform

#2998 

# General
- Test collections consist of set of tests, in each test a request message is sent to API and an expected result is defined. 
  - every set begins with admin login, the tokens are conveyed to following tests via environment
  - also other information can be (e.g. id of created API) conveyed via environment depending on test outcome
  - every set ends with admin logout
- test sets should be independent, so while running they create the needed resources, use them and remove them at the end, i.e. a test collection should be able to run successfully against both an empty DB and an DB with some data in it
  - while tests inside one collection depend on each other, the collections do not depend on each other
  - one exception: it must be ensured, that there is an existing admin user account, which is used in tests!
   

### Prerequisite
- stand-alone version of Postman is installed
- test sets are located in platform/tests/rest-api-tests
- an admin user account is existing

### Load tests
- Import test sets to Postman by pressing Import button
image

image

Test collections and set of tests inside collection can be seen under Collections tab. 
(Here collection APIs_OK is opened)

image

### Run tests

Start Runner by pressing Runner button

image

- Select test collection
- select environment
Install 
- press button [Run <collection name>]

image

The Run Results indicate, whether test collection was run successfully or failed.

image

In Run Results there are presented the expected results and actual outcomes of tests in collection.


wuder4:
This PR now contains all OK cases for Rest-API testing

marla-singer:
@wuder Provide information/instruction about how does review can test it

bajiat:
@matleppa Can you provide the instructions for testing?

tests/rest-api-tests/APIs_OK.postman_collection.json
```diff
@@ -0,0 +1,768 @@
...
+					}
+				}
+			],
+			"request": {
+				"url": "{{url}}/login",
+				"method": "POST",
+				"header": [
+					{
+						"key": "X-User-Id",
+						"value": "GFJzMtdzqEYgH8PHS",
```
marla-singer: What does this field "value" mean in the current context? Is it default value for X-User-Id header? Does it mean that this test will be work only in specified envirinment?

tests/rest-api-tests/APIs_OK.postman_collection.json
```diff
@@ -0,0 +1,768 @@
...
+						"description": "",
+						"disabled": true
+					},
+					{
+						"key": "Content-Type",
+						"value": "application/x-www-form-urlencoded",
+						"description": ""
+					},
+					{
+						"key": "X-User-Id",
```
marla-singer: Why key "X-User-Id" is repeated?

tests/rest-api-tests/APIs_OK.postman_collection.json
```diff
@@ -0,0 +1,768 @@
...
+						"value": "aGvjaBmoSKjNxye9Q",
+						"description": "",
+						"disabled": true
+					}
+				],
+				"body": {
```
marla-singer: Why do we need send body with GET method?

tests/rest-api-tests/APIs_OK.postman_collection.json
```diff
@@ -0,0 +1,768 @@
...
+							"",
+							""
+						]
+					}
+				}
+			],
+			"request": {
+				"url": "{{url}}/apis",
+				"method": "POST",
+				"header": [
```
marla-singer: Contains three objects with key "X-User-Id"

tests/rest-api-tests/APIs_OK.postman_collection.json
```diff
@@ -0,0 +1,768 @@
...
+							"disabled": true
+						},
+						{
+							"key": "password",
+							"value": "salasana",
+							"type": "text",
+							"disabled": true
+						},
+						{
+							"key": "company",
```
marla-singer: Why does POST method for endpoint `/apis` need body parameters as username, email, password, company?

tests/rest-api-tests/APIs_OK.postman_collection.json
```diff
@@ -0,0 +1,768 @@
...
+							"",
+							""
+						]
+					}
+				}
+			],
+			"request": {
+				"url": "{{url}}/apis/{{createdApi_id}}",
+				"method": "GET",
+				"header": [
```
marla-singer: Duplicated key "X-User-Id"

tests/rest-api-tests/APIs_OK.postman_collection.json
```diff
@@ -0,0 +1,768 @@
...
+						"description": ""
+					},
+					{
+						"key": "X-User-Id",
+						"value": "aGvjaBmoSKjNxye9Q",
+						"description": "",
+						"disabled": true
+					}
+				],
+				"body": {
```
marla-singer: The same questions:
1. Why does Get method need body parameter?
2. Why do you use this set of parameters?

tests/rest-api-tests/APIs_OK.postman_collection.json
```diff
@@ -0,0 +1,768 @@
...
+							"});",
+							"",
+							"",
+							""
+						]
+					}
+				}
+			],
+			"request": {
+				"url": "{{url}}/apis/{{createdApi_id}}",
```
marla-singer: Rename all variables "createApi_id" to "apiId" or "createdApiId"
And the other places as well

tests/rest-api-tests/APIs_OK.postman_collection.json
matleppa: ### Everything did not happen for a reason ;-)
That value for X-User-Id field is not intentional. 
The reason for it being there is just that the test sets contain of old tests, and some of them have been re-used, that's why there are extra fields.
However, they are errors, which must be corrected, just as you have stated.

### General speculation
Another thing is, that there is a weak point in this particular automated testing. 
The problem is, that there must exist a user account with admin privileges in order to be able to run test sets.

Perhaps a workaround for this is to have extra guidance.
The person running test collections must ensure, that in environment, where the tests are run, a certain admin user account with certain user id and related password must exist. 
It must not be changed and when new test sets are created, the same admin user account must be used. 
However, in case these test at some point might be run as part of Travis, some solution is needed for this primary admin problem. Or perhaps in other Travis tests there already are created some admin, which could be used also in these tests, we only need to know the name and psw of the test admin account.

marla-singer: @matleppa Can you please share with me docs about how write test for Postman as JSON format? As for me the provided tests look not good, a little bit dirty. And I don't know does it expect or code needs some clean up

matleppa: These tests were made by using Postman stand-alone version.

1. At first the tests were made and run.
2. Then collections is created and the related tests were attached to collections.
3. After that collections were exported. The export tool outputs the tests on JSON format.

So when new tests are needed, although it is possible to write the tests directly into JSON file, I think that rather than writing tests on JSON format and importing them to Postman, it is more practical to import collections from repository, then make new tests on Postman, add them to collection and then export the collections and store them to repository. 
 
I think that we can not affect on such things as indentation of outputted JSON file, at least I could not find an option for that in Postman settings.

marla-singer: > I think that we can not affect on such things as indentation of outputted JSON file, at least I could not find an option for that in Postman settings.

But it ins't readable for developers

marla-singer: But okay. Give this PR the care in your hands

matleppa: OK. 
Still I have to say, that you did very good work finding those matters which make tests messy, even you were not familiar with the tool. Thanks!

PR #3102, opened 2017-11-07
Update docker-compose config for simplify dump configuration

shaliko:
## Changes
- Switch Api-Umbrella to use shared 'mongo' service.
- Clean up old docker-compose configurations
- Update depends_on for API umbrella service. Set order for boot services
- Refactoring docker-compose env file names
- Use named volume for mongo service. For simplifying debug


anarva:
OK, started the review process.

PR #3306, opened 2018-01-05
Feature/rest api several doc links

### LINKED ISSUE #3220: Several documentation links adding possibility catalog api
### Description
PR 3201 changes functionality and structure of Apidocs such a way, that it is possible to provide several external documentation links for one API.

### Changes needed in Catalog API
After the PR 3201 is implemented, the structure of ApiDocs is changed. A change is needed to Catalog API in order to write external documentation links correctly.

New endpoints are needed to handle adding and removing external documentation links
- external links can be added and removed one by one
- when a link is added, 
  - it is checked, that a duplicate link does not already exist
  - maximum limit of existing links is 8
- only an existing link can be removed

Also functionality of documentationUrl is moved to these new endpoints.

That way there is no need to handle different DBs in one endpoint, which increases integrity.

matleppa:
Closes #3220 

## Changes
- POST /apis and PUT /apis
  - external documentation links stored into array instead of one field
  - checking of update operation outcome is added
  - rollback in case tried to update both API object and documentation and update fails partially
- DELETE /apis/id/documents 
  - without query parameters the apiDoc document for API in question is removed
  - with query parameter (DELETE /apis/id/documents?url=<documentation-link>) the identified documentation link is removed
    - links are removed one at a time
    - matching link is removed, checking is started from openAPI document link, then external documentation links
  - 

## Developer checklist
This checklist is to be completed by the PR developer:

- [ ] Alternative solutions were compared/discussed before writing code
    - [ ] trade-offs with this solution are considered acceptable
- [ ] Code in this PR adheres to the [project styleguide](CONTRIBUTING.md)
- [ ] This pull request does not decrease project test coverage
- [ ] If the code changes existing database collection(s), migration has been written
- [ ] If UI texts are added or changed, all texts are internationalized

## Reviewer checklist
Reviewed by: @username1

This list is to be completed by the pull request reviewer:
- [ ] Code works as described/expected
- [ ] Code seems to be error free
    - [ ] no browser console errors visible
    - [ ] no server console errors visible
    - [ ] passes CI build
- [ ] Code is written in a way that promotes maintainability
    - [ ] easy to understand
    - [ ] well organized
    - [ ] follows project coding standards and conventions
    - [ ] well documented

### DIFF
--- a/apinf_packages/apis/server/api.js
+++ b/apinf_packages/apis/server/api.js
@@ -638,7 +638,6 @@ CatalogV1.addCollection(Apis, {
         if (documentationUrl || externalDocumentation) {
           // Try to fetch existing documentation
           const apiDoc = ApiDocs.findOne({ apiId });
-          console.log('apiDoc=', apiDoc);
           // Regex for http(s) protocol
           const regex = SimpleSchema.RegEx.Url;
 
@@ -679,8 +678,10 @@ CatalogV1.addCollection(Apis, {
                 return errorMessagePayload(400, message);
               }
             }
-            // Prepare for rollback after possible failure
-            previousDocumentationUrl = apiDoc.remoteFileUrl;
+            // Prepare for rollback of openAPI documentation after possible failure
+            if (apiDoc && apiDoc.remoteFileUrl) {
+              previousDocumentationUrl = apiDoc.remoteFileUrl;
+            }
           }
 
           // Update Documentation
@@ -695,7 +696,6 @@ CatalogV1.addCollection(Apis, {
           );
           // If insertion of document link failed
           if (result === 0) {
-            // Remove newly created API document
             return errorMessagePayload(500, 'Update failed because Documentation update fail.');
           }
         }
@@ -707,11 +707,11 @@ CatalogV1.addCollection(Apis, {
         // Update API document
         const result = Apis.update(apiId, { $set: bodyParams });
         // Check if API update failed
-        if (result !== 0) {
-          // try to rollback documentation update, if necessary
+        if (result === 0) {
+          // Try to rollback documentation update, if necessary
           if (documentationUrl || externalDocumentation) {
             if (documentationUrl) {
-              // Restore previous documentation link, if it exist
+              // Restore previous openAPI documentation link, if it exists
               if (previousDocumentationUrl) {
                 ApiDocs.update(
                   { apiId },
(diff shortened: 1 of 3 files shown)


apinf_packages/apis/server/api.js
```diff
@@ -443,13 +444,14 @@ CatalogV1.addCollection(Apis, {
...
 
         // Documentation URL must have URL format
         if (documentationUrl) {
-          isValid = ApiDocs.simpleSchema().namedContext()
-            .validateOne({ remoteFileUrl: documentationUrl }, 'remoteFileUrl');
-
-          if (!isValid) {
+          // Check link validity
+          const regexUrl = regex.test(documentationUrl);
+          if (!regexUrl) {
```
marla-singer: We should use `regex.test(variable)` directly in the `if` condition
```js
if (!regex.text(documentationUrl)) { ... }
```
Because for every URL variable has owned checking function and I don't see any reason to store every result of testing in the separated variable. On another hand, the variable `regexUrl` is used only once

**Remove all variables regexUrl and use regex.text() directly**

apinf_packages/apis/server/api.js
```diff
@@ -617,57 +630,73 @@ CatalogV1.addCollection(Apis, {
...
         if (documentationUrl || externalDocumentation) {
           // Try to fetch existing documentation
           const apiDoc = ApiDocs.findOne({ apiId });
+          // Regex for http(s) protocol
+          const regex = SimpleSchema.RegEx.Url;
+
+          // Check link to Documentation URL
+          if (documentationUrl) {
+            // Check link validity
+            const regexUrl = regex.test(documentationUrl);
```
marla-singer: What will `regex.text(undefined)` return? Does it really need to check if `documentationUrl` exists?

apinf_packages/apis/server/api.js
```diff
@@ -617,57 +630,73 @@ CatalogV1.addCollection(Apis, {
...
+          const result = ApiDocs.update(
+            { apiId },
+            { $set: {
               type: 'url',
-              otherUrl: bodyParams.externalDocumentation,
               remoteFileUrl: bodyParams.documentationUrl,
-            });
+            },
+              $push: { otherUrl: bodyParams.externalDocumentation } },
+            { upsert: true },
```
marla-singer: Add a comment with couple words above about "upsert" option

apinf_packages/apis/server/api.js
```diff
@@ -767,3 +835,147 @@ CatalogV1.addCollection(Apis, {
...
+        }
+      } else {
+        // Remove all documentation, because no link identified
+        Meteor.call('removeApiDoc', apiId);
+      }
+
+      // Prepare data to response, extend it with Documentation URLs
+      const responseData = Object.assign(
+        // Get updated value of API
+        Apis.findOne(apiId),
```
marla-singer: Variable `api` already contains the API data and it doesn't need to be fetched again

**Use "api" variable instead of Apis.findOne(apiId)**

apinf_packages/apis/server/api.js
```diff
@@ -767,3 +835,147 @@ CatalogV1.addCollection(Apis, {
...
+          if (removeResult === 0) {
+            const message = 'OpenAPI Documentation link removal failure.';
+            return errorMessagePayload(500, message, 'url', documentUrl);
+          }
+        } else if (apiDoc.otherUrl.includes(documentUrl)) {
+          // Matching link found as external documentatin, try to remove
+          const removeResult = ApiDocs.update(
+            { apiId },
+            { $set: {
+              type: 'url',
```
marla-singer: Why do we need to update type in this case? For example, If the documentation is uploaded as file and a user removes the external URL via DELETE method then it doesn't expect to change `type`

**Remove that**

apinf_packages/apis/server/api.js
matleppa: Check for variable existence is needed, because regex.text(undefined) returns error.

apinf_packages/apis/server/api.js
```diff
@@ -474,20 +476,28 @@ CatalogV1.addCollection(Apis, {
...
-            remoteFileUrl: documentationUrl,
-          });
+          const result = ApiDocs.update(
+            { apiId },
+            { $set: {
+              type: 'url',
+              remoteFileUrl: bodyParams.documentationUrl,
+            },
+              $push: { otherUrl: bodyParams.externalDocumentation } },
+            { upsert: true },
```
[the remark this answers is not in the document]
matleppa: Indeed.

Oh well, what an amount of unnecessary work by me. 

At first place being blind for obvious solution 
- (insert and otherUrl: [externalDocumentation],)

 and trying to use $push instead, which lead to following error:
*"When the validation object contains mongo operators, you must set the modifier option to true"*, which lead to complicated work-around 
- (update, $push and upsert)

Blimey!

PR #3336, opened 2018-01-17
Added node v6 and Chrome version output.

anarva:
Do not merge.

Added some output for debugging Travis.


anarva:
Failed experiment.

PR #3519, opened 2018-07-03
Revoke my API key

vinaytech:
Closes # Revoke my API key

## Changes

- When API key is revoked, old API key still works. When API key is revoked, old one has to stop 
  working. The case is if API key gets lost, should be possible to generate new one, and disable old 
   one.

## Developer checklist
This checklist is to be completed by the PR developer:

- [ ] Alternative solutions were compared/discussed before writing code
    - [ ] trade-offs with this solution are considered acceptable
- [ ] Code in this PR adheres to the [project styleguide](CONTRIBUTING.md)
- [ ] This pull request does not decrease project test coverage
- [ ] If the code changes existing database collection(s), migration has been written
- [ ] If UI texts are added or changed, all texts are internationalized

## Reviewer checklist
Reviewed by: @username1

This list is to be completed by the pull request reviewer:
- [ ] Code works as described/expected
- [ ] Code seems to be error free
    - [ ] no browser console errors visible
    - [ ] no server console errors visible
    - [ ] passes CI build
- [ ] Code is written in a way that promotes maintainability
    - [ ] easy to understand
    - [ ] well organized
    - [ ] follows project coding standards and conventions
    - [ ] well documented


phanimahesh:
Deletes are disabled. https://github.com/apinf/api-umbrella/blob/master/src/api-umbrella/web-app/config/routes.rb#L43

It's an easy fix to enable them, and they do seem to work. But looking at the code, there's an option to disable a key without deleting it. However, it isn't exposed in the API, or I can not figure out how to use it. I'll hold this to see if we can do it the right way by disabling the keys. If not, we can go with the quick fix for deleting.

phanimahesh:
the required api to delete a user is now enabled on nightly api umbrella
:3002. It sets the disabled property internally and does not directly
delete the user/key. Disabled keys should return API_KEY_DISABLED error
code,  please verify.

PR #3699, opened 2019-01-23
Meteor update

brylie:
Upgrade Meteor to 1.8 and add Babel latest.


brylie:
This is superceded by #3705

